PR #1386, opened 2013-08-12
Adaptive slicing

platsch:
This is a first approach towards automatic layer height configuration.
The new Print Settings option "Adaptive slicing" determines the height of each individual layer by evaluating the objects surface at this region. It attempts to maintain the error caused by the stair-stepping-effect below the given cusp-value, basically as described in (e.g.) http://www-rp.me.vt.edu/bohn/projects/sabourin/thesis.pdf p.33ff, but with some modifications.

Some technical implementation details:
I've added a sub slice_adaptive in Object.pm which can be called instead of slice. Slice_adaptive makes use of some more or less generalized functions provided by AdaptiveSlicing.pm, where the surface evaluation stuff is stored. 
The cusp-height algorithm needs min and max layer height capabilities for every extruder in use. This options are therefore added to the Extruder config in Printer Settings.
The first layer height is taken from the config. Subsequent layers are generated based on the normals of the facets crossing the $print_z layer and refined to match horizontal features and peaks/abrupt slopes.
Some basic testing is implemented in adaptive_slicing.t.

This is work in progress. Please give some feedback/opinion whether people are interested in this feature and about my implementation and integration into Slic3r. My experience in open source development is very limited and I'd like to avoid spending a lot of time on things no one will ever use :)

Next step would be implementing further control parameters for surface quality and overhanging areas and multi-extruder support to mix extruder diameters depending on the local structure of an object.

adaptive_servo_gcode


oysteinkrog:
Has there been any progress on this?
I think this would be an awesome feature and if there is something blocking the merge it would be nice know what, so that those who wanted to could work on it.

platsch:
I have not touched it since my last commit in Aug. 2014. I tried to make it easy to merge this feature into the master branch back then and spend quite some time on it. I'm not fully aware of all the changes in slic3r, so I'm not sure how easy the merge would be now.
I dropped my efforts since no one seemed to be really interested in this work, but would of course be willing to contribute again if that changes.

alexrj:
@platsch, I'm definitely interested.  I'm just overwhelmed by the amount of work on Slic3r (and other things, as Slic3r lies in my spare time...) but I'll definitely get back to you.

platsch:
Well, yes, I see your problem, any chance to spread the work on more shoulders in the mid/long-term?

lordofhyphens:
@platsch if you want to port it to the XSslicer and produce some test cases for it I will take a look at merging it in.

platsch:
Uhm, it's been a while since I touched the code, but sure, I'd still love to see it integrated. Give me a few days, I try to find the time for a port.

platsch:
@lordofhyphens I updated the code to be compatible with the recent master branch. However, the computations are still done in perl, but porting as much as possible to XS is on my todo-list.
What is your requirement for test cases? There already is a set of perl unit tests for the adaptive slicing implementation.

lordofhyphens:
That is good enough for me. I'll take a look over it this week and tag it,
as I would like for it to get more widely tested "in the field" as it were.

Thanks!

lordofhyphens:
Right, is the behavior on a toggle, or is it a permanent change to the slicing behavior?  If it's permanent I think it would be useful to be able to turn it on/off via a Print setting (for speed). I can submit a PR to your branch if necessary to add the changes if necessary.

platsch:
Adaptive slicing can be toggled at the `Layer height` field in the `Layers and perimeters` section. Default is off, no side effects on the processing / memory / speed. It also requires a min and max value for the possible layer height achievable by the printer at the 'Extruder' section. I find it quite hard do accurately describe the effect of the options for this feature, any improvements are very welcome.

One remaining issue: all 3 options located in the `Print settings` section toggle a reprocessing, but I wasn't able to figure out how this can be done for the min/max layer height yet.

VanessaE:
Once built, it crashes when run:
`No adaptive_slicing in ConfigOptionsGroup config at /usr/share/perl5/Slic3r/GUI/OptionsGroup.pm line 344.`

If I move my Slic3r config out of reach, I can run it as long as I keep it in "Simple" mode.  Switch it to "Expert" mode and exit/restart, and it crashes with the same error.

Here's the full log of what I did (git output, build log, etc, 1532 lines):
http://pastebin.ubuntu.com/17074218/

lordofhyphens:
I'll take a look at this tonight; there's 3 or so places you have to look at to define a new option, all for different reasons. It's just crazy enough to make me consider some form of code generation for defining options, but not crazy enough for me to learn enough Perl to go through with it.

platsch:
@VanessaE apparently something went wrong with the git-magic. I tried to reproduce your steps and end up with the adaptive-slicing state from 2014 with an older options implementation. Manually pulling the adaptive-slicing branch from my repo solved the issue for me.

VanessaE:
Cloned from your repo, switched to the adaptive-slicing branch, built using the Build.PL method.  That works fine.   @lordofhyphens I guess it's your Debian code where that error comes up.  You bwoke it! :smile:

Ironically, the models I had in mind for testing this code don't actually benefit from it (at least the sliced result - I didn't re-print them).  Go figure.  So, I threw something more complex at it:

[aleph-2-color.zip](https://github.com/alexrj/Slic3r/files/310339/aleph-2-color.zip)

When I export this to g-code, with a starting layer height of 0.2mm, cusp 0.15, and a layer height range of 0.05 to 0.35, a dialog pops up shortly after the export starts, reading:

`Can't call method "facets" on an undefined value at /home/vanessa/RepRap/Slic3r-git-adaptive/lib/Slic3r/AdaptiveSlicing.pm line 20.`

...after which the "Print" and "Export G-code" buttons were greyed-out, forcing me to restart Slic3r.

If I restart Slic3r, switch to my "single material" and "single extruder" adaptive profiles (same adaptive settings as the "dual" ones; see the config bundles below), and then load the single-color version of that same model:

[aleph.stl.zip](https://github.com/alexrj/Slic3r/files/310368/aleph.stl.zip)

This one slices fine.  No apparent errors.

Config bundles: 
[Prusa i3 - Slic3r 1.3.0-dev config bundle-adaptive.ini.zip](https://github.com/alexrj/Slic3r/files/310364/Prusa.i3.-.Slic3r.1.3.0-dev.config.bundle-adaptive.ini.zip)
For the filament, I was using my "Inland PLA, white" profile.  For the Print and Printer settings, use the appropriate pair of "single, adaptive" or "dual, adaptive" profiles.

As an aside, you should add something that lets the user specify the exact gradation between min and max heights (lest the adaptive code create weird layer heights that can't be well-represented on a given user's hardware).

platsch:
I'm not entirely sure what exactly caused this problem. The dual material object consists of two objects, each with 2 "regions", but only one of the regions contains a volume. This leads to an uninitialized value in the `AdaptiveSlicing.pm`. I just fixed this by adding an additional check for empty region-volumes, but it doesn't really feel like a clean solution. It creates a number of empty layers where the "upper" part of the object contains no material. Those layers are later removed, so the print is not affected, but it confuses the user and leaves me unsettled.

@VanessaE regarding your suggestion: why should any uneven or weird layer height might not be well-representable on any printer? I don't see much of a problem here...

VanessaE:
The reason I suggest against it is that it will introduce wobble-like artifacts over a large enough section of a model.  Anything that doesn't advance the Z axis by an integer number of steps will force the printer to round the height to the nearest step, introducing repeating patterns into the print as the rounding error accumulates.

Consider a typical model like the Prusa i3, with 1/16 micro-stepping on the Z module, driving a pair of M5 threaded rods.  With a common 1.8 degree motor, that makes 4000 steps/mm for the Z axis.  In practice, a multiple of 0.004 is best, to avoid micro-stepping inaccuracies.

As for that model, Slic3r seems to think all four regions (two of which are just index/positioning marks) are manifold.  I'm curious, can you post a screenshot showing the "confusing" layers?  Might give me some idea of what needs fixed in the model.

platsch:
I can post a screenshot of the "confusing" layers, but it simply would be an empty layer...
Maybe my description was misleading. The object consists of two major parts: the sign or letter which occupies roughly the upper half and the "base" for the lower half. Slic3r generates layers for both parts independently. With "static" slicing, both parts generate the same set of layers, e.g. every 0.2mm. With adaptive slicing each part generates it's own set of layers which usually do not coincide. The lower part of your model only generates layers from 0 to it's top, the upper part does the same but contains no material at the lower half. The `Preview` and `Layers` tabs show all layers for all objects, including the empty layers.
You can reproduce this with the updated current version of the `adaptive_slicing` branch from my repo. Just slice your model, step through the layers in the `Layers` tab and look out for empty layers.

I can see your point for the gradation-issue, but on the other hand this would introduce additional complexity which makes it harder to use for someone without deep technical inside of the printer. For your Prusa I3 example this effect would cause an error of 4% every 4 layers for a minimal layer height of 0.1mm in a worst case scenario. Is there any evidence of this being a substantial problem? Any examples prints?

VanessaE:
This page explains it a bit better:  https://www.evernote.com/shard/s211/sh/701c36c4-ddd5-4669-a482-953d8924c71d/1ef992988295487c98c268dcdd2d687e

(search in page for the phrase "As people have been pushing for higher-quality" to jump to the relevant part)

I'll take a look at your update soon.

As for "deep" knowledge of the printer, I was under the impression that your Z steps/mm and the idea of sticking to whole steps were basic sorts of things to know.

platsch:
Ok, I'll think about an option to limit the layer height gradation.

lordofhyphens:
Here's a win32 build of the branch as of this comment; [slicer-adaptive.zip](https://github.com/alexrj/Slic3r/files/336169/slicer-adaptive.zip)

lordofhyphens:
@platsch I will try to keep an eye on this PR and update accordingly; I have the build server up and running to generate win32 builds for other people to try things:
https://bintray.com/lordofhyphens/Slic3r/Slic3r_Branches

I also rebased adaptive-slicing from your branch onto current master and merged the result into the branch. I suggest pulling that branch if you do any further work on this. 

Currently if it is a branch on my fork it will get processed by the Windows build environment.

Alternatively you could make a pull request against the same-named branch on my repository.

platsch:
Just tried to test the win32 packages. I'm not a windows user, so I set up a fresh installation in a virtual machine. Slic3r works, but I am unable to start the GUI. Seems to be an OpenGL issue, though the OpenGL.xs.dll is deployed correctly...

lordofhyphens:
... I was sure I fixed that -.-;.

Patola:
Any news about this feature? I find it so useful but it seems to be inactive now.

lordofhyphens:
@platsch is the person doing the work on it.

platsch:
Yes, and I'm still active and willing to contribute.
I'm a bit unsure on how to proceed with this. I'd be lucky to get any feedback / test results on this feature. The height gradation issue mentioned by @VanessaE  is still on my todo-list.
Any hints how to trigger re-slicing for the min/max layer heights values mentioned in https://github.com/alexrj/Slic3r/pull/1386#issuecomment-223959826?
Apart from this, is there any other important reason why this can't be merged yet?

lordofhyphens:
If you rebase to master again, I can pull it to my fork (which will trigger the automatic builds for Windows). This should let more people work on it.

As for "what next for merge", I need to review it more closely and then get @alexrj to approve it.

lordofhyphens:
As for adding it to the things that trigger reslicing: I haven't found it yet but you can probably start at 
https://github.com/alexrj/Slic3r/blob/master/lib/Slic3r/GUI/Plater.pm at the start background process or export gcode

bubnikv:
Guys, I am confused. Where can I find the latest reincarnation of this feature? I would love to have it. Thanks

lordofhyphens:
@bubnikv that would be as far as I know this pull request (or the branch it
came from).

If you want to address the issues that have been brought up so far then
great.

bubnikv:
> @bubnikv that would be as far as I know this pull request (or the branch it
came from).

I see. I am still learning github. I did not realize that the pull request has been extended with additional patches over time.

I hope I will look into this once I fix the supports to be comparable to Kisslicer or Simplify3D in regard to the ease of removal.

lordofhyphens:
Yeah, on github the pull request is always the current source branch. This
is why my work flow always starts from a clean copy of master. When I am
done with a feature, I don't commit unrelated things to that branch. Makes
it much easier to collaborate and forces you to think about making sweeping
changes.

I know of other projects where basically everything that gets committed to
the master branch is via a pull request from the contributors
(smoothieware).

lordofhyphens:
Clarification: after a pull request is merged, that's it; the branch is no
longer tracked for merging. If you made more commits after the merge
happens, submit another pull request.

platsch:
OK, just updated to current master. Would be great if you could pull & build for windows.

@bubnikv yes, this pull request (or the corresponding branch in my [fork] (https://github.com/platsch/Slic3r/tree/adaptive-slicing)) contains the most recent version. As @lordofhyphens explained, this branch contains only changes related to the adaptive slicing concept and is equal to alexjr:master otherwise.

lordofhyphens:
I've opened a separate PR onto my fork for the time being (can't use my normal git client at work). https://ci.appveyor.com/project/lordofhyphens/slic3r/build/1.3.0-master-195

platsch:
I tried to execute it on a plain Win7 installation (in a VirtualBox), but get an error: `Your vendor has not defined OpenGL macro GL_MULTISAMPLE used at 3DScene.pm line 634`.
I never use Slic3r under windows and I'm not very familiar with the OpenGL stuff. After a quick [search] (https://stackoverflow.com/questions/4207506/where-is-gl-multisample-defined), this seems to be plausible, but then it should have been broken since 2014 https://github.com/alexrj/Slic3r/commit/a82f95e9034c5bf59c2c40e24a57a6ee30692601, which is not the case?

trbielec:
FYI yesterday Autodesk released a CC BY 3.0 open source variable layer height algorithm called VariSlice™. It's a Processing script for ASCII STL files sliced at 5µm for the Ember printer, but the general algorithm should work for FDM as well. The output is a CSV file with the layer height for each layer. You can find the code and instructions here: http://www.instructables.com/id/Variable-Slicing-for-3D-Printing-on-Autodesk-Ember/

platsch:
Very interesting, thanks for the hint! My first impression is that they are basically using the same approach, but with a more iterative way of refining the the layer height for sloped facets intersecting only with the upper part of the current layer.  Besides, there algorithm is optimized for a fixed grid, which is not a general difference, but helps setting a good value for the resolution control parameter (cusp height).
I'm a bit confused be the low performance. Computing a set of layer heights for the included demo object `dumbbell-3-ASCII.stl` takes ~10.5s on my machine. Slic3r generates a full adaptively sliced toolpath with infill, support etc. in ~11s for the same model and resolution...

stevecrayons:
@platsch Hey! I just saw this pull request. A lil' background on VariSlice™: I work with the Ember 3D printer at Autodesk and I was looking for cool examples that showcase the open nature of the printer -- particularly how you can easily adjust any printing parameter on a per-layer basis with a simple CSV file. Variable layer heights came to mind and I wrote a super simple script that created the correct layer heights but ONLY if the model was a disk or dome. Would not work for anything else. So I printed some semi-disks that looked good. I asked myself if it was possible to have it work more generally by looking at an STL, and I was like "I think so". My brilliant colleague Cappie wrote the code that parses an ASCII STL, I used that and in a week hacked a minimum working script. I was super pleased with the result. If I had known about your code I would have acknowledged it, but likely still would have written the Processing script anyway. Mostly because your code is a bit beyond me and I don't really understand things unless I do them myself.

Confused about the low performance?? Simple explanation: the code was written by someone who knows almost nothing about optimizing code :D 

ANYWAY, i included a link to this code in the description to the VariSlice™ youtube video and also an annotation around the 6 min mark: https://www.youtube.com/watch?v=HAmneiL5-jQ

Looking forward to seeing your work progress!

(btw VariSlice™ isn't actually trademarked -- we just thought it looked more important with a little tm on the end)

platsch:
Well, it's good to see that there is still interest in this field after ~15years without significant contributions since Tyberg, Sabourin etc. published their research.

platsch:
Great, but would you consider setting the facebook post public? I'm currently not allowed to view it ;-)

chanders:
It's a 3D printing group with over 20,000 members if I am not mistaken. Unfortunately I am not the group owner :( you can request to join. Approval usually come in ages mins.

There are a lot of great info in the comments! Here is a snapshot of the post

https://imgur.com/gallery/1Z48g

chanders:
^^ "In a few mins"

platsch:
Ok, I didn't realize that it is a semi-public group plus I don't have a facebook account...

platsch:
Just added a few more testcases. Proved to be a good idea, found a bug in the horizontal feature detection right away.

platsch:
@VanessaE I found the time to implement the gradation feature you [suggested] (https://github.com/alexrj/Slic3r/pull/1386#issuecomment-225374251). It is available in this branch https://github.com/platsch/Slic3r/tree/adaptive-slicing-layer-gradation, feedback is very welcome.

lordofhyphens:
I'll make a PR on my local fork of Slic3r for it so it gets built.

ehybrid:
I haven't looked at the code so feel free to ignore if this is at all a hard or silly question:-
how easy would something like this to be to extend to only print external layers at the lower height?
(two strategies I've thought about illustrated here:- https://i.imgur.com/tNcQwzR.png)

platsch:
Should work out of the box. At least as sketched in the "Suggested" image. Just set the `Combine infill every layers` to a number > 1 (e.g. 10) then it will combine as many infill layers as possible (up to 10 layers in this example). There already was a short discussion in https://github.com/alexrj/Slic3r/pull/1386.

For your second "alternative" illustration: no idea how this could be done except for special cases.

ehybrid:
I was meaning only a single external perimeter as the variable thickness, with internal perimeters at full thickness, and infill, combined or otherwise, inside that.

platsch:
Ah, now I understand your image. That's probably possible, but I would expect some pitfalls: it kind of breaks the way Slic3r organizes layers, but that's already the case for combined infill. You would probably need something like an adaptive extrusion width which could produce a more coarse or "irregular" surface.

ehybrid:
Right, variable width for my 'middle' strategy, (labeled in picture, perhaps not clearly enough) ... or for the bottom 'overhang' strategy, variable volume for the squished-in part of the 2nd outermost perimeter (though I guess variable volume/width are the same thing if your nozzle size isn't changing)

ehybrid:
Hmm, ok, so they're just voiding out layers and extruding twice as much plastic for combined infill...  unless I'm mistaken I don't think my strategy can be implemented that way since I need to do it on perimeters, not regions...  coupled with the need for variable-extrusion to get the internal perimeters to line up with the externals,  probably needs a pretty deep rewrite of a chunk of the slicing code then...

Though... Isn't there has to be some variable-volume extrusion capability already:- how does the "infill small gap" code manage that when there's say a 0.2mm hole to fill in a top layer with a 0.4mm nozzle?  It doesn't just over-extrude until the filament slips does it?

ehybrid:
(is this the right place for this discussion?)
Slightly deeper look at the code suggests there'd be basically no way around rewriting PerimeterGenerator() .. seems it already has knowledge of previous layers too which should be enough for the variable-volume 1st-or-2nd-outer perimeter.    I'm a little unclear if I need to do the void-collapse trick to combine perimeter layers or if it's better to try and generalise it and not be shy about breaking layer assumptions.

ehybrid:
(and regarding your adaptive extrusion width producing irregular surfaces -- that's only true once you get below 105% of nozzle width from what I've been led to believe...  the internal perimeters can be shifted inwards or shrunk to guarantee the outer perimeter stays above the minimum width)

ehybrid:
@lordofhyphens I've been trying to use the win32 builds from bintray to test this adaptive feature, the version referenced above doesn't exist any more, and the few newer ones than that version I tried give a "perl interpreter has stopped working"  error on win10...  Is there a current known problem with the builds?  Would appreciate the cluebat me if not, ty.

lordofhyphens:
They were probably bit by the opengl bug listed in #3540. I will get the
build system to produce a slightly modified version that has those fixes in
it.

Otherwise the builds are as they come off of @platsch 's branch.

bubnikv:
I just checked the  adaptive-slicing-layer-gradation branch. IMHO the feature in the way it is implemented right now has a use limited to simple technical prints. The adaptive layer height is calculated from a maximum slope in a layer, independent of the model complexity, triangle size, visibility of the triangle etc. Therefore for a complex object the algorithm often picks a very low layer height even though the user would not benefit from it. Also I believe the layer height should change slowly over multiple layers to be visually pleasing.

Discussing with our in house print expert guy, he would rather use some kind of spline manipulator over the height of the model, which would allow him to vary the layer height smoothly.

I think some kind of triangle coloring by their cusp value and a user interface to relax the value where the visual quality is not important to the user would help to make the feature more useful.

lordofhyphens:
@platsch fyi I have done a rebase of this to current master in lordofhyphens/slic3r (branch adaptive-slicing). It should get picked up by my build server for win32.

I think I'm going to do a squash merge of it into my fork, you can force-pull the branch over to your repo, which should make future rebases easier at the cost of losing some granularity of your commits. 

Thoughts?

platsch:
@bubnikv very interesting feedback.
I already considered to factor in the triangle size for the height computation, but didn't come up with a good solution so far.
Slowly changing the layer height would probably require more pre-computation to check for height limitations in the future (i.e. in the next few layers), but is definitely a good and feasible idea.

Do you have good examples where the visibility of triangles or general object complexity causes unnecessary thin layers which could be avoided? In my experience this is mostly the case in situations where a feature somewhere on the perimeter causes the entire layer to shrink. I think this can only be avoided by manually relaxing the height limitation, heuristic solutions would fail in many situations. Currently, the layer height can be partly overridden by the ``Layers`` setting, but I agree that this is not very intuitive. This idea has already been [discussed](http://dx.doi.org/10.1108/13552540010337074) more than 15 years ago, but something like a spline based input could actually improve both user interaction and internal computation.

I plan to continue working on this, but it might take some time.

platsch:
@lordofhyphens thx. I'm a bit confused. Why would pulling your branch make future rebasing easier?

platsch:
@ehybrid I think the slicing core itself is not affected, starting with the perimeter generation is probably the best idea. 
I suggest you fork the code if you are planning to work on this and we move the discussion to a new thread with focus on this feature?

bubnikv:
>Do you have good examples where the visibility of triangles or general object complexity causes unnecessary thin layers which could be avoided?

I think it is a personal taste of which features have to be smoothed out by thin layers and which not.

> In my experience this is mostly the case in situations where a feature somewhere on the perimeter causes the entire layer to shrink. I think this can only be avoided by manually relaxing the height limitation, heuristic solutions would fail in many situations.

Exactly. What is needed is some kind of interactive layer thickness editing with live preview.

alexrj:
So, from what I see there are two features:

1. algorithm should automatically generate a smooth curve for transitions according to heuristics;
2. GUI should allow users to modify that curve.

@bubnikv, you were saying that implementation by @platsch is limited because only evaluates slopes locally. Do you think you can go further?  Or do you think it's not worth to improve the algorithm by @platsch, as user would be able to modify its output?

lordofhyphens:
@platsch did you ever publish this somewhere? Reading a copy of the paper
may help @alexrj.

I apologize if I seem more dense than usual, currently traveling and don't
have access to my usual setup.

bubnikv:
> @bubnikv <https://github.com/bubnikv>, you were saying that
implementation by @platsch <https://github.com/platsch> is limited because
only evaluates slopes locally. Do you think you can go further? Or do you
think it's not worth to improve the algorithm by @platsch
<https://github.com/platsch>, as user would be able to modify its output?
We are currently investigating the interactive approach. We have some
prototype of the UI and now we are evaluating the usability of the UI and
the feasibility of printing process. There seem to be dragons in the
printing itself.

After we have the interactive process ready, I think then it will be a good
time to work on the automatic solution.

platsch:
@lordofhyphens not really. It was part of my master [thesis](https://tams.informatik.uni-hamburg.de/publications/2014/adaptive_slicing_wasserfall.pdf), but I've never translated / published it elsewhere.

@bubnikv is your experimental version public? I'd really like to play with it.
I think a curve based UI, "seeded" by the heuristic height estimation is definitely worth some work. In fact I was planning to implement this in January (no time to do it this year), but it makes no sense to do it twice.

lordofhyphens:
https://bintray.com/lordofhyphens/Slic3r/download_file?file_path=slic3r-adaptive-slicing.2017.01.13.326.194cb24.zip for people wanting to play with it; it's been built with current master merged in as well.

ehybrid:
You tested the adaptive Lord?  I'm trying it now and no matter what I set cusp to I don't get any variation on layer height.    I'm probably doing something wrong...

lordofhyphens:
I don't test it, I just make sure the builds roll off of the server.

You can build from source yourself to check.

ehybrid:
@platsch I'm testing the adaptive code from lordofhyphens recent build (ty mate, builds are very convenient)...   it's not behaving like I've seen screenshots, but I have managed to get it to vary layer height, but it does it in really weird places...  I'm looking at the source code, am I right in reading that you for each facet that intersects the build-plane compare the surface normal of that facet with the axis of the build plane to determine if it's necessary to shrink the layer height to maintain a maximum stair-step (is this what you mean by cusp height, or am I misinterpreting)...   (For example, testing wtih 'boaty' the top surface which is only at a few degrees off the build plane doesn't get reduced to the minimum layer height even if I set cusp very low)

platsch:
@ehybrid you are referring to this [model](https://www.thingiverse.com/thing:763622)?
I just tried it with the binary provided by @lordofhyphens, works as intended, the top surface gets reduced to the minimum possible layer height. What set of parameter did you use? Are you aware that you have to set the min/max value for the nozzle you are using in the ``Printer settings`` tab?
Example: if I set both the min limit and the cusp value to 0.1, the top-surface layers are reduced to 0.1mm.
Your description of the algorithm is correct. It does some additional steps to achieve correct results for cases where the angle is changing in the range of the current layer, but that's implementation details.

forReason:
I would Love to see this feature in the future.
I've come up with an Idea for a GUI solution and an Ideo for the mentioned Problem of gradient layer thickness.

I don't know about the current approach. This also may mean a lot of codechange:
1. Evaluate layer thickness
2. Normalize it over a Gradient with like 5% of total hight setps
3. Slice and enable a gui for manual layer height change.

gui idea

here you can see A possible Gui. the gradient is over the full height. It is just a sketchuo and not representative right now as I just mocked in an Image I found online. It would be possible to adjust the Layerhight by dragging the dots to the right or left which snap to the "gradation steps" or type in a precise value

Hope this makes sence

platsch:
@forReason Thanks for the input, in fact I just created something very similar, you can check the current status at https://github.com/platsch/Slic3r/tree/adaptive-slicing-spline

I'm still messing around with a lot of details, but this is good occasion to push out a first version and ask for feedback. Expect a better description later this day.

forReason:
Problem Is: I am not able to compile with pearl as my set-up is somehow broken so I need a precompiled build

Patola:
I am trying to compile the adaptive-slicing-spline branch but it is erroring out on me, @platsch 

```
Successfully installed Slic3r-XS-0.01
1 distribution installed
t/adaptive_slicing.t ........... 1/8 
    #   Failed test 'second layer height.'
    #   at t/adaptive_slicing.t line 36.
Use of uninitialized value $got in numeric gt (>) at (eval in cmp_ok) t/adaptive_slicing.t line 38.

    #   Failed test 'horizontal facet matched'
    #   at t/adaptive_slicing.t line 38.
Use of uninitialized value $val in addition (+) at /usr/local/share/perl/5.22.1/Test/Builder.pm line 727.
    #     undef
    #         >
    #     '0'
    # Looks like you failed 2 tests of 3.
t/adaptive_slicing.t ........... 7/8 
#   Failed test 'shrink to match horizontal facets'
#   at t/adaptive_slicing.t line 111.

    #   Failed test 'second layer height.'
    #   at t/adaptive_slicing.t line 36.
Use of uninitialized value $got in numeric gt (>) at (eval in cmp_ok) t/adaptive_slicing.t line 38.

    #   Failed test 'horizontal facet matched'
    #   at t/adaptive_slicing.t line 38.
Use of uninitialized value $val in addition (+) at /usr/local/share/perl/5.22.1/Test/Builder.pm line 727.
    #     undef
    #         >
    #     '0'
    # Looks like you failed 2 tests of 3.

#   Failed test 'widen to match horizontal facets'
#   at t/adaptive_slicing.t line 119.
# Looks like you failed 2 tests of 8.
t/adaptive_slicing.t ........... Dubious, test returned 2 (wstat 512, 0x200)
Failed 2/8 subtests 
```

platsch:
@Patola you are right. The fails are caused by "smoothing" the layers after computing the adaptive layer heights. I didn't yet update the test cases since I'm still experimenting with different solutions. It should still run correctly.

Patola:
@platsch Ok, I compiled it, but where is the interface to change the gradient? The only difference I saw from regular slic3r at the moment were the [X] use adaptive slicing, cusp value [ ] and [x] Match horizontal surfaces.

platsch:
There should be a new button "Layers...", next to "Settings..." on the Plater. Try this :-)

corthiclem:
@platsch Hi, I have troubles to run it too: perl crashes when starting Slic3r. 
Anyways, could you send some screenshots of this new feature for our viewing pleasure? :)

Patola:
Got it!!!! Show screenshot. AWESOME!
slic3r_540

forReason:
... is that hust a heap of poo you tested it with? :D

Patola:
Yes... And it is also customizable! A thingiverse design I made a few months ago: http://www.thingiverse.com/thing:1872629

platsch:
As announced earlier, here comes a slightly better description.
This [branch](https://github.com/platsch/Slic3r/tree/adaptive-slicing-spline) currently contains some experimental features to control the layer height, I plan to merge it into the [adaptive-slicing branch](https://github.com/platsch/Slic3r/tree/adaptive-slicing) after collecting some feedback and further testing.

The idea is to plot the layer height distribution over the height of the object and to manually manipulate this "curve". The original distribution is plotted in gray (as reference, screenshot 1). The current set of layers is green and the preview (during a manipulation) is red.
The layer height distribution gets initialized with equidistant layers (without) or with variable layers (with adaptive slicing) and applies B-Spline smoothing and interpolation.

Manipulation is done by clicking into the plot and dragging the mouse

- left/right: for the desired layer thickness
- up/down: for the range this operation should affect.

A left click initializes a "quadratic" modifier curve (screenshot 2), a right click selects a range for linear modification (screenshot 3).
The transparent plane indicates at which height the user is currently modifying the layers (y-position of the mouse at the plot).

I currently don't have pre-compiled binaries, @lordofhyphens would you mind cloning it for the build server?

spline_plain
spline_quadratic
spline_linear

lordofhyphens:
I've created a new branch based off of master. There's a single conflict in xs/MANIFEST, fyi. 

I can fix it locally, but it will probably be an issue when you submit the PR.

lordofhyphens:
If that gets sorted out you can have Windows builds easily.

platsch:
yup, I was still doing the merge when you cloned, didn't expect you to be so quick :-) update comes in a few minutes.

lordofhyphens:
Open up a new PR on my fork when you're done; that should kick off a new win32 build. It'll show up in the Slic3r_Branches bintray folder.

lordofhyphens:
It should have shown up in https://bintray.com/lordofhyphens/Slic3r/Slic3r_Branches, and the logs indicate that it was uploaded, but the branch name was incorrect for the artifact. I currently don't know what's going on. 

Next time you make a commit, could you open a new PR against my fork? I'll leave it open this time.

lordofhyphens:
https://ci.appveyor.com/project/lordofhyphens/slic3r/build/1.3.0-adaptive-slicing-spline-346 should be it as well. So weird.

ehybrid:
As per the image below, i've set min and max layer heights to 0.08 and 0.32 respectively, and notice the shallow slant of the roof just below the layer I've selected has been set to maximum layer height...  Shouldn't the cusp value on that roof show large differences compared to the surface geometry and therefore get reduced to 0.08 to minimise the top surface stair-stepping?

Also, I don't understand as it seems lower in the model it's using minimum layer height even when there's very shallow over-or-underhangs which have very minimal surface deviations even for large layer heights...

https://i.imgur.com/AzpbCOx.png

Is there a better way to visualise how you intend this 'cusp' value to work, I'm assuming it's my lack of understanding here at this point rather than anything to do with the code...  No idea if colour coding the geometry that causes the cusp value changes would be possible or not with the current framework, but it would be useful to see into the algorithm...

Awesome interface to edit the layer heights, and very speedy build... Thanks muchly :)  

Feature-begs:-
1, I'd really love to be able to quantise the layer heights so I can guarantee full-step layers only (eg, on my i3 400steps/mm @ 16x microstepping = 25 steps/mm = 0.04mm/full step...   interface would ideally take in 0.04 and then only allow layer heights between min and max that are multiples of 0.04)

2, I'd like to see a linear slider so the auto-computed heights could be all moved simultaneously linearly left or right in addition to your more complex spline-editing mechanism... most of the time the auto deviation cusp thing (should) be fine and all that's needed is to adjust the time/accuracy tradeoff over the whole model, basically shifting the whole curve left or right and cropped to the printers min/max values.

forReason:
@platsch your Screenshots look very promising can't sit to test it out. However I consider red a bad choice with 2 reasons:

1. Red is widely known as "error" value,  users might think their values are bad or so 
2. Colorblind person's will have trouble to differentiate between red and blue and as far ss I know every 4th man is colorblind.  Consider using blue or another color instead :smile

lordofhyphens:
https://bintray.com/lordofhyphens/Slic3r/Slic3r_Branches/1.3.0-adaptive-slicing-spline-346#files should be a win32 build.

platsch:
Ok, I overlooked a remaining merge issue: the layer indication plane was not rendered because it changed upstream. Fixed with current commit / PR.

Regarding your feedback:
@ehybrid the rooftop in your image is in fact not sliced with high layer thickness. Unfortunately, the slider for scrolling through the visualization does not exactly correspond to the layer plot. The reason is, that the slider uses a set of equidistant positions, but the corresponding layers are not equidistant. One slider-tick can change the height by 0.08-0.32mm in your example. Please check out the commit I've just pushed to have an indicator that corresponds to the actual position, think that's helpful, though we should still find a better solution to avoid the misleading slider interpretation.

I still have a hard time to easily explain the meaning of the cusp value. Yesterday I was playing around with a second slider to set the cusp value right in the layer height visualization, but can't get it working and it doesn't convince me anyway. Technical problem is: I try to set the cusp value from the LayerHeight window (the config value in `print->object->config`) to affect only the current object, but the value remains unmodified if I read it during the slicing call in `print->object.pm`. The config system is somewhat complex, suspect the values are somehow copied for each object and I don't use the correct object or so.

This picture explains the cusp value: you are basically setting the cusp vector (cn). The algorithm assumes that each layer has a perfectly vertical boundary and reduces the layer height until the distance between printed- and part-surface is lower than the cusp value. One important thing to now is, that the layer thickness well never be reduced below the cusp value, even if your min-limit is lower.
grafik

Layer gradation is already on my list, moving all heights simultaneously could be an interesting addition.

@forReason Right. I used green to indicate the final outcome and red as "that's what you are currently doing, pay attention". Let's see if I can avoid the green-red combination while maintaining a high contrast.

forReason:
@platsch as already mentioned, blue or purple is a common replacement for red in colorblind modes ;)

platsch:
@forReason the console output should be fixed with the latest version with commit: https://github.com/platsch/Slic3r/commit/dc4ea7dbc2f806bed2e469cc5a7168c4668f85f5, are you using the latest appveyor binary https://ci.appveyor.com/api/buildjobs/a04w5tesaq7s425d/artifacts/slic3r-master.2017.01.27.350.95d1435.zip?

If the window is still unresponsive with this version, which OS? I'm currently unable to test on a proper windows, only limited Win7 in virtualBox env. no macOS.

ehybrid:
Small bug, when switching from adaptive back to regular, the 'preview' window stops working.  (unsure if the 'layers' dialog needs to have been opened or not, I did)

ehybrid:
Also, it should probably called something other than "layers" as there's already a layers dialog under 'settings' ..  perhaps 'adaptive layers' would be better  (let me know if this is the 'annoying' rather than 'helpful' type of bug report)

platsch:
@ehybrid can you provide more details? I can not reproduce the preview window problem.

The name `Layers...` for the button is indeed ambiguous, it is not obvious why there is a Layers tab, a print setting Layers and this dialog. Plus: there is the additional layer height table in the Settings... window.

I have the following thoughts for better integration:

- Moving the layer height control to the Preview tab or (better) creating a new tab with a preview approach better suited to visualize the varying layer heights
- Do we need both, the table based layer height setting and the spline based control?

I generally appreciate any feedback, if it is just a technical bug, you could consider reporting it as an issue to my repo https://github.com/platsch/Slic3r/tree/adaptive-slicing-spline to keep this discussion focused on "what should be merged".

ehybrid:
Maybe it was a specific model triggering it, i've tried several and can't get the same behaviour.  I'll be more cautious if I get it to happen again...  Is there an easy way of getting a function call trace or similar to help narrow it down?    It was repeatable the other day as it occurred 3 times in quick succession (could potentially be made worse as my computer was giving me out of memory warnings as firefox had misbehaved at the same time as I recall)...  I had a couple of strange crashes as well in the process of changing settings while it was processing in the background, but I couldn't get it to recur in my testing.  Gotta love intermittent bugs, right?

I really like the idea of moving it to the preview tab, it's fundamentally the same thing as the preview tab anyway, always having the layer-height setting dialog would also be pretty convenient  (the default UI of clicking layers, dragging to find layer heights, then going back to the 3d window, right click, settings, layers to change the layer height of sections of a model manually is a huge pain).  your spline editing thing is a really great start towards a more general and practical UI to editing layer heights even without the adaptive calculations having been done.  I don't think the spline editing thing as it stands is accurate enough to be able to entirely replace the layer editing stuff, but I don't think it needs a lot of work to be able to add some quality of life tools, for example, being able to select a region by shift-dragging the layer slider, and then entering a thickness make the other interface entirely redundant.

platsch:
@ehybrid I don't really like the idea of integrating it into the 3D preview tab, because the layer settings are done per object while the preview window shows the full toolpath for all objects. That's why I currently open a new window for one specific object (the 3D-preview still shows the full toolpath, quick solution to make it usable before implementing something more final).

ehybrid:
Good point, hadn't thought of that...

Perhaps it could be solved with a selection box to choose between
all-objects and each individual object on that tab?  From a functional
point of view it makes no sense to have the same function in two places
depending if its one object or all objects.  Perhaps me thinking of the
layers tab like this is broken though and its the 3d page that needs some
object layer height workflow improvements rather than trying to stuff those
into a preview tab.

I tend to have a workflow of:

1 Drag objects to 3d window
2 guess initial slicing settings in the print settings tab
3 switch to layers tab and inspect the layers for voids ovwehangs and
smoothness
4 edit settings on either printer settings or by switching back to 3d tab
to the same object and right clicking to edit specific object settings
5 repeat 3 and 4 until satisfied

... I guess im arguing for a ui that makes that repeated section more
convenient...  When I'm dragging through the layers not having to switch
tabs to get into an objects settings would save at least 2 mouse clicks per
iteration.  Changing the 3d page so people could see how its sliced and
having the settings stay on that page would also work and make the preview
page entirely redundant.

I've been debating if it makes sense to try and cloud parallelize this,
have 50 slightly different perimeter widths that all get simultaneously
sliced in the cloud and screenshots sent back so its really fast to scroll
back and forwards on each layer to find the one that gives the best packing
density.  Perhaps adapting something like the variable slicing algorithm to
the perimeter width to minimise voids where perimeters touch is a better
solution in this case (bonus points if this can be modulated in a single
layer, eg all my perimeters are 0.5 out of a 0.4 nozzle except where I have
a 0.9mm thin wall in my model, and there and only there my extrusion width
drops to 0.45 (plus a little for good adhesion)

ehybrid:
crash report:-

http://www.thingiverse.com/thing:1069671
wand base
rotate z 45 (note it rotates in opposite direction in 1.2.9)
generate support on
change 'overhang threshold' a few times and let it regenerate
within a few changes it crashes
has also crashed when merely enabling or disabling adaptive mode...

this bug may not be specific to the adaptive changes.  Tested in 1.2.9 and
no crash despite many more repeats of the same procedure.

corthiclem:
Hi everyone,

I am trying to use this Adaptive Slicing feature and I face an issue when trying to import two different STL (representing an Assembly) files with the settings window. 

Actually, I get errors that I cannot read from the bottom output tool.

capture

capture2
View of the settings window.

Does anyone else encountered this issue?
@platsch do you have any idea where does it come from?


I'm using Windows 7 and the last commit of the Adaptive Slicing Branch. 

Thanks for your work. It looks really promising.

platsch:
Hi @corthiclem 
good point. 
You can start slic3r in debug mode (perl slic3r.pl --debug) to see the full error message.
Looks like the second mesh is not correctly initialized by the AdaptiveSlicing method.
Fixed it in the original adaptive-slicing [branch](https://github.com/platsch/Slic3r/tree/adaptive-slicing) and in the spline [branch](https://github.com/platsch/Slic3r/tree/adaptive-slicing-spline).
Binary comes in a minute.

ceryen:
I wonder if the cusp height heuristic is sufficient for nearly horizontal surfaces.  For example, consider a surface that has a normal_z of 0.95 and a cusp height value of 0.1.  In this case, the surface would be nearly horizontal, so we would expect use of the minimum layer height.  However, the cusp height heuristic will calculate a layer height of cusp height value / normal_z = 0.1 / 0.95 = 0.105, which would probably be larger than the minimum layer height.  It seems to me that to get the minimum layer height in this example, the cusp height value would basically need to be set to the minimum layer height.  Is that how we're expected to set the cusp height value?

For comparison, the Varislice method uses a slope-based heuristic instead.  It calculates the slope for a surface, then multiplies the slope against a ratio to calculate a layer height.  So consider a surface with a normal_z of 0.95 and a slope-to-height ratio of 0.1.  In this case, the slope would be tan(acos(0.95)) = 0.329, so the calculated layer height would be slope * ratio = 0.329 * 0.1 = 0.033, which would probably trigger the minimum layer height for the surface as expected.

platsch:
@ceryen your observation is correct. The idea of the cusp approach was to define a maximum allowed error. For your example, the error would be 'small enough' for the nearly horizontal facet.

There also is a 3rd method of computing the height which was introduced by Pandey in a paper in 2003, where the layer height is computed based on surface roughness (essentially `factor * cos(slope)`), where `factor`describes the required surface roughness.

I think all 3 approaches have their up- and downsides e.g. for the varislice method there is no limitation for the layer height for nearly horizontal / vertical facets, i.e. a horizontal facet will always cause the minimum height, a vertical facet causes the maximum height, which kind of prevents us from using the control parameter as a "compromise between speed and quality".

I'm currently exploring a combination of 2 or 3 methods to make the control parameter more intuitive, but that also is a compromise and not optimal for all cases.
One question is: if you set the control parameter to a very low value (high quality, slow printing) would the user expect thinner layers even at vertical features? Both, cusp and varislice would generate maximum thickness layers (limited only by the printer capabilities). Same for a very high control value.

platsch:
@ceryen two more remarks:

- a normal_z of 0.95 equals a slope of ~72° (due to the sinus term).
- you can basically "cut" the lower boundary by choosing a cusp value < min_layer_height. cusp = min_layer_h converges exactly for horizontal facets.
And yes, that is not intuitive by any means :-\

lordofhyphens:
I think this will fix/address #1827, what do you think @platsch

bubnikv:
FYI, we have released a beta with our implementation of the manual variable
layer height editing.

https://github.com/prusa3d/Slic3r/releases/tag/version_1.33.3

You are welcome to reuse whatever you want from it. The OpenGL shaders are
very effective for real-time simulation of the layering.

platsch:
@bubnikv Huh, nice interface. I really like the height preview in the 3D window, would love to integrate that...

would be great if you could modify the general layer height right in the control element.
Did you just disable the old table based thing? seems to be not working in your version.

bubnikv:
> would be great if you could modify the general layer height right in the
control element.
> Did you just disable the old table based thing? seems to be not working
in your version.

I did not disable the old table, but I did not want to feed it with the too
many points of the spline manipulator, so I keep a second table. If the
dense table is populated, then the sparse table is ignored. If you fill in
the sparse table first and then open the interactive tool, it will be
initialized from it.
It is quite possible though that I omitted something and it does not work
as intended. Please provide an issue into the issue tracker :-)

This release is a beta, so there will be sharp edges.

bubnikv:
@platsch 

I did some changes to the interactive layer editor recently. They are in the official binary release as well:
https://github.com/prusa3d/Slic3r/releases

Namely, I added a help overlay and a reset button, and I implemented level of detail, which hides the layer strips if the camera is far from the model. I have also implemented quite extensive debugging for the OpenGL compatibility issues.

You are most welcome to build upon it. There is a SlicingAdaptive.{cpp,h} skeleton, which contains your original automatic layer height code rewritten to C++, hopefully without introducing new errors :-) In the current state it is not usable to the public, it needs a lot of filtering, maybe based on visibility of the features from the outside, maybe on their size. I don't currently have time for that, there are more pressing issues in Slic3r IMHO.

It would be great to extend the layer editing bar with some form of legend, maybe a tooltip, which would display the current layer height below the cursor.

> would be great if you could modify the general layer height right in the control element. Did you just disable the old table based thing? seems to be not working in your version.

The two features work separately: You either use the table or the interactive editor. The values painted in the editor are not reflected back into the table as the table would be too big and useless.

If you enter the height values into the table, they will be used until you open the layer editor. Then the layer profile will be initialized from the table and since that point, the layer graph will override the layer table. You may revert to the layer table by clicking the "Reset" button on the layer editor.

Since this feature is fresh, I admit I never tested this scenario thoroughly. I did now and it mostly works with a single glitch: When the layer profile is initialized from the table, the model is colorized correctly in the layer preview, but the graph at the vertical bar is not updated until you do the first edit of the layer height profile. I should fix that glitch.

Other than that, I think the behavior is reasonable.

platsch:
@bubnikv that's good news!
I just check out your code yesterday, no time over the weekend, but next week I definitely see if I can do some integration.

ehybrid:
Another suggestion, the top and bottom 'solid layers' should probably be measured in mm rather than 'layers' for variable slicing so people can get known thickness shells on their models

whosawhatsis:
Actually, top and bottom solid layers should have minimum values in both mm and layers, so that whichever number is greater will be used.

lordofhyphens:
Hey @platsch I moved the build server over to pull from this repo, not mine. 

I think I have it working with pull requests, so if you add commits here the hooks *should* pick it up.

platsch:
@lordofhyphens I just opened a PR to your repo for an updated version to trigger the build process. This is not a final version, but I'd like to have a base for further discussions.

The current state is:

- The basic adaptive algorithm is integrated into the prusa interface, PR: https://github.com/prusa3d/Slic3r/pull/173
- The suface analyzer is ported to C++, but currently not usable since the `slice` method in `Print/Model.pm` is still perl. I could create an XS interface, but would prefer porting the `slice` method to C++. We could use the prusa3d code as a start, I would open a separate pull request for this if you are interested.
- I see two ways of UI integration: we could try to use the prusa3d interface for layer height manipulation, I'm not sure how much effort this is. Or I finish the spline based control and do a better integration into the 3D tab, probably by using the shader approach for a per-object preview. I personally prefer the toolpath preview, because it provides a better feeling of how the print comes out, but it doesn't allow selection or preview of a single object. Superimposing the toolpath with a transparent model might be a solution?
- Lots of cleanup still open.

Opinions?

lordofhyphens:
How much effort depends on how tightly @bubnikv integrated things into his UI side. If it's another window that is its own package, it shouldn't be difficult to merge in. 

Go ahead and open a separate PR for the slice method.

lordofhyphens:
I'm not sure if this got mentioned or not in this long, long thread, but it is important that there is an option to set the quanta used for adaptive slicing (that is, the smallest +/- increment used by the algorithm to fit). 

Several printers have weird full step heights (one of them is the Monoprice Select Mini) and don't do well when Z is not one of those full steps.

platsch:
@lordofhyphens @bubnikv s UI is fully entangled with the GLCanvas preview window. I prefer a separate WX element for the control, as it is now. That can be then placed next to the preview element with a vertical sizer. However, the texture stuff must be part of the 3D element obviously.

The gradation option is intended for the limiting the printer to certain z-positions. That should work, but is not extensively tested.

bubnikv:
> @bubnikv <https://github.com/bubnikv> s UI is fully entangled with the
GLCanvas preview window. I prefer a separate WX element for the control, as
it is now. That can be then placed next to the preview element with a
vertical sizer. However, the texture stuff must be part of the 3D element
obviously.

I admit the implementation is not the most elegant, but the customer does
not care. Through the years of professional software development I have
learned, that a product has a limited life span,and that an optimization
aka gold plating of an implementation, which has no impact on the customer
experience, only slows down the development of features, which he will
actually benefit from.

lordofhyphens:
@platsch so long as it will apply cleanly as a PR and it's not too hackish, I don't have much issue with the UI being entangled as in the Prusa3d branch.

lordofhyphens:
https://ci.appveyor.com/api/buildjobs/wypaif3a28jokd8t/artifacts/slic3r-master.2017.03.07.457.82581b3.zip appveyor build btw.

lordofhyphens:
@platsch, @alexrj would prefer a modal editor, similar to how we have the Cut and Object Settings editors at present. I *think* you can copy/reuse the existing preview code.

lordofhyphens:
@platsch #1827 just got done, how will it interact with your PR?

platsch:
@lordofhyphens I'm on vacation, expect an answer next week :-)

platsch:
@lordofhyphens It probably won't interact with the adaptive stuff at all, since it only affects the static layer height. But I can use the config-value for the gradation feature, so the option has the same effect.

alexrj:
@platsch, I've been thinking about the things we discussed a few days ago. Invalidating and recalculating only a subset of they layers is doable but might require more work than expected, so I wouldn't do it now. I think when we open the dialog we only need to require the `posPerimeters` step to be calculated. Infact the dialog will not provide the slider for hiding layers (like the current toolpaths preview does), so we only care about the exterior to be visible. And we can provide real-time preview by just squishing and repeating existing layers at the visualization stage (thus in 3DScene). I recommend we proceed step by step, where first step would be just to color *existing* layers/toolpaths according to the configured rules without changing them, and second step would be to squish and repeat them.

platsch:
Just updated the PR. Current state is:
- everything except UI is ported to C++
- the interactive height dialog is moved to the settings window (the ``Layers...`` button is a shortcut)
- the spline object is stored at the model_object, but not exported to AMF-files yet
- preview shows only the toolpath of the selected object, no coloring yet
- lots of minor changes and refactoring

I added a new processing step ``posLayers`` to distinguish the generation of layer heights from the actual slicing step.
Any feedback / comments / questions?

corthiclem:
Hi @platsch, I use your branch "adaptive-slicing-spline" a lot and I really enjoy the work you've done: This is great!

Few comments:
1. As you mentioned, saving the layer height profile into the AMF file (I believe the prusa fork does so) is in my opinion, one of the most important missing feature to your work.
2. When a raft is used, the transparent/white plane does not indicate the correct layer. Actually, it does not take the number of raft layers into account so the white plane is shifted and misplaced. For example, it never reaches the top of the object:

untitled


Thank you again for your work

platsch:
@corthiclem thanks for the positive feedback!
Saving to AMF files is on the list, but I wanted to collect some feedback first.
2. is a good point, I very rarely use a raft so I just missed that problem. Should be fixed now.

lordofhyphens:
I don't think we'll get enough eyes on this in a PR to push it forward to be 100%.

I propose pulling this into master at the next opportunity so as to solict more feedback.

platsch:
@lordofhyphens I agree, it doesn't look like there will come any more feedback here.
I think the scaling problem should be fixed before merging, we don't need any external feedback for that but I'd like to have an opinion from @alexrj. Is it useful / necessary to have two separate scaling mechanisms? I don't want to adapt to both...
As far as I remember, either the mesh is scaled or the scaling parameter in the per_instance object, depending on whether the scaling is applied to the entire object or per-axis.

ehybrid:
Is there anything in particular you would like fed-back?  Feature is awesome...

platsch:
@ehybrid from my perspective it is mostly about bug reports to cover more than the setups I frequently use, opinions on "is there anything we must definitely change before it is useful" and in best case code review.

corthiclem:
@platsch there is something in the Settings tab that might be confusing. 
When the adaptive quality is changed in the "Adaptive Layer tab", it creates an overidable "Adaptive slicing quality" in the Parts tab. (See screenshots below).

adaptive_slicing_0 
adaptive_slicing_1

It seems that changing from the Adaptive Layers Tab changes the input in the Parts Tab but changing from the Parts Tab doesn't change the Adaptive Layers Tab:

adaptive_slicing_2
adaptive_slicing_3

I think the user will be confused and won't know which tab gives the correct Adaptive Slicing Quality. 
What is the purpose of creating the overidable "Adaptive slicing quality" in the Parts tab? Which value of Adaptive Slicing Quality will be used when 'Ok' is clicked?

lordofhyphens:
@platsch seems to be missing the three icons for variable layer height.

VanessaE:
image

VanessaE:
Also, it crashes when I try to edit a print profile:

`No adaptive_slicing in ConfigOptionsGroup config at /home/vanessa/RepRap/Slic3r-alexrj/lib/Slic3r/GUI/OptionsGroup.pm line 369.`

VanessaE:
This is all under Debian 9.0 "stretch", with the adaptive slicing pull request applied on top of commit 9f0335b4, and compiled as follows:
```bash
#!/bin/bash

set -e

cd /home/vanessa/RepRap/Slic3r-alexrj/xs

eval $(perl -Mlocal::lib=/home/vanessa/RepRap/Slic3r-alexrj/local-lib)
export LDLOADLIBS=-lstdc++

perl Build.PL
./Build
./Build install
```

ehybrid:
I havent had a chance to test this merge, one of the showstoppers for me
previously was the interaction with the flow rate limit code that caused
many many layetd to slow down to single digit mm\sec rates....  Without
that, setting linear speed limits result in thin layers extruding very
small volumes of plastic with a significant impact on overall print time
(some were 10x longer) ...  I'll retest it, but I suspect it's still an
issue as even without variable slicing I'm still getting major slowdowns
with the volume limit set.  I know there was an interaction with the gap
filling code that made this way worse, but I don't think its the only issue
as I was seeing extremely slow layers even with gap filling off.  If the
extruder can maintain its volume of plastic by speeding up on thin layers,
the variable slicing should only increase print times proportional to the
amount of extra acceleration events.

VanessaE:
Meanwhile (since I can't check due to the above crash), is there a setting that'll let me limit how thin the layers are on overhanging parts?  I ask because no matter what I do, I can't really get a good overhang beyond about 50 or 55 degrees from the horizontal if I use 0.1 mm layers, depending on how much more there is to print above the overhang.  At 0.2 mm layers, I can get 85 degree overhangs with ease.

Consider this simple drawing (a variation on the 30-to-80 degree arch overhang test on thingiverse):
image
*("Please excuse the crudity of this model.  I didn't have time to build it to scale or paint it")*

In zones 1 through 6, I would expect the adaptive code to start out with ~0.35 mm layers, using gradually-decreasing layer heights through zones 1 through 6.  However, when it gets to zones 7 and 8, layers would have to remain thick enough to accommodate that 45 degree overhang, while zone 9 would be thin layers again.

VanessaE:
(this is a re-post because I made a few errors)

Meanwhile (since I can't check due to the above crash), is there a setting that'll let me limit how thin the layers are on overhanging parts?  I ask because my setup has some limits:  when printing with 0.1 mm layers, I can't get good overhangs beyond 55 or so degrees, depending on how much the plastic curls up and how much more "stuff" there is to print above the overhang.  At 0.2 mm layers, I can rock out an 85 degree overhang with ease.

So, consider this simple drawing (a variation on the 30-to-80 degree arch overhang test on Thingiverse):
image
*("Please excuse the crudity of this model.  I didn't have time to build it to scale or paint it")*

In zones 1 through 6, I would expect the adaptive code to start out with thick layers (say 0.35 mm, since I have a 0.4 mm nozzle), using gradually-decreasing layer heights as it goes up.  However, when it gets to zones 7 and 8, although the program will want to use thin layers to make the top-side slopes on the right look nice, the layers would have to remain thick enough to accommodate the overhang on the left (assume the overhang is much steeper than in the picture).  Zone 9 would return to thinner layers again since there's no overhang there (just vertical and top sloping surfaces).

VanessaE:
Another glitch:
1. Load some model, preferably something complex but small enough to slice quickly.  
2. Click on the model, then open the Layer Heights dialog.  Let it generate the preview.
3. Close the dialog, and go change some relevant setting elsewhere in Slic3r, such as turning support material on.
4. Re-open the Layer Heights dialog.  No preview will be displayed -- you must twiddle the quality slider to get it to generate a new preview.

VanessaE:
As an aside:  I like the highlight on the sliced result when you mouse over the layer heights graph, but I think it would be a good idea if you also displayed the height of the currently-highlighted layer somewhere.

platsch:
Yeah, lots of feedback :)

@corthiclem it should work in both directions, but the values are updated only after closing and opening the dialog. I just use slic3rs config mechanism, the slider in the layer-preview tab changes the overridable per object (which makes sense, since we want to tweak the parameter per object) and Slic3r automatically adds the modified value to the object settings section. But I agree that instant synchronization would be desirable.

@lordofhyphens ahm, yes. Fixed it.

@VanessaE regarding your overhang-feature request: I was discussing the same idea for some time with a few colleagues. We came up with two opinions:
* It is useful to implement this as an option and limit the layer thickness below a certain negative threshold similar to the support structure threshold.
* There are too many situations and exceptions where this wouldn't help, just let the user modify the height distribution manually.

I will probably try to implement this as soon as I find some time.

For the preview update problem: I experienced the same, but had the impression that it is not related to adaptive slicing since the Preview tab also didn't update. The UI update and changed background processing logic caused some trouble, but I'm not able to reproduce this since the last merge.

Current mousover-layer-height is now displayed.

platsch:
FYI: we just published a [paper](https://tams.informatik.uni-hamburg.de/publications/2017/Adaptive%20Slicing%20for%20the%20FDM%20Process%20Revisited.pdf) on this topic at the 13th IEEE Conference on Automation Science and Engineering ([CASE2017](http://case2017.org)). Thx for a lot of inspiration that I got in this thread.

lordofhyphens:
Congratulations on the conference paper. I know those can be quite a pain to get into a good conference. Any plans for a longer journal article?

platsch:
@lordofhyphens currently not. I think the main ideas are documented now, getting a good journal article published is a lot of extra work and I'd prefer to spend this time on the electronic printing project (or on getting the adaptive feature merged...).

lordofhyphens:
@platsch It's merged; I had to resolve a minor merge conflict (the changes for 3MF happened to hit similar lines). 

Thank you, and I'll be sure to tag you when people write up issues about it ;)

platsch:
@lordofhyphens Great! So I better prepare myself for the wave of issues and bug reports ;-)

Patola:
hooray!!!! This is great!!! Thank you @platsch and @lordofhyphens !!!

lordofhyphens:
@platsch would you be willing to write a description and a brief how-to guide for the Manual? I promise it won't take nearly as long to merge ;)

platsch:
Yes, of course I'm willing, good point. I also found 2 bugs in my implementation and AMF support is still on the todo. It's merely a time issue...

lordofhyphens:
@platsch http://manual.slic3r.org/expert-mode/variable-layer-height you can extend this page

curiouspl2:
+1 on the overhangs problem, as in https://github.com/alexrj/Slic3r/issues/2120

i think this could be simply implemented as running your code as usual, and then running detect overhang pass to detect overhung layers .
it is great You did implement the 'quality' slider and i think same system could be used for deciding about overhangs :)

it would be great if user could set up at least two thresholds, like up to certain % of overhang actually lower layer height looks better so this could be 'overhang lower threshold'
then after this threshold increasing layer height helps instead of decreasing. 
it would be great that another range could be selected for boosting layer height for overhangs , 
with default of sth like 20-40%.
then final range for 40-100% would be nice with separate boost value. 

Then also i like the Layers... dialog with the graph, and with all those three sliders there user could quickly set the model up , especially if overhangs could be also painted on graph for visual cue.

PR #1551, opened 2013-11-24
Potentially fixed contact infill, spacing, and the too much contact.

cakeller98:
See what you think, i've done the best I can to implement what I believe is an improvement to the contact infill.

I haven't addressed the fact that not using contact has the support touching the part, but that is another issue, I think.

Also there is an issue with floating islands not being suported unless raft is turned on to 1 layer.

With these commits and a few test files, I've gotten significantly better results that I was getting, but it will need testing.

However I am optimistic this makes some good points, though I'm sure they could be coded better - my apologies in advance for any horrible hackery.  I don't know perl, and she doesn't like me much.

And for what it's worth, I've spent a lot of time going over how your support algo works I really hope I put everything in the right place.


alexrj:
Okay, most of the things in here have been addressed meanwhile.

PR #1901, opened 2014-04-04
Allow for both nozzles to be purged every layer

tjmustard:
While this is not an absolute fix, as more than 2 nozzles would not be supported, it would purge the two first nozzles every layer.

A better fix would to send nozzle[i] to skirt immediately before printing out nozzle[i]'s support/perimeter/infill/etc. This would purge the nozzle immediately before you print to the main model.

ie:
for layer i:
  support nozzle[1] -> skirt nozzle[1]
  print skirt nozzle[1]
  print support nozzle[1]
  perimeter nozzle [0] -> skirt nozzle[0]
  print skirt nozzle[0]
  print perimeter nozzle[0]
  print infill nozzle[0]
  i++

Not sure how to do this, but it would be a great addition.

Thank you,
TJ Mustard

### DIFF
--- a/lib/Slic3r/GCode/Layer.pm
+++ b/lib/Slic3r/GCode/Layer.pm
@@ -89,6 +89,14 @@ sub process_layer {
             for my $i (0 .. $#skirt_loops) {
                 # when printing layers > 0 ignore 'min_skirt_length' and 
                 # just use the 'skirts' setting; also just use the current extruder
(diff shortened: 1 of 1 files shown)


lordofhyphens:
As we have custom gcode for every layer now, I think this PR is not necessary anymore. Can it be closed @tjmustard ?

tjmustard:
Yeah.

PR #1973, opened 2014-04-27
Configurable layer speeds per object

jiripech:
Hey guys! I wanted to have a possibility of slowing down (or speeding up) the print for the whole layer within one of the objects on my build plate. It came to my mind today (after five failed prints) that it may be easily doable in the _Object Settings / Layers_ configuration dialog, so I simply added a column to the dialog.

slic3r speed screenshot

The Layer object needed to be extended by adding the speed attribute to it. And this attribute is then used to alter the speed in the GCode. Simply by dividing the standard speed by this user defined percentage.

As a bonus, I added check for the overlapping layer ranges, because there was a TODO section in ObjectSettingsDialog.pm waiting for it.

Please note, that I've never done anything in Perl and I'm also fabulously lazy to read any manuals, coding standards and such a kind of stuff. Please kindly tell me if anything is wrong with this little piece of change and I may try to do it better next time. Otherwise, I'm glad enough if everyone enjoys this new feature. Even if you perhaps decide to choose another implementation scenario.

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,6 +257,8 @@ The author of the Silk icon set is Mark James.
         --support-material-enforce-layers
                             Enforce support material on the specified number of layers from bottom,
                             regardless of --support-material and threshold (0+, default: 0)
+        --dont-support-bridges
+                            Experimental option for preventing support material from being generated under bridged areas (default: yes)
     
        Retraction options:
         --retract-length    Length of retraction in mm when pausing extrusion (default: 1)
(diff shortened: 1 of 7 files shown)


lordofhyphens:
@jiripech with the improved modifier mesh setup this probably isn't necessary as it is.

jiripech:
@lordofhyphens Yes, I think so. Thanks for reviewing it anyway.

Issue #2068, opened 2014-06-02
Automatic threshold setting fails to create support when needed


harriv:
Support is not generated when needed when using automatic settings.

Files:

https://drive.google.com/folderview?id=0Bw1VY8X1WZhdbTNaMVlnWDdlMnM&usp=sharing

Command line: C:\Program Files\Repetier-Host\Slic3r-1.1.3\slic3r.exe --load "C:\Documents and Settings\harriv\Local Settings\Application Data\RepetierHost\slic3r.ini" --print-center 75,74 -o "C:\Documents and Settings\harriv\Local Settings\Application Data\RepetierHost\composition.gcode" "C:\Documents and Settings\harriv\Local Settings\Application Data\RepetierHost\composition.obj"

Using Slic3r 1.1.3 32 bit on Windows XP.

![slic3r_support_bug](https://cloud.githubusercontent.com/assets/2283563/3152640/ec990e24-ea90-11e3-94e0-884321fe4366.png)

PR #2098, opened 2014-06-16
Fix automatic overhang threshold

hyperair:
We should be supporting perimeters that are overhung further than half a
perimeter out, rather than two times the perimeter width.

Fixes: #2068

### DIFF
--- a/lib/Slic3r/Print/SupportMaterial.pm
+++ b/lib/Slic3r/Print/SupportMaterial.pm
@@ -152,7 +152,7 @@ sub contact_area {
                 } else {
                     $diff = diff(
                         [ map $_->p, @{$layerm->slices} ],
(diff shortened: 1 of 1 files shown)


lordofhyphens:
@hyperair  How does this work with the default spacing (which is one support width from the model) for support?

hyperair:
I remember it working fine -- the only thing it changed was whether or not some things got supported. But I got tired of keeping track of the diff, so I stopped using it altogether.

bubnikv:
I believe this fix is correct, though I may be wrong and I don't understand the reasoning of @alexrj.

lordofhyphens:
@hyperair could you make the width overhang (relative to perimeter width) configurable instead with a default of 0.5?

lordofhyphens:
I extended it in #3752. Closing this PR.

PR #2421, opened 2014-12-15
M140 S0 (Bed cooldown) integration into Custom G-code End Commands

blue-ice:
The change that I made in this pull request is to set "turn off the bed once the print is done" (M140 S0) into the default Slic3r Custom G-code End Commands.

When printing, I find it helpful to shut off the bed once the print is done, so that the parts can be removed. I added this command into my setup of Slic3r and find it useful, which is why I think it should be a default command.

Comments and thoughts appreciated.

blue-ice

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,9 +215,10 @@ The author of the Silk icon set is Mark James.
         --solid-fill-pattern Pattern to use to fill solid layers (default: rectilinear)
         --start-gcode       Load initial G-code from the supplied file. This will overwrite
                             the default command (home all axes [G28]).
-        --end-gcode         Load final G-code from the supplied file. This will overwrite
-                            the default commands (turn off temperature [M104 S0],
-                            home X axis [G28 X], disable motors [M84]).
+    --end-gcode         Load final G-code from the supplied file. This will overwrite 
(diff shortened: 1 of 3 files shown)


strahlex:
The M104 code does not work for all firmwares. I think this should be made firmware independent.

blue-ice:
What are the alternative commands for the other firmwares?

strahlex:
See GCodeOutput.cpp

strahlex:
It would be nice if we would have GCode flavour specific default begin and end GCode commands.

lordofhyphens:
There's too many different printers and firmware that slic3r has to support for this to be particularly useful. Even for ones with similar dialects they often behave differently, like M190, etc. 

Better to just put the defaults in a config and let the user set them up right for their printer.
I think this PR should be closed.

7eggert:
So if M140 is the most common command and assuming it does not harm other devices, you could write

; uncomment the correct line:
; M140 S0; Turn headbed off on most devices
; M0815 S0; Turn headbed off on $that_other_device

lordofhyphens:
Implemented as d8a743d (end gcode now acts like start gcode).

PR #2614, opened 2015-02-02
Added helper scripts to set local::lib before building on linux

lordofhyphens:
These should keep everything that cpanm wants to install neatly packed inside of slic3r's directory, as well as the XS lib.


lordofhyphens:
@alexrj this PR is convenient to not clobber system perl packages, especially on Debian based distros.

bubnikv:
I see a great usefulness of setting a separate site perl site lib dir on a build server for example. It would be even greater, if it could be done as part of the Build.PL script, configured by a command line switch. Or at least if the trick could be documented.

alexrj:
@lordofhyphens I guess this is not needed anymore, right?

lordofhyphens:
Wouldn't be. Go ahead and close.

PR #2623, opened 2015-02-03
OctoPrint SD card upload option

quillford:
Added SD card upload for OctoPrint users.

### DIFF
--- a/lib/Slic3r/GUI/Plater.pm
+++ b/lib/Slic3r/GUI/Plater.pm
@@ -7,10 +7,10 @@ use File::Basename qw(basename dirname);
 use List::Util qw(sum first);
 use Slic3r::Geometry qw(X Y Z MIN MAX scale unscale deg2rad);
 use threads::shared qw(shared_clone);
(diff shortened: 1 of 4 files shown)


lordofhyphens:
How does this differ from send gcode to printer that is already available?

quillford:
It allows the gcode to be uploaded to an SD card on your printer, so that if something happens to OctoPrint it will keep printing.

lordofhyphens:
That seems to be the case already with my octoprint install and the send
gcode

alexrj:
Thank you for the pull request. I have a couple comments: option names never have uppercase characters or camelCase style, so I'd rename that to `octoprint_loc` for coherence. Second: your text editor changed a lot of whitespace lines, so this ends up to a huge dirty patch which is not good.
But nevermind, I can hand edit the change and add the option myself.

I wish Octoprint had a default location setting and we could just call `POST /api/files` without storing yet one more config option...

quillford:
I fixed the variable names. Onto whitespace ;-)

quillford:
I don't really have a need for this anymore and it's pretty easy to implement if you want to do it, so I'll close this pr.

PR #2643, opened 2015-02-08
Revert the extra options to the creation of the icons for preset save

Lenbok:
and delete, which had resulted in the icons being virtually invisible
under linux. The icons now look good again:

image


Lenbok:
AR fixed in 13d1393ae44f

PR #2678, opened 2015-02-21
Bugfix: correctly draw circular beds in the bed-shape dialog

robstarling:
Wx::DrawLine wasn't happy with a polyline (encountered when changing to a circular build plate), so we pass the sequence of points to Wx::DrawLines


alexrj:
Well, thank you but I'm not pulling this because:

1. I don't want to introduce a dependency on List::MoreUtils
2. This is a regression of the Clipper library (recently updated to a newer version) that returns extra vertices in grid segments. So I'm escalating the bug report to the upstream library author and introducing a workaround to the crash in Slic3r.

robstarling:
No problem! Thanks for looking into it. Let me know if i can help.

PR #2707, opened 2015-03-04
Bugfix: fix incorrect comments to temperature-setting gcode.

mdebski:
Since porting gcode writer to XS (ee3fb7caa) comments on M190 and M140 (setting
bed temperature) were swapped. This change fixes them and also clarifies that
M109/M190 is "set and wait", not only "wait" (as opposed to M116).


mdebski:
@alexrj ping

PR #2716, opened 2015-03-07
Adding velocity extrusion support

strahlex:
This patch adds support for velocity extrusion to Slic3r. Velocity extrusion means that the extrusion rate of the extruder is purely controlled based on the movement of the extrusion nozzle and a specified line cross section. This removes the additional extruder axis if velocity extrusion is enabled. The advantage of this method is the trajectory planner is able to optimze the movement easier. Furthermore this should increase the resulting printing precision (no rounding in the extruder axis movement).

For this purpose this patch adds an additional configuration option `use-velocity-extrusion`. In this mode no extruder output is generated, but an additional M600 command is added when the line cross section changes. This mode is only supported in combination with firmware retraction. At the moment the only firmware that supports velocity extrusion is Machinekit.

In order to do my modifactions I needed to move the command that outputs G1 movements to the GCodeWriter as it should be. So this patch implements following changes:

- moving GCode output from ExtrusionEntity to GCodeWriter
- calculate extruder movement in GCodeWriter instead of in GCode.pm -> cross section (mm3_per_mm) and line_length is used as parameters
- added additional config option `use-velocity-extrusion`
- added the config option to the UI
- added `set_cross_section` function to GCodeWriter
- using the `set_cross_section` function in GCode.pm
- small corrections in the documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,7 @@ The author of the Silk icon set is Mark James.
         --use-relative-e-distances Enable this to get relative E values (default: no)
         --use-firmware-retraction  Enable firmware-controlled retraction using G10/G11 (default: no)
         --use-volumetric-e  Express E in cubic millimeters and prepend M200 (default: no)
+        --use-velocity-extrusion Express E in velocity mode using the M600 command (default: no)
         --gcode-arcs        Use G2/G3 commands for native arcs (experimental, not supported
                             by all firmwares)
         --gcode-comments    Make G-code verbose by adding comments (default: no)
--- a/lib/Slic3r/Config.pm
+++ b/lib/Slic3r/Config.pm
@@ -242,7 +242,7 @@ sub validate {
     die "Invalid value for --gcode-flavor\n"
         if !first { $_ eq $self->gcode_flavor } @{$Options->{gcode_flavor}{values}};
     
-    die "--use-firmware-retraction is only supported by Marlin firmware\n"
+    die "--use-firmware-retraction is only supported by Marlin and Machinekit firmware\n"
         if $self->use_firmware_retraction && $self->gcode_flavor ne 'reprap' && $self->gcode_flavor ne 'machinekit';
     
     die "--use-firmware-retraction is not compatible with --wipe\n"
--- a/lib/Slic3r/Config.pm
+++ b/lib/Slic3r/Config.pm
@@ -247,6 +247,13 @@ sub validate {
     
     die "--use-firmware-retraction is not compatible with --wipe\n"
         if $self->use_firmware_retraction && first {$_} @{$self->wipe};
+
+    # --use-velocity-extrusion
+    die "--use-velocity-extrusion is only supported by Machinekit firmware\n"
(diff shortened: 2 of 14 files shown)


strahlex:
@alexrj I would like to add some tests too. Is there some documentation on how to specify the test cases correctly?

nophead:
How does this produce a different result from sending E? You can work out
the cross section from the line length and E and the firmware keeps E in
sync with the axes so it just seems like a different notation for the same
information.

strahlex:
@nophead I have created a Python script to verify this method: https://github.com/strahlex/Machinekit-Slic3r-postprocess/blob/master/velocity-test.py
This script calculates the cross section from the E axis. Problem is that the E axis movement has a lot of rounding errors in it. When I print the cross sections they differ a few percent for every movement. So why introducing additional error if it is not necessary?

And one obvious advantage is that we get smaller GCode files...

alexrj:
Thank you for the contribution. I think this will not be merged very soon as it affects important parts of the code and there's no general consensus around this feature. M600 has another meaning in Marlin, and only Machinekit implements this velocity extrusion. Still, I'm not saying this will never be implemented.

Regarding error, it's true that decimal truncation in G-code introduces an error for each single command (< 0.00001mm of feedstock), but that error doesn't sum up, as Slic3r increments its E as a double-precision floating number internally, and then outputs each line from that one.
M600 is subject to decimal truncation as well, so it still has an error...

alexrj:
By the way, where's Machinekit M600 documented?

strahlex:
Interesting. Marlins M600 is not documented in the RepRap GCode
documentation: http://reprap.org/wiki/G-code We could use any other
MCode as well if that makes things easier.

This is my repository for standardizing the Machinekit FDM GCodes:
https://github.com/strahlex/Machinekit-FDM-Standard/blob/master/GCode.md#m600-set-line-cross-section

Regarding floating point error: It is true that the error does not sum
up. But, especially for short line segments the error is not unimportant
and can be as I already mentioned a few percent of the indented cross
section. When using velocity extrusion and the cross section MCode there
is still an error but it is always the same -> resulting in more
constant overall extrusion precision. The smaller layer heights we use
and the more arcs we use the more important this will be. (unless we use
G2/G3 of course).

When you take a look at my changes you may see that for example the
changes from ExtrusionEntity to GCodeWriter would not be necessary for
this to work. If it makes things easier I can create a separate PR for
these. Velocity extrusion is in fact only an additional output option
and should not influence the behavior of Slic3r at all.

Alessandro Ranellucci writes:

> By the way, where's Machinekit M600 documented?
>
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/alexrj/Slic3r/pull/2716#issuecomment-77867927

-- 
Alexander Rössler

strahlex:
I split up the changes to make them easier to merge:
https://github.com/alexrj/Slic3r/pull/2743
https://github.com/alexrj/Slic3r/pull/2742

strahlex:
Also using M700 instead of M600 to avoid aliasing with the Marlin firmware.

strahlex:
https://github.com/alexrj/Slic3r/pull/2744

Issue #2725, opened 2015-03-10
placement of models specified on command line does not honor bed dimensions


galexander1:
I made a simple stl with just a 1000mm rectangle in it.  That is much too big.  At 1.0.1, I would get this error:

1 parts won't fit in your print area!

But 1.1.7 and 1.2.6 and git newest all accept it without comment.  I don't know if the same problem exists in the GUI, though I suspect it does.

At 1.0.1, Geometry::arrange would get its $areax,$areay from Model::_arrange, which used to config->bed_size.  Now there is a $bb, passed down all the way from Model::duplicate and duplicate_objects.  But the calls to those two subs from Print::Simple::set_model doesn't provide a $bb.

I'm going to attempt to get the info from config->bed_shape in Print::Simple::set_model, and submit a pull request if successful... But I figure whoever changed this probably has a grand plan and I'd rather not get in the way of that..

Thanks! I love slic3r as a user, and now that I'm looking under the hood, I am even more impressed!!

galexander1:
Well, I made the change in pull request #2736.  It works for me, in that now it complains if I use commandline to add an object that is too big, and it centers the object in the center of the bed_shape region.  I'm not sure if this is really what the plan was when bed_shape was introduced, so I sure won't be offended if a different patch is prefered.  But for me this patch was necessary in order to be able to use the new Slic3r, otherwise it tried to place the object centered at 100,100 no matter what.

BTW - new Slic3r (version 1.2.x) is *MUCH* faster than version 1.0.1.  Thanks!!!!

alexrj:
Let's first start from identifying what problem(s) are you reporting. Good description of problems is the necessary premise before jumping to solutions.
As far as I understand you're reporting three distinct things:

1. Command-line slicing will not prevent you from supplying a 1000mm model even though your print bed is supposed to be smaller.
2. Command-line slicing will center objects objects at 100,100 by default instead of computing the center point dynamically from the configured bed shape.
3. Command-line slicing will not take the actual bed shape into account when auto arranging parts.

Regarding point 1, that happens by design. Command-line slicing does not enforce bed size for validating model input because we just do what user wants. the CLI API is used by host software (for example Repetier-Host) that already check print boundaries, and we don't want to require additional data from them. Still, a warning and/or an additional command line option to enforce bed size could be useful. But it shouldn't be based on the `arrange()` algorithm (which can lead to false negatives) anyway.

[...]

PR #2736, opened 2015-03-15
use bed_shape to determine bounding box and center of print area

galexander1:
in Slic3r::Print::Simple (commandline model positioning).  This fixes issue #2725.


alexrj:
Thank you for the contribution. Unfortunately I can't merge this because it would have two unwanted effects:

* the `--print-center` command line option would have no effect anymore
* Slic3r will enforce a not very accurate bed check preventing user from slicing legit plates

PR #2742, opened 2015-03-17
correct string in error message

strahlex:
This patch corrects an error message.


strahlex:
I am pretty sure the build did not fail because of the change in the string.

alexrj:
Thank you, applied.

PR #2744, opened 2015-03-17
Moving GCode output from ExtrusionEntity to GCodeWriter

strahlex:
For some reason extrusion GCode is output directly from ExtrusionEntity. This patch moves the output to GCodeWriter to make porting for different firmwares easier.

### DIFF
--- a/lib/Slic3r/GCode.pm
+++ b/lib/Slic3r/GCode.pm
@@ -309,7 +309,7 @@ sub _extrude_path {
     my $path_length = unscale $path->length;
     {
         my $extruder_offset = $self->config->get_at('extruder_offset', $self->writer->extruder->id);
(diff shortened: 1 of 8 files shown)


lordofhyphens:
@strahlex Could you rebase this onto current master and resolve the conflicts, although I wonder if the work has already been done by @alexrj separately.

strahlex:
There have been some structural changes in the projects it seems. Not sure what has been done since I proposed this PR.

lordofhyphens:
Since your change is straightforward, probably easiest to look at
gcodewriter or extrusionentity and see if it looks like similar changes got
made.

alexrj:
Thank you @strahlex for working on this, I confirm this change was applied. Not exactly as in your commit, but same result.

PR #2799, opened 2015-04-25
Support for standalone C++ library

jdfr:
This pull request contains some minor bug fixes to compile the core C++ sources as a standalone library, as well as some minor changes to make it more convenient as a library (mainly un-hardcoding values and behaviours). 

The last commit adds several overloaded versions of offset2() in ClipperUtils. I can merge all offset2() definitions into a single templated function, if it is better that way

### DIFF
--- a/xs/src/libslic3r/BridgeDetector.cpp
+++ b/xs/src/libslic3r/BridgeDetector.cpp
@@ -3,6 +3,10 @@
 #include "Geometry.hpp"
 #include <algorithm>
 
+#ifdef SLIC3R_DEBUG
+#include <stdio.h>
+#endif
+
 namespace Slic3r {
 
 class BridgeDirectionComparator {
--- a/xs/src/libslic3r/ClipperUtils.cpp
+++ b/xs/src/libslic3r/ClipperUtils.cpp
@@ -597,7 +597,7 @@ void union_pt_chained(const Slic3r::Polygons &subject, Slic3r::Polygons* retval,
     traverse_pt(pt.Childs, retval);
 }
 
-static void traverse_pt(ClipperLib::PolyNodes &nodes, Slic3r::Polygons* retval)
+void traverse_pt(ClipperLib::PolyNodes &nodes, Slic3r::Polygons* retval)
 {
     /* use a nearest neighbor search to order these children
        TODO: supply start_near to chained_path() too? */
--- a/xs/src/libslic3r/ClipperUtils.cpp
+++ b/xs/src/libslic3r/ClipperUtils.cpp
@@ -89,6 +89,24 @@ Slic3rMultiPoints_to_ClipperPaths(const T &input, ClipperLib::Paths* output)
     }
 }
 
+void Add_Slic3rExPolygon_to_ClipperPaths(const Slic3r::ExPolygon &input, ClipperLib::Paths* output) {
+    ClipperLib::Path p;
+    Slic3rMultiPoint_to_ClipperPath(input.contour, &p);
+    output->push_back(p);
+    for (Slic3r::Polygons::const_iterator it = input.holes.begin(); it != input.holes.end(); ++it) {
+        Slic3rMultiPoint_to_ClipperPath(*it, &p);
+        output->push_back(p);
+    }
+}
+
+void Slic3rExPolygons_to_ClipperPaths(const Slic3r::ExPolygons &input, ClipperLib::Paths* output) {
+    output->clear();
+    for (Slic3r::ExPolygons::const_iterator it = input.begin(); it != input.end(); ++it) {
+      Add_Slic3rExPolygon_to_ClipperPaths(*it, output);
+    }
+}
+
+
 void
 scaleClipperPolygons(ClipperLib::Paths &polygons, const double scale)
 {
@@ -229,6 +247,35 @@ offset_ex(const Slic3r::Polygons &polygons, const float delta,
 }
 
 void
+offset2(ClipperLib::Paths &input, ClipperLib::Paths* retval, const float delta1,
+    const float delta2, const double scale, const ClipperLib::JoinType joinType, const double miterLimit)
+{
+    // scale input
+    scaleClipperPolygons(input, scale);
+    
+    // prepare ClipperOffset object
+    ClipperLib::ClipperOffset co;
+    if (joinType == jtRound) {
+        co.ArcTolerance = miterLimit;
+    } else {
+        co.MiterLimit = miterLimit;
+    }
+    
+    // perform first offset
+    ClipperLib::Paths output1;
+    co.AddPaths(input, joinType, ClipperLib::etClosedPolygon);
+    co.Execute(output1, (delta1*scale));
+    
+    // perform second offset
+    co.Clear();
+    co.AddPaths(output1, joinType, ClipperLib::etClosedPolygon);
+    co.Execute(*retval, (delta2*scale));
+    
+    // unscale output
+    scaleClipperPolygons(*retval, 1/scale);
+}
+
+void
 offset2(const Slic3r::Polygons &polygons, ClipperLib::Paths* retval, const float delta1,
     const float delta2, const double scale, const ClipperLib::JoinType joinType, const double miterLimit)
 {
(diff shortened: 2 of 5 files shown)


alexrj:
Thank you, I appreciate that. Give me some days for reviewing the patches.

alexrj:
So, first off I'd like to thank you for the effort spent on libslic3r. I'll definitely merge the changes related to compilation and the safety_offset used in TriangleMesh.

I disagree with the changes to ClipperUtils.cpp. `eraseOutput` looks superfluous: callers can easily call `clear()` themselves if they want to overwrite an existing vector, instead of supplying the `true` argument (an explicit `clear()` is also more clear than a `true` argument). If they don't clear the vector, ClipperUtils.cpp will just append results to their vector, which is indeed useful when collecting the results of multiple operations into a single vector.

Also, I think that overloading functions just for accepting input `ExPolygons` objects is overkill and leads to a lot of code duplication. We just need a global `Polygons to_Polygons(const ExPolygons &expp)` function that callers can use to convert their input at call time...

lordofhyphens:
On 05/24/2015 05:28 PM, Alessandro Ranellucci wrote:
>
> So, first off I'd like to thank you for the effort spent on libslic3r. 
> I'll definitely merge the changes related to compilation and the 
> safety_offset used in TriangleMesh.
>
> I disagree with the changes to ClipperUtils.cpp. |eraseOutput| looks 
> superfluous: callers can easily call |clear()| themselves if they want 
> to overwrite an existing vector, instead of supplying the |true| 
> argument (an explicit |clear()| is also more clear than a |true| 
> argument). If they don't clear the vector, ClipperUtils.cpp will just 
> append results to their vector, which is indeed useful when collecting 
> the results of multiple operations into a single vector.
>
> Also, I think that overloading functions just for accepting input 
> |ExPolygons| objects is overkill and leads to a lot of code 
> duplication. We just need a global |Polygons to_Polygons(const 
> ExPolygons &expp)| function that callers can use to convert their 
> input at call time...
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/alexrj/Slic3r/pull/2799#issuecomment-105072036>.
>
Could we create a Polygons constructor that takes a ExPolygon instead of 
a global function? Maybe with something that takes an rvalue as well (so 
we could use std::move or std::forward in client code)?

-- 
--Joseph Lenox, BS, MS
I'm an engineer. I solve problems.

alexrj:
@lordofhyphens, it would be definitely nicer but unfortunately Polygons and ExPolygons are typedefs of std::vector<> and std::vector<> can't be subclassed/extended...

lordofhyphens:
Since the vector::copy algorithm works through the assignment operator, we
can still package things nicer by including an assignment operator for
Polygon -> ExPolygon. This should render conversion to as simple as
"ExPolygons dst(); std::copy(src.begin(), src.end(), dst.begin());".

It looks like the vector range constructor works similarly (I don't use
that particular constructor often), so "ExPolygons dst(src.begin(),
src.end())" could work as well if we provide a constructor for ExPolygon
that takes a Polygon as an argument or the proper cast operator.

Still avoids the need for a global function to do conversions.



jdfr:
> I disagree with the changes to ClipperUtils.cpp. eraseOutput looks superfluous: callers can easily call clear() themselves if they want to overwrite an existing vector, instead of supplying the true argument (an explicit clear() is also more clear than a true argument). If they don't clear the vector, ClipperUtils.cpp will just append results to their vector, which is indeed useful when collecting the results of multiple operations into a single vector.
> 
> Also, I think that overloading functions just for accepting input ExPolygons objects is overkill and leads to a lot of code duplication. 

I completely agree. Revisiting my last commit, I also find it overkill (to be fair, the previous code also contained quite a lot of duplications, but that commit just made things far worse). And anyway I am now using directly ClipperLib. So, I suppose you'll merge all but the last 3 commits. Thank you for taking the time to review it!

Please excuse my ignorance, I am new to githhub: should I remove this pull request and create a new one just with the relevant commits, or will you do it yourself?

> We just need a global Polygons to_Polygons(const ExPolygons &expp) function that callers can use to convert their input at call time...

But this would add an unnecessary conversion step from `ExPolygons` to `ClipperLib::Paths`. Not that it is so important, of course.

Just a comment, may be helpful: in a previous version of the pull request's last commit, I remixed the conversion functions `XXX_to_ClipperPaths` and `ClipperPaths_to_XXX` into a pair of overloaded functions, and tried to use them to create templated versions of `offset, offset2, _clipper_do` and `_clipper`. The goal was to create just one template for each one of these functions, to remove entirely the duplicated code. However, I abandoned this because the current overloaded versions of `offset` and `_clipper_do` have several different semantics depending on the type of some parameters, so the refactoring got too complex quite fast. It can be done, but the code becomes rather unintuitive.

jdfr:
There is also another question: I would also like to port the C++ sources to be compilable as a DLL in Visual Studio, but the changeset would be huge, mainly because to compile a Visual Studio DLL you have to add boilerplate to every single declaration you want to export. Afterwards, merging upstream commits would probably be quite time-consuming because almost every change in exported function declarations would require manual merging.

So, my question is: if I do the port, would you be interested in merging it upstream? I understand that you do not need it for Slic3r, but on the other hand it would likely introduce the library to a wider community of developers.

lordofhyphens:
If you need to add 'boilerplate', would it be possible instead to generate
a script to add the boilerplate (like Windows Powershell or similar)
instead and merge that into upstream? With a companion script to strip the
boilerplate, you should be able to more easily merge upstream changes and
still get the benefits.

jdfr:
> I disagree with the changes to ClipperUtils.cpp. eraseOutput looks superfluous: callers can easily call clear() themselves if they want to overwrite an existing vector, instead of supplying the true argument (an explicit clear() is also more clear than a true argument). If they don't clear the vector, ClipperUtils.cpp will just append results to their vector, which is indeed useful when collecting the results of multiple operations into a single vector.

mmmmm, I have to partially retract my previous agreement: while my original intention to add `eraseOutput` to all `ClipperUtils.cpp` functions was certainly overkill, it comes in handy for transformation functions, such as 'ClipperPaths_to_Slic3rMultiPoints(const ClipperLib::Paths &input, T* output)' and the like. In your current implementation, these functions will inconditionally do `output->clear()` or the equivalent. This actually seems to me to be the default behaviour for almost all of these functions in `ClipperUtils.cpp`, so now I think that I have misunderstood your words in some way.


> If you need to add 'boilerplate', would it be possible instead to generate
> a script to add the boilerplate (like Windows Powershell or similar)
> instead and merge that into upstream? With a companion script to strip the
> boilerplate, you should be able to more easily merge upstream changes and
> still get the benefits.

Definitely worth a try. Thank you for the suggestion!

I just noticed that I have not disclosed why I have done these patchs: I find perl a rather difficult language, so as a side project to reuse slic3r's excellent codebase, I am preparing python bindings for `Clipper.cpp`, `ExPolygon.cpp` and `TriangleMesh.cpp`, fudamentally. This way, I can write python scripts to slice 3d objects and modify the sliced contours in arbitrary ways.

I have also prepared a quick-and-dirty patch to enable slic3r to accept as input a file containing a serialized sequence of z values and `ClipperLib::Paths` instead of just 3d files. This way, after processing the contours with my scripts, I can use slic3r for infilling and gcode generation.

I can also do a pull request for that patch if you are interested, but I understand that the use case is a bit exotic for the average user of slic3r. Also, the patch as it currently stands is a bit of a hack. A more proper solution would be to read the layers from SVG files and enable fancy stuff like multiple extruders, I will do that if I find the time.

alexrj:
Yay for Python bindings! :)

Regarding importing paths, I think that libslic3r should never expose `ClipperLib::*` objects as the usage of Clipper is an implementation detail that might (hopefully not) change in the future. Only `Slic3r::*` primitives should be exposed by the library. Importing stuff should be done with other standard formats like SVG or DXF, or any simpler one out there...

alexrj:
@jdfr, I have no experience with Visual Studio. How does that boilerplate look like? Can you show an example of the changes you'd need?

jdfr:
> @jdfr, I have no experience with Visual Studio. How does that boilerplate look like? Can you show an example of the changes you'd need?

I don't have experience either, but most people recommend something like the following. In a header included everywhere, this has to be defined:

```c++
#ifdef ( defined(_WIN32) || defined(_WIN64) )
#    ifdef LIBRARY_EXPORTS
#        define LIBRARY_API __declspec(dllexport)
#    else
#        define LIBRARY_API __declspec(dllimport)
#    endif
#elseif
#    define LIBRARY_API
#endif
```

`LIBRARY_EXPORTS` is to be defined only when compiling the DLL. Then, on each *.hpp and *.cpp with functions or classes to be exported, you must add `LIBRARY_API` to the functions and classes to be exported, both in the *.hpp and the *.cpp files:

```c++
LIBRARY_API void chained_path(const Points &points, std::vector<Points::size_type> &retval, Point start_near);

class LIBRARY_API TriangleMesh {
```

To be fair, I am still pondering if this is worth the effort. Technically, I could write the bindings in just one python extension module, and statically link the c++ sources with them. But writing everything in one giant module has a number of disadvantages (unwieldy, really big cython source, and very slow recompilation while debugging). Another way to do it is just to statically link a copy of the c++ sources for each python extension module, but that's very ugly. 

> Regarding importing paths, I think that libslic3r should never expose ClipperLib::* objects as the usage of Clipper is an implementation detail that might (hopefully not) change in the future. Only Slic3r::* primitives should be exposed by the library.

I see your point, and it's sensible. Problem is, functions in ClipperUtils.cpp internally convert from/to `Polylines/Polygons/Expolygons` and `ClipperLib:Paths` every time you use them. Not a big penalty hit for the way you use them, but I would like to support workflows with tens of clipping/offseting operations and really huge slices. Working with naked `ClipperLib::Paths` has an obvious advantage in that case. I'll think about this.

> Importing stuff should be done with other standard formats like SVG or DXF, or any simpler one out there...

Yes, it's the most reasonable way to do it.

lordofhyphens:
What about taking advantage of move semantics for the conversion? It should
minimize the overhead and let us keep the interface clean.

If the boilerplate is all behind ifdefs (that is, the code will still
compile everywhere else, I would be more okay with it being upstream.

jdfr:
> What about taking advantage of move semantics for the conversion? It should
> minimize the overhead and let us keep the interface clean.

I don't use regularly c++, so I'm not sure of understanding how this would help. I would do something like this, for example for `offset()`:

```c++
//instantiate/overload these two functions for each slic3r type 
template <class T> to_ClipperPaths(T &input, ClipperLib::Paths &output);
template <class T> from_ClipperPaths(ClipperLib::Paths &input, T &output);

template <class Input, class Output>  offset (Input &input, Output &output, const float delta, double scale, ClipperLib::JoinType joinType, double miterLimit) {
ClipperLib::Paths inp, outp;
  to_ClipperPaths(input, inp, delta, scale, joinType, miterLimit);
  offset(inp, outp);
  from_ClipperPaths(outp, output);
}

template <ClipperLib::Paths,ClipperLib::Paths> offset(ClipperLib::Paths &input, ClipperLib::Paths &output, const float delta, double scale, ClipperLib::JoinType joinType, double miterLimit) {
    // scale input
    scaleClipperPolygons(input, scale);
    
    // perform offset
    ClipperLib::ClipperOffset co;
    if (joinType == jtRound) {
        co.ArcTolerance = miterLimit;
    } else {
        co.MiterLimit = miterLimit;
    }
    co.AddPaths(input, joinType, ClipperLib::etClosedPolygon);
    co.Execute(output, (delta*scale));
    
    // unscale output
    scaleClipperPolygons(output, 1/scale);
  
}
```

Unless I missed something critical, the only problem I see with this approach is that functions implemented in this way must have the same semantics for all types, so you cannot use `ClipperLib::etClosedPolygon` for `ExPolygons` and `ClipperLib::etOpenButt` for `Polylines` automatically (well, you can, but you need to do some contrived things with templates, AFAIK), so this should be made an explicit argument with this approach. `_clipper_do()` and `_clipper()` would be trickier.

lordofhyphens:
A move constructor is a shallow copy that leave the previous object
unintialized.

alexrj:
Ouch, that boilerplate is ugly.  I've never seen it in any open source library out there. Is it common practice to put it in the upstream repo?

Regarding the conversions between Slic3r primitives and Clipper primitives, I totally agree with your concern about performance penalty. Clipper should be patched in order to work with custom types (it would be a matter of macros or traits, just like Boost.Polygon does). Still, I think that if you want to work with Clipper types directly in your application you don't need to use ClipperUtils.cpp at all as it's just a wrapper... can't you just call Clipper?

BTW, any ideas about how to start documenting libslic3r API?

jdfr:
> Ouch, that boilerplate is ugly.

Yes. I have yet to see beautiful boilerplate ;)

> I've never seen it in any open source library out there.

It's actually quite used in big projects, in one form or another, OpenCV springs to my mind. Hey, even some Boost sublibraries use it, there are `__declspec` definitions in your own source tree in the `boost` subdir!

cmake eases the pain by automatically generating a header file with all the ifdefs, but you still have to explicitly add the boilerplate to the functions you want to export. 

> Is it common practice to put it in the upstream repo?

I don't know. To be frank, I develop in linux and supporting windows would be quite nice, but Visual Studio requirements are also annoying for me. Like I just said in my previous comment, I may just end up statically linking everything, to avoid this problem.

> Still, I think that if you want to work with Clipper types directly in your application you don't need to use ClipperUtils.cpp at all as it's just a wrapper... can't you just call Clipper?

That's actually what I am doing right now. I expose interconvertible python wrappers for both `std::vector<ExPolygons>` and lists of `ClipperLib::Paths`. The former is the preferred one, and the later is mainly used for clipping/offseting operations.

In fact, my last few comments about `ClipperUtils.cpp` were just answering to your line:

> libslic3r should never expose ClipperLib::* objects"

and lordofhyphens' followup:

> What about taking advantage of move semantics for the conversion?

PR #2873, opened 2015-06-01
Fixes #2858

### LINKED ISSUE #2858: Missing Filament color config causes crash on startup
Slic3r version: aa8b6af (bug introduced in this commit -- previous commits work fine).
wxWidgets version wxWidgets 3.0.2, Wx version 0.9927

Platform: Arch Linux

### Steps to reproduce:

**Method 1**:

1. Use a .Slic3r directory, and slic3r.ini from a previous installation of Slic3r before this commit
1. Make sure 'expert' mode is enabled in slic3r.ini
1. install version aa8b6af or 310212e
1. Run `slic3r.pl`

**Method 2**:
(Backup your existing .Slic3r data directory somewhere else)

1.  Completely remove your .Slic3r directory
1. install version aa8b6af or 310212e
1. Run `slic3r.pl`
    - This should run successfully, and ask setup questions about your machine. Defaults are fine, or click cancel
1. Enable Expert Mode, and exit.
1. Run `slic3r.pl` again

**Results**:

Program immediately stops/crashes (no GUI ever shows), and gives the output:
```Can't call method "config" on an undefined value at /usr/share/perl5/vendor_perl/Slic3r/GUI/Tab.pm line 407.```

From what I can gather, at some point during filament config loading, the second parameter given to `get_preset_config`, that is, `preset` is null or empty, and the program crashes. 

The resulting `slic3r.ini` from the second method is pasted below, which should trigger the behavior:

```
# generated by Slic3r 1.2.8-dev on Tue May 26 13:19:04 2015
autocenter = 1
background_processing = 1
main_frame_maximized = 
main_frame_pos = 6,26
main_frame_size = 1341,509
mode = expert
version = 1.2.8-dev
version_check = 1

[presets]
filament = 
print = 
printer =
```

darenschwenke:
The consumers of filament_presets expected an array index.

### DIFF
--- a/lib/Slic3r/GUI/Plater.pm
+++ b/lib/Slic3r/GUI/Plater.pm
@@ -471,7 +471,7 @@ sub update_presets {
 sub filament_presets {
     my $self = shift;
     
-    return map $_->GetSelection, @{ $self->{preset_choosers}{filament} };
+    return map scalar($_->GetSelection), @{ $self->{preset_choosers}{filament} };
 }
 
 sub add {


alexrj:
`GetSelection()` does return an array index... putting a `scalar()` around it will always return 1 since `scalar()` counts the number of items in a list, and that will always be 1...

darenschwenke:
If scalar is passed a list, yes it counts the number of items.  
That is not what is happening here.
scalar is just telling GetSelection we want a scalar back.  GetSelection, when called in a scalar context, returns the index of the selected item.... which is what you need here.  
It works.

alexrj:
Okay, it wasn't clear to me (due to poor wxPerl documentation) that `GetSelection()` is context-aware and returns a list when called in a list context.

Reference:
https://metacpan.org/source/MDOOTSON/Wx-0.9927/XS/ComboBox.xs#L184

PR #2990, opened 2015-07-08
To let Slic3r compile in Windows

a4jp-com:
setenv isn't compatable in Windows but _putenv_s is.

How can I update all my files to your recent ones. I synced but the files I have are 8 months old for some reason.

### DIFF
--- a/xs/src/boost/test/utils/runtime/config.hpp
+++ b/xs/src/boost/test/utils/runtime/config.hpp
@@ -79,7 +79,7 @@ putenv_impl( cstring name, cstring value )
 {
     using namespace std;
     // !! this may actually fail. What should we do?
-    setenv( name.begin(), value.begin(), 1 );
+    _putenv_s( name.begin(), value.begin());
 }
 #else
 inline void


Mikk36:
Did you not miss the xs\src\libslic3r\Config.cpp file?

kefir-:
You have to update your fork. 

https://help.github.com/articles/syncing-a-fork/

alexrj:
No please, this would fix Windows and break all other other operating systems.

a4jp-com:
Would an if clause of some type work? If (Windows) {blah} else {blah blah};

`my $windows=($^O=~/Win/)?1:0;# Are we running on windows?`

`my $osname = $^O;


if( $osname eq 'MSWin32' ){{
  eval { require Win32; } or last;
  $osname = Win32::GetOSName();

  # work around for historical reasons
  $osname = 'WinXP' if $osname =~ /^WinXP/;
}}
Derived from sysinfo.t, which I wrote the original version.`

For more detailed information:

`my ( $osvername, $major, $minor, $id ) = Win32::GetOSVersion();`

lordofhyphens:
If it's all in the XS C/C++ files, probably better to do something akin to

#ifdef windows
#define PUTENV _putenv_s
#else
#define PUTENV putenv
#endif

filling in the correct variable that is defined by the environment. Then
you'd replace all putenv() calls with PUTENV so that the macro would work.

a4jp-com:
Config.cpp and config.hpp anre the 2 files that need fixing so Windows PCs can compile the code again.

I currently use this code:

Config.cpp
Line 115: `_putenv_s(envname.c_str(), this->serialize(*it).c_str());`

config.hpp
Line 82: `_putenv_s( name.begin(), value.begin());`

But an if statement is needed to make all the operating systems work properly. If you know a good way to fix this, could you help a bit? I would love to be able to compile the source code again. I'm on a Windows machine so I need at least the code above to compile the files.

PR #3033, opened 2015-08-04
add support for possible use of integer coordinates beyond 2^31

jdfr:
The default voronoi implementation in Boost uses int32 for the coordinates. When compiling with 64-bit support, `long` has 64 bits. When `construct_voronoi()` is called, your 64-bit coordinates are truncated to 32 bits. For details, see the summary at the bottom of  [boost voronoi's basic tutorial](http://www.boost.org/doc/libs/1_54_0/libs/polygon/doc/voronoi_basic_tutorial.htm), or the definition of `voronoi_ctype_traits` in `boost/polygon/detail/voronoi_ctypes.hpp`.

Now, this is not a problem for Slic3r because the `SCALING_FACTOR` is 1e-6, so your coordinates allow to specify objects up to ~ 4 meters wide while they still fit in 32 bits, and I bet nobody has ever attempted to an object bigger than that with Slic3r.

However, if you ever change the `SCALING_FACTOR` or Slic3r is used for really big objects (think of the Chinese 3D-printed houses XD), the current code may run into problems. An obvious fix would be to change the definition of `voronoi_ctype_traits`, (I believe) that would imply a big performance penalty, as it requires non-standard integer types (see [boost's voronoi advanced tutorial](http://www.boost.org/doc/libs/1_54_0/libs/polygon/doc/voronoi_advanced_tutorial.htm)). Instead, I have just implemented an optional transformation for the geometric objects used in MedialAxis::build. In most cases, a simple translation will be enough to fit the geometry in 32 bits. If it is not enough, scaling is applied.

Ok, so this is not really needed right now, but it enables support if you need it in the future. Just thinking you may find it useful.


alexrj:
I appreciate that, thank you. Give me some time for studying it before merging. I'm a but busy these days.

Slic3r is being used on machines larger than 4 meters, but in those environments units are not assumed to be mm anymore :)

a4jp-com:
Sounds like we will be able to print houses with this software soon, or even a car ^^

lordofhyphens:
@alexrj did you still want to review this for inclusion or should it be closed?

bubnikv:
My preferred solution to this issue, which may or may not happen in the future if someone tries to build houses, is to add a static assert to test the SCALING_FACTOR.

jdfr:
@bubnikv AFAICT, static asserts are an addition from c++11. Would it be possible to compile the c++ modules as c++11 and then use them from Perl?

bubnikv:
@jdfr You are right, for example the Strawberry Perl is packed with a GCC, which does not support the static assert construct. But boost provides a similar macro for older C++ standards.

PR #3112, opened 2015-11-04
Fix temperature setting on multiple extruders, fixes #3101

### LINKED ISSUE #3101: Changing the filament temperature for "other layers" causes an unintended tool change.
On a dual carriage, dual extruder printer, I've set the filament temperatures to "First layer: 205" and "Other layers: 220".

I've set "Tool change G-code" to G199, which is my own gcode for parking the current tool (and is carriage) out of the way of the other carriage.

The gcode generated at the first layer change is as follows:

G1 X104.146 Y129.919 E30.22393
M104 S220 T0 ; set temperature
G1 E29.72393 F900.00000
G92 E0
G1 Z0.500 F7500.000
G1 E-9.50000 F900.00000
G92 E0
G199
T0
G92 E0
G1 X96.279 Y130.679 F7500.000

At the beginning of this code, the current tool is T1, and T0 is parked offsides on the left hand side of the printer.

When the M104 line is reached, smoothieware makes a tool change (as specified by the NIST gcode standard).  That tool change was made without parking T1.

When the G199 command is reached, the printer parks T0 (which is already parked).

The T0 command has become a NOP, because the printer has already switched tools.

When the final G1 command executes, the T0 carriage crashes into the T1 carriage, because T1 was never parked.

If the M104 command is removed, this code works as expected - when the G199 command is reached, it parks T1 on the right hand side of the printer in preparation for the T0 command.

mdebski:
By [0], repraps does not support toolhead "T%d" parameter M104.
This is indeed the case for Smoothie. It interpretes "M104 Snnn Tn" as
M104 Snnn
Tn
so it may cause unwanted toolhead change.
[0] http://reprap.org/wiki/G-code#M104:_Set_Extruder_Temperature

### DIFF
--- a/t/custom_gcode.t
+++ b/t/custom_gcode.t
@@ -77,16 +77,16 @@ use Slic3r::Test;
     {
         my $print = Slic3r::Test::init_print('20mm_cube', config => $config);
         my $gcode = Slic3r::Test::gcode($print);
-        ok $gcode =~ /M104 S205 T1/, 'temperature set correctly for non-zero yet single extruder';
-        ok $gcode !~ /M104 S\d+ T0/, 'unused extruder correctly ignored';
+        ok $gcode =~ /T1\nM104 S205/, 'temperature set correctly for non-zero yet single extruder';
+        ok $gcode !~ /T0\nM104 S\d+/m, 'unused extruder correctly ignored' . substr($gcode, 0, 400);
     }
     
     $config->set('infill_extruder', 1);
     {
         my $print = Slic3r::Test::init_print('20mm_cube', config => $config);
         my $gcode = Slic3r::Test::gcode($print);
-        ok $gcode =~ /M104 S200 T0/, 'temperature set correctly for first extruder';
-        ok $gcode =~ /M104 S205 T1/, 'temperature set correctly for second extruder';
+        ok $gcode =~ /T0\nM104 S200/m, 'temperature set correctly for first extruder';
+        ok $gcode =~ /T1\nM104 S205/m, 'temperature set correctly for second extruder';
     }
     
     $config->set('start_gcode', qq!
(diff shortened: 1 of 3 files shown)


mdebski:
I'm not sure how to solve this correctly. This approach works, but is not very nice and causes problems with some tests.

mdebski:
cura seems to generate "M104 T0 S210.000000".
Maybe this should be added as separate flavor?

Looking forward to your opinions, I'm willing to iterate on this.

alexrj:
What syntax does Smoothie want?

By the way, parsing `M104 Sx Tx` as `M104 Sx` + `Tx` doesn't seem to make sense for any purpose. I think Smoothie should be changed to accept the `T` parameter for `M104`. Otherwise we'd need to add a Smoothie-specific G-code flavor which would be a little penalization for it in my opinion.

mdebski:
reprap wiki does not specify T parameter as standard, only notices that some firmwares support it.
I guess the only correct on Smoothie is 
Tn
M104 Snnn
I'll try slicing tomorrow with the propertiary software for my printer and see what it generates.
I guess parsing as M104 + T is just an unwanted effect of how the parser works.

mdebski:
I asked in the Smoothieware project - please see the discussion, especially [0] -> "Tn anywhere on the line will select that tool for that line and any subsequent line"

This allows for another idea for set_temperature (which may be what cura does, I'm not sure) - emit M10[49] Snnn Tn as currently, and then if tool != extruder->id emit additional Tn with n=extruder->id. This should be perfectly backwards-compatible and maybe even won't break many tests :P. On the other hand, it may generate superfluous "Tn"s - which should be a small issue as temperature changes do not happen often. What do you think ?

[0] https://github.com/Smoothieware/Smoothieware/issues/760#issuecomment-154208693

alke:
Why always bed_temperature taken from filament 1. When i print with extruder 2 its only set temperature from filament 2 not bed_temperature, it is taken from filament 1. is it the way it shuld be?

Regards Kent

lordofhyphens:
@alke I think you can use [bed_temperature_1] (use whatever the temp is in the dump of config settings at the end of the gcode) in your start-gcode. This would require that you have two profiles (one for each preferred bed temp). This is for initial heatup anyway.

Bed temp also doesn't tend to change a lot and it only most important for layers contacting the bed.

PR #3232, opened 2016-02-04
emit M200 D0 on disabled volumetric extrusion

sorki:
This fixes an issues that was haunting me for hours trying to fix my seemingly random under/over-extrusion issues which were caused by volumetric extrusion enabled in firmware (from prior attempts to use it) but not in Slic3r.

After I've disabled volumetric in firmware everything went back to normal.

This is quite suboptimal as there are two places where you have to disable volumetric to actually disable it and may cause headaches for people unaware of it.

### DIFF
--- a/xs/src/libslic3r/GCodeWriter.cpp
+++ b/xs/src/libslic3r/GCodeWriter.cpp
@@ -61,6 +61,10 @@ GCodeWriter::preamble()
             gcode << "M82 ; use absolute distances for extrusion\n";
         }
         gcode << this->reset_e(true);
+
+        if (!this->config.use_volumetric_e) {
+            gcode << "M200 D0 ; reset filament diameter (due to disabled volumetric extrusion)\n";
+        }
     }
     
     return gcode.str();


lordofhyphens:
Is it prepended to the start gcode or appended?

sorki:
Prepended

alexrj:
I'd like to hear comments from @whosawhatsis, @wolfmanjm and @Lenbok since the current behavior was decided after the following discussion:

https://github.com/alexrj/Slic3r/commit/47e4e8bb66740af791e56b36f0ecb67143a94b3b#comments

Lenbok:
The previous comments still stand, I think it is the domain of the firmware setup (or auto0.g etc) to do the volumetric enabling/diameter setting etc. 

Another reason  why M200 D0 should not be added, as that would break compatibility for people who achieve volumetric extrusion by setting their filament diameter to 1.128379 in the slicer (e.g. see http://reprap.org/wiki/Triffid_Hunter's_Calibration_Guide#Optional:_Switch_to_volumetric_E_units) by turning off the volumetric setting in their firmware.

wolfmanjm:
I agree, M200 D0 should not be emitted under any circumstances, This is up to the user to turn it on or off in the firmware. The agreement reached after lengthy discussions seemed to make most people happy :)

whosawhatsis:
As a compromise, how about adding a variable that would contain the filament diameter if volumetric extrusion is enabled and 0 if it is not, so you could just put something like the following in your start code:

M200 D[volumetric_diameter]

This way, the line could be included, and D0 would be sent if the slicer is not configured for volumetric extrusion.

wolfmanjm:
Doesn't that completely go against the whole point for volumetric extrusion in the firmware? being able to set the filament diameter regardless of the slicer setting? So I can slice once and print with different filament without having to reslice?

whosawhatsis:
@wolfmanjm Technically, this should be coming from the filament (either a measurement device or something like a UFID EEPROM), but most printers don't have either. That leaves storing a value on the machine or storing it in the slicer, and the latter is still supported by more people's setups than the former. Currently, I'm including an M200 line in all of my start codes. If I want to use a different filament diameter (I've even tested this sharing gcode between a 1.75 and 3mm machine), I just have to change the line in the start code. It's not how it's intended to be used ultimately, but it works with current systems, and I think this is probably how most people are using it now. The beauty of this solution is that it doesn't automatically generate an M200, so it will do what the OP is asking for if you include it in your start code, and won't do anything if you don't.

alexrj:
Thank you for (re)discussing this, it's very precious for me.
We could also think about something more interactive. When user checks the "Use Volumetric E" option we might inspect the custom start G-code and warn him with: *"If volumetric extrusion is not enabled in your firmware by default, you might want to add a `M200 D[filament_diameter]` command in your custom start G-code. Shall I add it for you?"* or, in case they're disabling volumetric E but they have a `M200`: *"You might want to remove the `M200` command from your custom start G-code. Shall I remove it for you or turn it into an explicit `M200 D0` which disables volumetric extrusion on the firmware as well?"* with three options: "No", "Remove it", "Use M200D0".
What do you think?

whosawhatsis:
@alexrj You're thinking of popping up a dialog box when they have possibly incompatible settings? Do you currently do this for anything else that might be OK as it is (where you don't force a change)? Seems like inconsistent behavior.

It also raises the question of what will happen if the user responds that they want to leave it the way it is. Do you store that information somewhere so that it won't ask again every time? If not, it's quite annoying, if so, a novice user just has to misunderstand the message once, and is likely to get confused and not know how to fix it if it doesn't work.

A warning icon (that is always shown, but can be ignored) as part of the main interface that appears next to the volumetric mode checkbox if there is not an M200 in the start code would be better. Clicking that icon could show the dialog box you describe, with the option to do nothing (though the icon would remain, and the dialog should mention that depending on your machine's configuration, it may be correct). This warning should be coded such that it is not displayed with a naked M200 (with no D parameter) in the start code (which is how machines with a measurement device or UFID EEPROM are intended to work). You should also handle multiple M200s correctly. For instance, it may be desirable to create a start code that sets M200 D0 to load and/or prime the nozzle in a dump position, then sets M200 D[filament_diameter] (or just M200) before printing.

I still think the [volumetric_diameter] (or whatever it's called) idea is a good one, though.

alexrj:
@whosawhatsis, a dialog is shown with Spiral Vase now and I think it's a nice aid for keeping the configuration consistent. Having a permanent warning is definitely good, and we need that pattern for more things as well (#3271).

sorki:
+1 for warning dialog.

lordofhyphens:
Looks like we would need a config option to store the flag to not repeatedly annoy the user. I would set and clear the flag during a validation routine for volumetric e checkbox. It would otherwise persist with profiles.

@sorki there is a list of files in the wiki you would need to touch to add a config option, and examples in the Spiral vase gui perl code.

PR #3250, opened 2016-02-23
Manually cast result of >> to bool

hroncok:
This is needed on newer versions of gcc (6.x.x), otherwise we get an error:

    cannot convert 'std::basic_istream<char>::__istream_type {aka std::basic_istream<char>}' to 'bool' in return

### DIFF
--- a/xs/src/libslic3r/Config.hpp
+++ b/xs/src/libslic3r/Config.hpp
@@ -87,7 +87,7 @@ class ConfigOptionFloat : public ConfigOptionSingle<double>
     
     bool deserialize(std::string str) {
         std::istringstream iss(str);
-        return iss >> this->value;
+        return static_cast<bool>(iss >> this->value);
     };
 };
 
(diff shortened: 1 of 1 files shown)


ledvinap:
You probably want `http://en.cppreference.com/w/cpp/io/basic_ios/good`

hroncok:
as in:

    iss >> this->value;
    return iss.good();

?

ledvinap:
Sorry, got lost in C++ documentation. Ignore the comment, your fix looks good.

alexrj:
Actually, the return value for `deserialize()` is supposed to indicate whether the deserialization was successful or not (and the way I implemented it was wrong). I fixed this in 6e5938c8330b5bdb6b85c3ca8dc188605ee56b98 with `return !iss.fail()`.
(@ledvinap was almost right, but `good()` also fails when EOF is reached in the input stream)
Thank you for raising the issue.

PR #3275, opened 2016-03-18
Work-in-progress addition to add arbitrary gcode after objects when sequential printing.

### LINKED ISSUE #3264: between objects (custom) G-code
When printing objects sequentially slic3r sets the first layer temperature for every objects first layer. But it does not wait for the temperature to be reached. It would be nice if slic3r would move the head to somewhere between objects but keep it at least 5mm above the bed, then set and wait for the first layer temperature and last print a (mini) skirt around the next object or extrude some extra material to refill the melt chamber after the pause.

On a more generic front the Print Settings->Custom G-code page should have a box for custom G-code between objects, which would get triggered somewhere in the empty space between objects and before lowering the head.

lordofhyphens:
Fixes #3264.

It's incomplete because I don't know how to properly add items to the Config. I think I have most of it, but it's dying at line 344 lib/Slic3r/GUI/OptionsGroup.pm because it can't find "after_object_gcode" which is the new one. 

XS builds just fine though :)

The test needs work as well, but I'd rather figure out what I'm doing wrong so I can also work out a test.

### DIFF
--- a/lib/Slic3r/GUI/Tab.pm
+++ b/lib/Slic3r/GUI/Tab.pm
@@ -993,7 +993,7 @@ sub build {
         octoprint_host octoprint_apikey
         use_firmware_retraction pressure_advance vibration_limit
         use_volumetric_e
-        start_gcode end_gcode before_layer_gcode layer_gcode toolchange_gcode
+        start_gcode end_gcode before_layer_gcode layer_gcode toolchange_gcode after_object_gcode
         nozzle_diameter extruder_offset
         retract_length retract_lift retract_speed retract_restart_extra retract_before_travel retract_layer_change wipe
         retract_length_toolchange retract_restart_extra_toolchange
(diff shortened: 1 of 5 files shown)


alexrj:
You also need to add the new option to the `GCodeConfig` class definition, as well as its `optptr()` method.

After you have fixed the test you also have to update the test count at the beginning of the test file.

lordofhyphens:
Thanks! I will get on that.

lordofhyphens:
The basic functionality works now, need to get the tests going. I'm new to most things Perl, is there a way I can just run the tests in the build system? The VM I'm building in is a bit slow to run a full compile and I've got my code running against my ad hoc tests.

There's one bug in the current code; the code is currently only placed after objects added to the plater if and only if they were both loaded separately; using the "more" button doesn't hit the code path I'm using.

Just something more to investigate.

lordofhyphens:
Can the test architecture test for the presence of a comment?

lordofhyphens:
I'm actually pretty sure it's not written for that.. that or I am just derping. Functionality of it works though. :)

alexrj:
Implemented in bc96a1a

Issue #3282, opened 2016-03-20
Pressure advance overrides print speed


Sebastianv650:
First, a big thanks for all the hard work to alexrj and all the other developers!
I was playing with the pressure advance setting in the last days, and once calibrated properly it's working realy good. I'm using 0.1 for a gregs wade extruder now.

I'm using the latest Master code with this config:
[config.txt](https://github.com/alexrj/Slic3r/files/181380/config.txt)
Test cube:
[20x20x10.stl.txt](https://github.com/alexrj/Slic3r/files/181383/20x20x10.stl.txt)

But I noticed a bug regarding the print speed, it looks like the printer is sometime printing slower than expected when pressure advance is enabled. I had a short look into the gcode, you can see the error here:

G1 Z0.540 F9000.000 ; move to next layer (2)
G1 E12.52740 F2100.000 ; pressure discharge
G92 E12.53477 ; restore E
G10 ; retract
G92 E0 ; reset extrusion distance
G1 X158.323 Y146.823 F9000.000 ; move to first perimeter point
G11 ; unretract
G92 E0 ; reset extrusion distance
**G1 F2700**
G1 E0.03596 **F2100.000** ; pressure advance
G92 E0.00000 ; restore E
G1 X139.677 Y146.823 E0.24848 ; perimeter
G1 X139.677 Y128.177 E0.49695 ; perimeter
G1 X158.323 Y128.177 E0.74543 ; perimeter
G1 X158.323 Y146.763 E0.99311 ; perimeter
G1 X158.760 Y147.260 F9000.000 ; move to first perimeter point
**G1 F1350**
G1 E0.96614 F2100.000 ; pressure advance
G92 E0.99311 ; restore E
G1 X139.240 Y147.260 E1.25324 ; perimeter
G1 X139.240 Y127.740 E1.51336 ; perimeter
G1 X158.760 Y127.740 E1.77349 ; perimeter
G1 X158.760 Y147.200 E2.03282 ; perimeter
G1 X158.374 Y147.156 F9000.000 ; move inwards before travel

---

I think this way the feed of the pressure control overrides the real intended feed rate?
PS: Please excuse some maybe wrong wording, I give my very best but english is not my first language ;)

markwal:
When you build from master, it's helpful if you also include a commit number (first 8 characters) because "master" will be meaningless as time goes on and looking up the commit later based on your report time is a PITA.

[...]

Issue #3283, opened 2016-03-20
Setting for Retraction speed greyed out with FW retraction and Pressure advance enabled


Sebastianv650:
Another one regarding Pressure control:
Slic3r version: Commit Nr. 5ff7511

When FW retraction is used, the field for retraction speed is greyed out. If Pressure control is enabled (not 0), it uses the retraction speed for it's pressure regulation moves. In this case, the field should be enabled I think.

whosawhatsis:
The field probably shouldn't be greyed-out, because the user might want to set retraction speed in the start code like this:

M207 F[retract_speed]

You might say it doesn't make much sense to use firmware retraction in that case, but that's the way I do a lot of things until the printer has a way to store those values. It still gives me the ability to alter those variables without reslicing, and to use the firmware feature of doing the un-retract at a different speed (which works really nicely, and might be worth considering for Slic3r, too).

alexrj:
@Sebastianv650, good catch.

@whosawhatsis, I see your point but both retract_speed and start_gcode are defined within the same Printer Settings preset, so it's pointless to use a placeholder. If you need to supply a value to `M207` you can write it directly in the start_gcode. Disabling fields is a key part of making the interface more friendly for the average user, since it provides immediate feedback about what effect other options have. Since any option might be used as a placeholder, we might want to leave everything always enabled...

whosawhatsis:
@alexrj A better way to do that is with the warning icons we discussed in another thread. There is so much variation in machines and configurations that disabling fields is risky (as the OP's issue demonstrates), but a warning icon attached to a dialog box will avoid this type of conflict while doing a better job of providing the feedback you mentioned.

alexrj:
@whosawhatsis, my point is that options should be disabled when Slic3r does not use them, which is not an ambiguous criteria. The OP case is just a bug (when Pressure Advance is enabled retract_speed *is* used). In other cases there's nothing to warn about :)

whosawhatsis:
Even if you do disable the input, I would still put one of these icons next to any disabled input. Maybe not a warning symbol, but maybe a question mark or info symbol. Clicking on the icon should pop up an explanation of why the box has been disabled, and which setting(s) need to be changed to re-enable it.

alexrj:
@Sebastianv650, I fixed this. Thank you for reporting.

PR #3287, opened 2016-03-22
Pressure advance overrides print speeds patch

Sebastianv650:
This solves #3282. It resets the print speed after a advance/discharge and it checks if the speed is unchanged, but an advance has to be made because a discharge was done before.

Please note, this is my first own coding acitivity at github and I'm not familiar with perl, only a little bit of C. If some informations are missing for the request, please let me know!

### DIFF
--- a/lib/Slic3r/GCode/PressureRegulator.pm
+++ b/lib/Slic3r/GCode/PressureRegulator.pm
@@ -36,7 +36,7 @@ sub process {
         } elsif ($info->{extruding} && $info->{dist_XY} > 0) {
             # This is a print move.
             my $F = $args->{F} // $reader->F;
-            if ($F != $self->_last_print_F || ($F == $self->_last_print_F && $self->_advance = 0)) {
+            if ($F != $self->_last_print_F || ($F == $self->_last_print_F && $self->_advance == 0)) {
                 # We are setting a (potentially) new speed or a discharge event happend since the last speed change, so we calculate the new advance amount.
             
                 # First calculate relative flow rate (mm of filament over mm of travel)
--- a/lib/Slic3r/GCode/PressureRegulator.pm
+++ b/lib/Slic3r/GCode/PressureRegulator.pm
@@ -62,7 +62,7 @@ sub process {
             }
         } elsif (($info->{retracting} || $cmd eq 'G10') && $self->_advance != 0) {
             # We need to bring pressure to zero when retracting.
-            $new_gcode .= $self->_discharge($args->{F});
+            $new_gcode .= $self->_discharge($args->{F}, $args->{F} // $reader->F);
         }
         
         $new_gcode .= "$info->{raw}\n";
@@ -76,14 +76,14 @@ sub process {
 }
 
 sub _discharge {
-    my ($self, $F) = @_;
+    my ($self, $F, $oldSpeed) = @_;
     
     my $new_E = ($self->config->use_relative_e_distances ? 0 : $self->reader->E) - $self->_advance;
     my $gcode = sprintf "G1 %s%.5f F%.3f ; pressure discharge\n",
         $self->_extrusion_axis, $new_E, $F // $self->_unretract_speed;
     $gcode .= sprintf "G92 %s%.5f ; restore E\n", $self->_extrusion_axis, $self->reader->E
         if !$self->config->use_relative_e_distances;
-	$gcode .= sprintf "G1 F%.3f ; restore F\n", $args->{F} // $reader->F;
+	$gcode .= sprintf "G1 F%.3f ; restore F\n", $oldSpeed;
     $self->_advance(0);
     
     return $gcode;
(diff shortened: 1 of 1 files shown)


VanessaE:
This seems to eliminate the unexpectedly-fast print moves in the model I was testing with.  Speeds seem to be similar to what I had configured prior to enabling pressure advance.  One note though: the following warning appears on the controlling terminal when exporting g-code:

`Use of uninitialized value $oldSpeed in sprintf at /home/vanessa/RepRap/Slic3r-git-pressure/lib/Slic3r/GCode/PressureRegulator.pm line 86.`

Sebastianv650:
@VanessaE, you might also have a look at https://github.com/MarlinFirmware/Marlin/pull/3676, this Marlin "extension" handles the nozzle pressure much better than it can be done inside a slicer. I wrote that code when I recognised that my prints are better with Slic3rs pressure control, but not perfect.

VanessaE:
I'll pass but thanks for the suggestion.  Ultimately pressure advance wasn't what I needed, but at least your code seems to work. :smile:

VanessaE:
This code seems to generally work, but with one glitch: as I approach the right amount of advance for my setup (common NEMA17 motor with MK8 hob, no extra gearing, B'struder drive block, feeding into 55cm of bowden tube), the extruder motor tends to skip a step at restart if there's a big enough change in flow rate from one retract to the next restart, or at least the behavior of the motor is very similar.

I have to wonder if it's appropriate to ramp the extruder speed up so quickly at the start of a layer.

My retract speed is usually at 200 mm/sec, with 5mm of retract length.

Sebastianv650:
I assume you are speaking about the Slic3r feature and not the Marlin one.
200mm/s seems to be quite high in my eyes. Usual values I'm used to are around 40-70mm/s max. extruder speed. At last if you get skipped steps, your speed is to high.
Also have in mind that you can't exceed the mm³/s limit of your hot end for a pressure prime movement, which is the same as a printing move.

VanessaE:
Hm.  I'll have to give it a try at the lower speed then.

Maybe it would be a good idea to add a setting for max speed of advance (at least forward)?

lordofhyphens:
Isn't that covered by your max speed of the E axis?

VanessaE:
You would be correct, were I trying to exceed the motor's max unloaded speed. :-) 

My motor presently has no gearing, so it's weaksauce even with a MK8 hob, but while extruding through the nozzle into free air, it can move the filament at about 7 mm/sec before it starts to skip steps.  I'm not sure one way or the other if the motor's having any problems at 200 mm/sec, but I don't hear it skipping steps when I'm not using pressure advance.

My Google-fu fails me today, so I'm not sure how that translates into actual printing speeds, but since a normal restart moves the same length of filament as the preceding retract, I wouldn't think there would be any significant load on the motor, regardless of retract speed.

lordofhyphens:
With pressure advance it is deliberately overextruding for a brief period
to rebuild pressure. Odds are it is either too much, there is a bug in the
code (I don't know if the code/math matters if the flow units are mm^3 and
not mm for example), or your weaksauce motor will never be able to give you
the torque you want without some serious gearing work :(

VanessaE:
Well I'm going to eventually do something about gearing, but as for the pressure advance, my point is that it should pressurize-after-restart at more normal print-like speeds, not at the retract speed.

VanessaE:
Also bear in mind that Slic3r has a default speed of 40 mm/sec for retracts, which is still going to be far higher than a 0.4mm nozzle will allow for even with a huge gear reduction at the extruder.

Sebastianv650:
Slic3r takes the retract speed for pressure advance lines, see here: https://github.com/alexrj/Slic3r/issues/3283
It's true that a pressure build up movement should have another speed value than normal retract, but same is true for nomal prime movements after a retract. A "unretract" at 40mm/s can only work due to the flex in the filament: During the retract, the half-molten filament gets streched and it will cool down. During the prime, the now longer (and thinner) filament is forced against the molten filament in the nozzle which can lead to skipped steps in special cases. Thats why FW retract in Marlin has a default prime speed of only 8mm/s.

lordofhyphens:
We could add an option for a prime speed for advance, default it to 8mm/s

VanessaE:
The problem that I was having with apparent skipped steps here seems to be resolved - I'm not sure a separate pressurize speed is needed now.  The cause was a clogged nozzle, but of an unusual enough nature that I simply had not begun to consider the possibility (more on that another time).  Meanwhile, my adventure with pressure advance continues: #3457

alexrj:
Do you all agree this is mergeable? :-)

VanessaE:
It was good last time I tried, however you need to make sure that it never generates speeds greater than what the user has already configured.  Related:  make sure wiping speed is similarly controlled.

hexane360:
This has been generating the correct g-code for me. It definitely fixes the speed override problem. I don't see why any debate about the useability of Pressure Advance should affect this PR. The feature is currently in mainline Slic3r, it's just implemented incorrectly.

Regarding VanessaE's concerns, the code changes don't affect how Pressure Advance determines priming speeds. It only resets the feedrate after the extruder has been primed.

lordofhyphens:
Mechanically it looks fine to me, and @hexane360 has a point here. I will squash and merge. If it all goes horrid, we can revert ;-)

Issue #3095, opened 2015-10-16
Feature Request: distance between support and object


mrvn:
I'm printing a 40 teeth GT2 pulley with a hole for a M8 nut in the bottom (no, I can't print it upside down). The support being added is placed all around the inside of the hexagonal hole just a miniscule distance from the object. This causes the support to fuse with the object during print making it hard to impossible to later remove the support. It would be so much better if there were a setting for a minimum horizontal distance between the object and support.

KiteLab:
How about adding a one layer diaphragm on top of the nut, and then printing without support?
If that works well enough depends on your ability to print bridges.

pyrophreek:
Sounds like one of those things that is relatively part specific, so better to include the support in the model rather than rely on the slicer to generate it.  
Still, it is an interesting idea.

mrvn:
I don't think this is part specific at all. I had problems with the support being too near my object in basically every case I used support. Unless the support is inside a hole you can usually break it off but more horizontal clearance would be better. I'm fine with printing a 1-3mm bridge so I don't need support to be 0.1mm from the object.

lonelymyp:
Fuse with the object arise from problems with the mechanics. check belt tension, etc.
or do the jumper, similar like here: http://www.thingiverse.com/thing:135799
they are easy to break down after printing.

mrvn:
It might be the fault of bad mechanics but that's the quality I have. So it falls to the software to work around it. (Not sure what that extruder link was ment to be)

lonelymyp:
this is an old problem. but Alex does not want to solve it :(
https://github.com/alexrj/Slic3r/issues/2916

lonelymyp:
thin membrane then easily removed. model can be printed without the supports, or may be added to the support that is easy to break off.
![default](https://cloud.githubusercontent.com/assets/2085051/11165655/0cf703a6-8b26-11e5-8cfc-2dbbd8d443d0.jpg)

mrvn:
I tried adding "thin" membranes. But "thin" is relative. If the membrane is to thin the slicer skips it altogether. If it is to thick it will have many layers and has to be cut with difficulty or drilled. You have to adjust the membrane to the layer height in the slicer. And even then I always end up with walls 2 lines thick or membranes 2 layers thick. Most of the time all that would be needed would be 2 or 3 strands for the rest to hold.

lordofhyphens:
Duplicates #2916

lordofhyphens:
[...]

Issue #854, opened 2012-12-11
add setting value in GUI for  "distance_from_object"


funbart:
It would be great to have direct access for the value in the distance from object line:
 
my $distance_from_object    = VALUE * $flow->scaled_width;

In object.pm

Although I know have Perl installed on my computer because of this ;-)

Bart

lordofhyphens:
Closing, please see #1539 for further discussion (and see PR for workaround/fix)

Issue #3298, opened 2016-04-06
horizontal contact distance for supports ?


jfpion:
slic3r 1.2.9
win 7 64

the pb with support with some plastic is th supports stick to well to the part horizontally. there is a distance  Z contact tunable but not a horizontally distance contact avaible to tune

it could be a new feature

specially useful for the small intricate supports
thank you

lordofhyphens:
Duplicate of #2916

Issue #2916, opened 2015-06-15
Feature request : Choose the XY spacing between support and object


bitouns:
In slic3r the spacing between support and object have a default value. But with some prints you can need to change this default value to get a better result .It also depend of the material you are using : supports with PLA don't need the same spacing as supports with ABS.
If it's a default value may be it's easy to just add a little option in slic3r support interface to chose this value.
What do you think ? Is it possible ?

lonelymyp:
It's an old problem. 
In my case, support stick to the perimeter, especially in the corners, where excessive extrusion, due to bowden and acceleration.

bitouns:
Yes so if you could choose the distance between support and object you would not have this problem.

dropinsight:
Also make the z-distance to the support variable, not either soluble 0mm or detachable 0.2mm.

lonelymyp:
It variable, you just have not tried ;)

![default](https://cloud.githubusercontent.com/assets/2085051/8258782/70be6224-16c0-11e5-984a-c0cc9bede8cb.jpg)

bitouns:
Yes it possible for Z but not for XY

alexrj:
I dislike starting from solutions without having discussed concrete test cases. I usually look at a test case, then I try to identify the problem, then I try to figure possible solutions. Let's do it this way, and let's not skip steps.

"XY spacing" means little to nothing actually.  Support material is NOT supposed to have horizontal contact. There should be NO contact in XY. We only have vertical contact (Z). The XY distance between object and any projected support material should be large enough so that there's NO contact, but large enough so that the upper support material layer can be supported. This should be algorithmically determined, and I really don't see why this algorithm would change for each print or why user should decide to use a different value.

lonelymyp:
large nozzles is a problem. Support sticks at all angles.
but even with small nozzles is not so good ...

moment ago specially printed the worst-case scenario:
![img_0152 1](https://cloud.githubusercontent.com/assets/2085051/8261017/eea40874-16cf-11e5-88a4-df5b7d90d344.jpg)


I want to quickly print, so I have to use some acceleration and bowden
Therefore, in the corners there are thickening. they adhere to the support.
I almost always use a knife after printing if the item has been with support, If only I could increase the XY distance between object and support material +0.2 mm, everything would be fine :)

an alternative option is to change the logic of adding support, let support goes around the corners of large radius.
[...]

PR #3302, opened 2016-04-09
Tweakable xy supports

lordofhyphens:
Added option to increase XY supports (addresses #2916 and #3298), through a new config option and adjusting the clip_with_object and the contact layer. Defaults to 0 (current behavior), and may cause weird things to happen with support generation if set high. 

May be useful to adjust the minimum span to support to include this value as well.

### DIFF
--- a/lib/Slic3r/GUI/Tab.pm
+++ b/lib/Slic3r/GUI/Tab.pm
@@ -823,7 +823,7 @@ sub _update {
         for qw(support_material_threshold support_material_pattern
             support_material_spacing support_material_angle
             support_material_interface_layers dont_support_bridges
-            support_material_extrusion_width support_material_contact_distance);
+            support_material_extrusion_width support_material_contact_distance support_material_xy_spacing);
     $self->get_field($_)->toggle($have_support_material && $have_support_interface)
         for qw(support_material_interface_spacing support_material_interface_extruder
             support_material_interface_speed);
--- a/lib/Slic3r/GUI/Tab.pm
+++ b/lib/Slic3r/GUI/Tab.pm
@@ -467,6 +467,7 @@ sub build {
         max_print_speed max_volumetric_speed
         perimeter_speed small_perimeter_speed external_perimeter_speed infill_speed 
         solid_infill_speed top_solid_infill_speed support_material_speed 
+        support_material_xy_spacing
         support_material_interface_speed bridge_speed gap_fill_speed
         travel_speed
         first_layer_speed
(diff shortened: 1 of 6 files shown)


lordofhyphens:
This should also fix #3095.

Obviously I can't guarantee it will do what you expect w/r/t support generation everywhere for all values.

lordofhyphens:
@alexrj I'd appreciate any comments from you on this (experimental in nature) feature.

lordofhyphens:
This would probably also fix #854 if I am reading it correctly.

lordofhyphens:
https://github.com/alexrj/Slic3r/files/336282/slic3r-most_prs.zip

Experimental build for win32 for testing

lordofhyphens:
#1539

lonelymyp:
I am one of those who have had problems, but I can not compile, you have compiled file?

Error: The locale codeset (cp1251) isn't one that perl can decode, stopped at Encode/Locale.pm line 120.
Compilation failed in require at Slic3r.pm line 71.

lordofhyphens:
@lonelymyp see link above https://github.com/alexrj/Slic3r/files/336282/slic3r-most_prs.zip

It's packaged, so I can't guarantee it'll work on all systems for all codepages.

lordofhyphens:
@alexrj this PR is definitely experimental (as in results may not always be what is desired). If you have some interest in it as a feature I can continue to look for bugs on it.

There has been a lot of calls for better control over support generation.

bubnikv:
Thanks @lordofhyphens for the work.

Comparing Slic3r with Simplify3D, the Simplify3D has two settings to configure the horizontal offset:
1) XY offset
2) Vertical (above / below) offset in number of layers.

Slic3r does not do 2), so even if there is a gap in the layer, it will fuse to the layer above or below AFAIK.

Another issue is the extrusion height of the bottom interface layer, which is tune for bridge flow, but extruded as normal flow. This probably leads to corners of the extrusion not sticking well enough and being dragged into a covex corner of an object.

lordofhyphens:
@bubnikv what Slic3r does do is offset Z in mm, and it does honor it for interfaces. What it does also do, however, is enforce a minimum gap of {nozzle_width} (0 is special-cased here to avoid this), the setting in the config is actually {nozzle_width} + value, not value (as is often assumed).  There's a related PR out to change this behavior.

I haven't looked too closely at the flow attributes for the interface layers (assuming you are referring to  the furthest layer of the interface against the underside of an object, not the closest layer of support against a top surface of an object.

lordofhyphens:
As this is a hack on the current supports and we're looking to replace them, better to not add and remove options that people will complain about ;)

Issue #3134, opened 2015-11-20
Slowdown not working


evgs:
I made a test to verify slowdown:
https://www.dropbox.com/s/tmmy28etrc1z6g7/pyramid%20test.stl?dl=0

This is pyramid with square base, side=10mm

To make gcode be easy to read I have infill disabled, only 2 perimeters.
Also retraction is disabled.
Slowdown if time less than 15 seconds, and minimal speed is 3mm/s.

Here is my config:
https://www.dropbox.com/s/jckuf6rhr5rs31b/config.ini?dl=0

and resulting gcode:
https://www.dropbox.com/s/3pab11ch8izeko4/pyramid%20test.gcode?dl=0

After looking for speed in gcode I see F3300 for printing and F12000 for travelling (for middle layers with no infill).
F3300 is 55mm/sec - this is small perimeter speed from my config. Also F12000 is 200mm/s.
This speeds was applied for every layers (except base layer)

I know my test geometry - total printing path in layer should be less than 10mm * 4sides * 2perimeters = 80mm. Travel movement should will be less than 14mm (diagonal movement).
So total printing time I've got for layer is less than 80/55 + 14/200 = 1.5second.

I expected to see printing speed something like 80mm/15s = 5.3mm/s. Higher than minimal speed 3mm/s.

Acceleration was not set in my Slic3r config, so not taken into account.

Or I count wrong?

p.s. Slic3r version 1.2.10-dev 80620a5e94e476876c933560acc1c347c7baea7a , running on Gentoo Linux 64 bit

PR #3303, opened 2016-04-10
Fix layer cooldown time

hyperair:
The recent GCode writer changes which put the speed changes on a line of
their own have caused the layer time slowdown to be ignored by the regex
in CoolingBuffer.pm.


lordofhyphens:
Is there a test case so I can see the difference between the effect of this PR and the previous setup?

hyperair:
Issue is at https://github.com/alexrj/Slic3r/issues/3134

lordofhyphens:
All checks are passing and it looks like it works and doesn't cause any other problems from my limited testing. Going to merge this in, @alexrj for 1.3.0

Issue #3325, opened 2016-05-12
No slowdown


MacFlieger:
### Version
1.2.9-243-g510ca9f

### Operating system type + version
OS X 10.11.4

### Behavior
I have enabled "Auto cooling" and slow down printing if the print time is below 20s with minimum print speed of 1mm/s.
With this configuration the print slows down to approximately 20s on small objects if I am using Slic3r 1.2.9.

But if I am using the above mentioned actual developer version there is no slowdown.

It is easy to replicate, even as dry run:
- cube with 20mm size
- perimeter:1, no infill
- speed for perimeter 60mm/s
- one layer is 80mm long

- without auto cooling:
layer time around 2s

with auto-cooling (slow down if below 20s, min print speed 1mm/s):
layer time is again around 2s (no slow down), but should be around 20s.

Mikk36:
This was supposedly fixed with 510ca9f9e2f0e550ee3733510f7738e6ca0128aa

MacFlieger:
I know, but I am using this version [1.2.9-243-g510ca9f](https://github.com/alexrj/Slic3r/commit/510ca9f9e2f0e550ee3733510f7738e6ca0128aa).

Mikk36:
So what's your point, if it has been fixed already and will be incorporated into the next release?

MacFlieger:
As stated I am using this fixed version [510ca9f](https://github.com/alexrj/Slic3r/commit/510ca9f9e2f0e550ee3733510f7738e6ca0128aa), but there is no slowdown.
I have expected, that if I am using this fixed version, the slowdown is working.

Mikk36:
Oh blimey, I must have been blind not noticing the checksum version being the same.
Sorry for the confusion :)

Mikk36:
Can't replicate the issue though.
https://dl.dropboxusercontent.com/u/54281287/20mm-box%20%282%29.gcode

Also, how are you seeing your version as being 1.2.9?
It should be 1.3.0 already.

MacFlieger:
Strange.
This is my GCode for a 20mm Cube, 1 Perimeter, no infill, no top/bottom solid, 60mm/s, auto cooling 20s down to 1mm/s.
[slowdown.gcode.zip](https://github.com/alexrj/Slic3r/files/261031/slowdown.gcode.zip)

The speed for the perimeters is always 60mm/s (F3600).

> Also, how are you seeing your version as being 1.2.9?
> It should be 1.3.0 already.

[...]

PR #3326, opened 2016-05-12
Fix cooling not working if !gcode_comments

hyperair:
The cooling markers were being passed into GCodeWriter::set_speed() as a
comment which were being ignored if gcode_comments was false.

Fixes: #3325


hyperair:
Untested thus far. Someone wanna test?

hyperair:
Okay, tested, and it appears to work.

alexrj:
Thank you, it looks good. I'll merge it, but I want to write a regression test for that.

lordofhyphens:
@alexrj I'm going to merge this in now so I can close that issue related to this. Regression test can be in another commit :)

PR #3339, opened 2016-05-27
Removed "soluable" and "detachable" descriptors

### LINKED ISSUE #3338: Instability in Slicer
### Version
_Version of Slic3r used goes here_
Slicer 1.2.9 for mac

OSX El Captain 10.11.5 (15F34)

### Behavior
- When I am entering the manual value in the fields where supports configuration in contact z distance, to enter the value if I exit the setup and go back or save, the value returns to 0.2 (detachable). Sometimes i need a different value of 0.1, 0.3

*Reproduce problem:
 - Change value and don't remove the (detachable) name. Change the parameters and go back!
All Slic3r previous do that!

Sometimes i have a crash screen.
* _Expected Results_
 - Remove all value when click on the value area. Allow you to use a different value
 - Yes

[Archive.zip](https://github.com/alexrj/Slic3r/files/285744/Archive.zip)

lordofhyphens:
These seem to be causing some issues with editing that text box. Should partially fix #3338.

### DIFF
--- a/xs/src/libslic3r/PrintConfig.cpp
+++ b/xs/src/libslic3r/PrintConfig.cpp
@@ -1058,8 +1058,8 @@ PrintConfigDef::PrintConfigDef()
     def->min = 0;
     def->enum_values.push_back("0");
     def->enum_values.push_back("0.2");
-    def->enum_labels.push_back("0 (soluble)");
-    def->enum_labels.push_back("0.2 (detachable)");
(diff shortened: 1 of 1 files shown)


lordofhyphens:
This a thing that would be welcome?

alexrj:
I'd prefer to fix the issue itself rather than push this workaround

Issue #3220, opened 2016-01-26
[1.3.0-dev] Nonzero Max Volumetric Speed causing negative feedrates even if no automatic speeds are used


lordofhyphens:
Found this one, similar manifestation as #3045, except set max volumetric and make sure all other speed settings are nonzero.

alexrj:
How to reproduce?

lordofhyphens:
I'm honestly not sure anymore; I'll see if I can get it to pop up under current -dev.

JBKingdon:
I think I hit this today with master of Mar 19th. g-code is making very slow "printing" moves with negative feedrate and negative extrusion:

G1 X66.344 Y111.665 F12000.000 ; move to first infill point
G1 F-582474
G1 X66.142 Y111.941 E20.38560 ; infill
G1 X65.857 Y112.405 E20.38255 ; infill
G1 X65.777 Y113.020 E20.37909 ; infill
G1 X65.751 Y113.631 E20.37567 ; infill
G1 X65.750 Y130.364 E20.28220 ; infill
G1 X65.765 Y130.825 E20.27963 ; infill
G1 X65.813 Y131.316 E20.27687 ; infill
G1 X65.967 Y132.141 E20.27218 ; infill
G1 X66.103 Y132.638 E20.26930 ; infill

[stl-gcode-config.zip](https://github.com/alexrj/Slic3r/files/188427/stl-gcode-config.zip)

JBKingdon:
Ah, it seems to be more complicated than I thought. Setting max volumetric to 0 got rid of the negative F values, but I'm still seeing bizarre negative extrusion infill moves

G1 X74.181 Y141.190 F12000.000 ; move inwards before travel
G1 X74.772 Y140.725 F12000.000 ; move to first infill point
G1 F1200
G1 X74.971 Y140.865 E17.58194 ; infill
G1 X75.452 Y141.150 E17.57882 ; infill
G1 X76.036 Y141.224 E17.57553 ; infill
G1 X76.629 Y141.249 E17.57221 ; infill
G1 X93.361 Y141.250 E17.47875 ; infill
G1 X94.165 Y141.205 E17.47425 ; infill
G1 X94.686 Y141.128 E17.47131 ; infill
G1 X95.050 Y140.890 E17.46888 ; infill
G1 X95.446 Y140.590 E17.46611 ; infill
G1 X97.104 Y140.033 F12000.000 ; move to first infill point

lordofhyphens:
[...]

Issue #3045, opened 2015-08-19
[1.2.9-dev] Slic3r autospeed causing negative F values to be output in gcode


lordofhyphens:
On the provided STL, setting autospeed causes negative speed values to be used in output gcode.

I can confirm it doesn't have this behavior on the 5mm calibration step cube on youmagine/thingiverse

Test stl and config:
http://www.2shared.com/file/ckhcBgcM/negative-accel.html

W3DRK:
I have also witnessed this using Slic3r built from the current source.  It has occurred on every STL I've tried slicing so far, and causes my print controller to freeze.

Here's an example snippet of offending GCode:

G1 Z0.450 F7800.000
G1 X129.694 Y121.278 F7800.000
G1 Z0.250 F7800.000
G1 E4.00000 F6000.00000
G1 F-360.595
G1 X106.300 Y121.278 E5.15010
G1 X106.296 Y120.642 E5.18137
G1 X109.185 Y120.639 E5.32340
G1 X109.730 Y120.572 E5.35041

koghe:
I have the same issue https://github.com/alexrj/Slic3r/issues/3108#issuecomment-156864235
was some days ago, not tried with the current git because a don't want to clog my nozzle again...

![](https://cloud.githubusercontent.com/assets/745801/11170995/51dcabe2-8be4-11e5-965e-c08b2d7e8f40.jpg)

remove autospeed slicing is correct 
![](https://cloud.githubusercontent.com/assets/745801/11171178/d16267d0-8be9-11e5-81b4-134b2c40dd0d.jpg)

ini : https://www.dropbox.com/s/zeu3vmjcnhtqixz/config-autospeed-bug.ini?dl=0
stl : https://www.dropbox.com/s/eb7a36wj5c33p9g/coolingFan-new.stl?dl=0
gcode : https://www.dropbox.com/s/ut30b0o0n4tvznh/coolingFan-new.gcode?dl=0

lordofhyphens:
I think part of it is the model itself, some ratio of infill to perimeters. I think you can make it happen more by setting perimeters to be some multiple of extrusion of the perimeters (say 0.5mm perimeter and 1-2mm infill).

alexrj:
I tested @lordofhyphens's STL and config on 1.2.10-dev (stable branch) and I couldn't find any occurrence of the "F-" string in the G-code. Will test on master.

alexrj:
[...]

PR #3341, opened 2016-05-27
Workaround #3220

lordofhyphens:
Workaround to at least enforce no surprise negative feedrates on extrude moves.  What really needs to happen, I think, is there should be an assertion (or CONFESS()) call so it can be debugged up at where it is generated. However, that's probably less than ideal for a majority of users, so this omits the speed change command if the feedrate is negative or 0 (both of which are illegal so far as I know). 

Fixes/works around #3220 and #3045


lordofhyphens:
Reworked it to move the logic into GCodeWriter, did so to keep it identical to -stable patch.

alexrj:
Let's fix the issue, not slow down execution ;)

Issue #3295, opened 2016-03-27
Gapfill causing large negative feedrates in output


lordofhyphens:
### Version

Slic3r 1.3.0-dev 9576973b578f0c39f8cbd73c225187ba65b6cef4
### Operating system type + version

Debian Linux 8 x64
### Behavior

Output gcode contains large negative feedrates before all gapfill moves. 

_Expected Results_
No feedrate should <=0, because it's impossible.

 _Actual Results_
*Feedrates like G1 F-199232 or G1 F-3.58987e+06

_Is this a new feature request?_
#### STL/Config (.ZIP) where problem occurs

[negfeedrate.zip](https://github.com/alexrj/Slic3r/files/191047/negfeedrate.zip)

alexrj:
I just tested this with current master and issue looks gone. If you have a chance to test yourself we can close this issue.

lordofhyphens:
Well my test case no longer shows the issue.

PR #3342, opened 2016-05-27
Workaround to avoid writing negative extrusion feedrates in gcode.

lordofhyphens:
Writes a comment instead.

This is probably better than the PR in the other branch.


lordofhyphens:
Should "fix" #3220 and #3045

lordofhyphens:
#3295 should also be fixed by this.

alexrj:
Let's fix the issue instead of merging this :-)

PR #3346, opened 2016-05-29
Added a short OpenSCAD utility

lordofhyphens:
Generate a model that describes layers for incorporation as a modifier mesh. Parameters are in the OpenSCAD script. 

Main improvement I could see to this is porting it to OpenJSCAD so we can take advantage of the editable boxes.


lordofhyphens:
Merging this in myself, as it's useful and it doesn't hurt anything else. I might add a JSCAD version later if I have the time.

Issue #3293, opened 2016-03-26
[Proposal] More efficient GCode export


cangelis:
Current GCode export method sometimes takes forever when there are many islands in the input. Please correct me if I am wrong, this is the pseudo code of current flow of perimeter generation:

```
// perimeter generation
// part of process() of PerimeterGenerator
foreach $island in $region {
    // perimeter collection is the collection that holds all the perimeters of the layer
    $perimeterCollection->append(generate_perimeter($island));
}
// gcode export
// part of process_layer() in Slic3r/Print/GCode.pm
// group extrusions by extruder and then by island
foreach $perimeter in $perimeterCollection {
    // loop to find out the island of the perimeter
    foreach $island in $layer {
        if ($perimeter is in $island) {
             // group $perimeter
        }
    }
}
```

If we say perimeter count = p, and island count = i
Current algorithm is `O(p * i)`

There is no problem if there are only a couple of islands but if there are hundreds or thousands of islands, the export may never end.

For example this model: http://www.thingiverse.com/thing:399474 has about 1000 islands in each layer so there are 1000 perimeters in each layer. The algorithm needs to iterate 1000*1000 times to group the perimeters (same for infills).

What I propose is to push each perimeter to the collection in its own island. So only one iteration would be enough. Pseudo code:

```
// perimeter generation
// part of process() of PerimeterGenerator
foreach $island in $region {
    $island->perimeterCollection->push(generate_perimeters($island));
}
// gcode export
// part of process_layer() in Slic3r/Print/GCode.pm
// group extrusions by extruder and then by island
foreach $island in $layer {
    // group $island->perimeterCollection
}
```

So the algorithm does not need to iterate to find island of the perimeters again because perimeters are already in the island object as a collection.

If we say island count = i

Current algorithm is `O(i)` rather than `O(i*p)`

I really would like to create a pull request rather than just explaining it but I am not good at Perl and XS, anybody knows Perl, XS and codebase can implement this in an hour I think.

alexrj:
Thank you for taking your time to contribute with this idea. It makes sense, of course. I remember looking already into this, and then giving up. I think there's a reason behind the current behavior, but I don't remember it off the top of my head and I'd need to check the code. Will do.

cangelis:
Thank you. I have been researching about the performance issues of gcode export and this is my conclusion. I also try to implement my proposal but looks like it will take some time. 

Please let us know if you make any progress.

lordofhyphens:
[...]

Issue #2790, opened 2015-04-17
Slicer does nothing


pOmelchenko:
Hi! Yesterday noticed a problem with a long work slicer.

Dispatched slicer model and for a long time export gcode do nothing. Wherein the disk file is not increased.

Here is a model that tried to cut

http://www.thingiverse.com/thing:762168/#files

However, other models are cut properly. This model is opened and exported to kisslayser normal. Without mistakes.

pOmelchenko:
http://pastebin.com/banveS6f

backup config

pOmelchenko:
Found the cause of the problem, the thing functions "avoid crossing perimeters". I turned off this feature, the file is cut in seconds

lordofhyphens:
Would also have been very helpful to know the version you were using. There
are some known issues w/r/t "Avoid Crossing Perimeters".

"Nothing unreal exists." - Kiri-kin-tha's First Law of Metaphysics.

pOmelchenko:
I tested and stable 1.1.7 and 1.2.6 on a pilot.

As it turned out it it is certainly doing something, but it does so very slowly.

After I cut model (link above) without this feature, the size of about 14MB gcode.

And when you use this function the file size up to 1.5Mbit grew 30 minutes.

pOmelchenko:
Today, the compiled version 1.2.7, the problem with the Save function. STL file slice for a long time, while vyrostaet total consumption of system resources.

VanessaE:
This is still a problem with current code.  To gauge the speed, give this STL a try:

http://www.thingiverse.com/thing:1271651/#files

The initial trangulated mesh and perimeters steps take a bit, but when it goes off into the final exporting-to-gcode step, it just .... stops.  It does eventually finish exporting, and the result looks okay in Pronterface (loaded it there just to see the resulting travel paths).

lordofhyphens:
@VanessaE did you check with the PR for streamlined Gcode export by
@bubnikv or just 1.2.9 or just current master?

"Nothing unreal exists." - Kiri-kin-tha's First Law of Metaphysics.

VanessaE:
Just current master; I didn't remember that PR.

lordofhyphens:
Okay, just wanted to be sure.

"Nothing unreal exists." - Kiri-kin-tha's First Law of Metaphysics.

VanessaE:
[...]

PR #3355, opened 2016-06-03
Fix of git #3293: More efficient GCode export

lordofhyphens:
This fixes two problems of repeatedly copying data from the C++ layer
to perl, by factoring the copying out of an inner loop.

Actually done by @bubnikv. I cherry-picked the commit out of his rc 1.3 branch from the prusa3d repo.

Addresses #2790


lordofhyphens:
https://github.com/alexrj/Slic3r/files/336282/slic3r-most_prs.zip this PR is in this, please test and provide feedback

lordofhyphens:
#2790 should be mostly addressed by this PR as well.

Martronic-SA:
I've had the same problem and it took me some time to understand that it was coming from "Avoid crossing permeters"
That feature sounds great but it has to be optimised! It seems that for the path calculation it's taking the point that is the most far away from all the surrounding permieters instead of going in straight lines.
The result is the nice path in red in the picture below:

This is nice but not optimised at all! Very long to generate, huge output file and also 2-3 times printing duration.

lordofhyphens:
The general 2d routing problem is NP Complete, IIRC. There is no optimal
solution available that doesn't take exponential time (or space).

If someone  can work out a better heuristic for the average case but there
will always be corner cases that take forever.

kefir-:
Perhaps "avoid crossing perimeters" should be considered an advisory option, and not a requirement? That way you can make some heuristic as you say, and if it takes too long to calculate or causes some sort of troublesome move (too long, for example), just cross the perimeter instead. Or lift instead?

PR #3343, opened 2016-05-27
Support multi sampled anti-aliasing with wxWidgets 3.0.3.

bubnikv:
This patch improves rendering of the 3D scenes. The improvement is significant on the 3D path simulation.

1) Increases resolution of the depth buffer from 16 bits to 24 bits. 16 bits cause some jagged lines at the  intersection of triangles. I believe the computers offering a 16 bit depth buffer only are already in the landfill.

2) Enables multi sample anti aliasing. This is finally supported by wxWidgets 3.1 (February 29, 2016) and the OpenGL support for creation of anti aliased windows has been back ported to wxWidgets 3.0.3 (currently in beta stage). We at Prusa Research have hacked the Alien::Wx and Wx::GLCanvas to support this feature. Please accept this patch to enable anti aliasing on Prusa builds. This change shall not have any effect with the official non-hacked Alien::Wx and Wx::GLCanvas modules.

Vojtech

PR #3356, opened 2016-06-03
Support multi sampled anti-aliasing with wxWidgets 3.0.3.

lordofhyphens:
Cherry-picked the changes into their own branch for clean merging. What follows below is the original post (#3343)

https://github.com/alexrj/Slic3r/files/295253/Patches.zip is required to test/view MSAA with wxWidgets 3.0.3
------------------
This patch improves rendering of the 3D scenes. The improvement is significant on the 3D path simulation.

1) Increases resolution of the depth buffer from 16 bits to 24 bits. 16 bits cause some jagged lines at the intersection of triangles. I believe the computers offering a 16 bit depth buffer only are already in the landfill.

2) Enables multi sample anti aliasing. This is finally supported by wxWidgets 3.1 (February 29, 2016) and the OpenGL support for creation of anti aliased windows has been back ported to wxWidgets 3.0.3 (currently in beta stage). We at Prusa Research have hacked the Alien::Wx and Wx::GLCanvas to support this feature. Please accept this patch to enable anti aliasing on Prusa builds. This change shall not have any effect with the official non-hacked Alien::Wx and Wx::GLCanvas modules.

Vojtech

### DIFF
--- a/lib/Slic3r/GUI/3DScene.pm
+++ b/lib/Slic3r/GUI/3DScene.pm
@@ -11,7 +11,7 @@ use List::Util qw(reduce min max first);
 use Slic3r::Geometry qw(X Y Z MIN MAX triangle_normal normalize deg2rad tan scale unscale scaled_epsilon);
 use Slic3r::Geometry::Clipper qw(offset_ex intersection_pl);
 use Wx::GLCanvas qw(:all);
- 
+
 __PACKAGE__->mk_accessors( qw(_quat _dirty init
                               enable_picking
                               enable_moving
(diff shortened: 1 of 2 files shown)


lordofhyphens:
@bubnikv Does it break anything if you haven't hacked 3.0.3? We need broad compatibility. If it makes new features work but doesn't break backwards compatibility I'd be okay with pulling it in just to make things look nicer for users with newer libs.

Requiring that libraries be patched in order for the program to not fail is not something I would pull.

bubnikv:
Joe, I have tested the modified Slic3r against the official wxperl and it
works. So the changes are backward compatible.
Vojtech

alexrj:
@bubnikv, did you contribute the changes back to the Alien::Wx and Wx::GLCanvas maintainers? I invite you to do so.

Despite this patch relies on changes to upstream modules that are not accepted yet, I'm okay with merging this pull request because this code can be included with no problems in the C++ port.

bubnikv:
I contacted the original author of the perl wx module. Mark Dootson no more
maintains his brain child. He advised me to ask on the wxperl user list,
which I did, but I did not receive any response. The wxWidgets maintainers
showed an interest though to find out why wxwidgets 3.1 break wxperl. I may
consult that issue with them in the feature, but as of today I am fine with
my custom patches. And another patch seems to come soon to disable smart
dashes in edit fields on OSX.

PR #3379, opened 2016-06-26
Changed contact distance to layer height + extra 

### LINKED ISSUE #2930: Contact Z-distance is double what is supposed to be
I have created a model to test support materials and noticed that Slic3r 1.2.9 has an issue when creating a 'Contact Z-distance' that is not equal to zero. For instance when the 'Contact Z-distance' is set to 0.2mm the distance is actually 0.4mm. This happens nearly for all the features I have created in that model. Here are some images to show what I am talking about (this is going to be a cylinder where the axis of the cylinder is parallel to the print bed):

Layer Height 3.2mm
cylinder3 2

Layer Height 3.4mm
cylinder3 4

Layer Height 3.6mm
cylinder3 6

Layer Height 3.8mm
cylinder3 8

### LINKED ISSUE #3376: Support material z-distance: wrong variable used
### Version
1.3.0-dev
1.2.9-265-gcb72967

Win7x64 (any)

### Behavior
Distance between support material and first object layer is calculated wrong: $nozzle_diameter is added but not usefull, if distance is set manually. Maybe I'm wrong, but please check this.

Current SupportMaterial.pm:
```perl
sub contact_distance {
    my ($self, $layer_height, $nozzle_diameter) = @_;
    
    my $extra = $self->object_config->support_material_contact_distance;
    if ($extra == 0) {
        return $layer_height;
    } else {
        return $nozzle_diameter + $extra;
    }
}
```

==> Maybe should be:
```perl
sub contact_distance {
    my ($self, $layer_height, $nozzle_diameter) = @_;
    
    my $extra = $self->object_config->support_material_contact_distance;
    if ($extra == 0) {
        return $layer_height;
    } else {
        return $extra;
    }
}
```
br
Hendrik

lordofhyphens:
Fixes #3376 
Fixes #2930

### DIFF
--- a/lib/Slic3r/Print/SupportMaterial.pm
+++ b/lib/Slic3r/Print/SupportMaterial.pm
@@ -976,7 +976,7 @@ sub overlapping_layers {
 }
 
 sub contact_distance {
-    my ($self, $layer_height, $nozzle_diameter) = @_;
+    my ($self, $layer_height) = @_;
     
     my $extra = $self->object_config->support_material_contact_distance;
     if ($extra == 0) {
--- a/lib/Slic3r/Print/SupportMaterial.pm
+++ b/lib/Slic3r/Print/SupportMaterial.pm
@@ -982,7 +982,7 @@ sub contact_distance {
     if ($extra == 0) {
         return $layer_height;
     } else {
-        return $nozzle_diameter + $extra;
+        return $layer_height + $extra;
     }
 }
 
--- a/lib/Slic3r/Print/SupportMaterial.pm
+++ b/lib/Slic3r/Print/SupportMaterial.pm
@@ -977,13 +977,9 @@ sub overlapping_layers {
 
 sub contact_distance {
     my ($self, $layer_height) = @_;
-    
     my $extra = $self->object_config->support_material_contact_distance;
-    if ($extra == 0) {
-        return $layer_height;
-    } else {
-        return $layer_height + $extra;
-    }
+
+    return $layer_height + $extra;
 }
 
 1;


lordofhyphens:
https://github.com/lordofhyphens/slic3r/tree/support-fix-3376 for people who are looking for a zip file to download of the source with the patch applied.

lordofhyphens:
[slic3r-most_prs.zip](https://github.com/alexrj/Slic3r/files/336282/slic3r-most_prs.zip)

Mikk36:
[27.06.2016 17:03:21] <LoH> Mikk36: could you add that to the PR so I don't forget it?

[27.06.2016 10:14:56] <Mikk36> LoH, initial testing shows it to be working just as expected
[27.06.2016 10:15:02] <Mikk36> But
[27.06.2016 10:15:18] <Mikk36> Let's take the bridge torture test by triffid as an example
[27.06.2016 10:15:27] <Mikk36> Bridge bottom height is 10 mm
[27.06.2016 10:15:36] <Mikk36> Layer height: 0.2 mm
[27.06.2016 10:15:52] <Mikk36> Nozzle diameter 0.4 mm
[27.06.2016 10:15:59] <Mikk36> Contact Z distance: 0.19 mm
[27.06.2016 10:16:45] <Mikk36> First support layer is 0.2 mm, second is 0.01 mm to immediately get to the targeted offset of 0.19 mm
[27.06.2016 10:17:09] <Mikk36> That is, second layer of support has a layer height of 0.21 mm
[27.06.2016 10:17:21] <Mikk36> Third is 0.51
[27.06.2016 10:18:24] <Mikk36> I'd say a good solution is to do a minimum of 0.5*nozzle width layers until the targeted offset is reached
[27.06.2016 10:18:41] <Mikk36> Instead of immediately jumping to the targeted offset
[27.06.2016 10:19:12] <Mikk36> You also need to fix support.t
[27.06.2016 10:19:17] <Mikk36> It fails with that pull request
[27.06.2016 10:19:37] <Mikk36> t\support.t .................... 25/27 # Looks like you failed 2 tests of 27.
[27.06.2016 10:31:23] <Mikk36> So for example with that specific case, it would lay down layers: 0.2 (first layer, always same as layer height), 0.2, 0.2 and 0.21 (final height 0.81), from there on it'd put down 0.3 layers as normal
[27.06.2016 10:31:52] <Mikk36> One thing I'm wondering though is how to handle a case where there isn't enough room to reach that target
[27.06.2016 10:32:28] <Mikk36> Just try to get as close as possible and deal with it?

bubnikv:
Unfortunately this issue is much more tricky than it seems. Currently Slic3r will generate a reasonably working to interface layers if perimeter bridging flow is enabled, but it will create a sticky bottom interface layer. This fix may improve the situation somehow by defining a thinner extrusion for the 1st bottom interface layer, but it is not a clean solution. A clean solution would need to enforce a predefined bridging flow at the 1st bottom interface layer, which is difficult to achieve in a single line change of the current support code.

lordofhyphens:
I think that part of it as well is that the current behavior is something of a UI and expectations failure, especially when using small layer heights.

leopoletti:
Regarding the default Z distance and value and the possibility to choose custom value indepandently to nozzle diameter or layer height why not use the same method as Cura?
https://ultimaker.com/en/resources/20422-cura-support-settings
For the normal UI user would only choose if support or not. 
For advanced UI, user would choose custom value or choose the default calculated value as for 0 and 0.2.

lordofhyphens:
That's not what this is for; there's an extra parameter inserted into all spacings > 0, the nozzle value. So if you use the default of 0.2 and have a 0.4mm nozzle, you have a real separation of 0.6mm 

That isn't exposed to the user and since 0 is truly 0, there's some breaking of user expectations here. It works though for a lot of layer heights on its own though.

However if you're trying to specify really small separations and your layer  height is tiny, you can end up with wider gaps than you expect between interface and part.

I believe it's the bridging layer used for the bottom interface that was the impetus for using the nozzle diameter, as @bubnikv pointed out.

Issue #3311, opened 2016-04-21
Enabling gap fill greatly increases print time w/ autospeed


jhnphm:
### Version
510ca9f9e2f0e550ee3733510f7738e6ca0128aa 

### Operating system type + version
Arch Linux

### Behavior
Autospeed enabled (speeds set to 0) and nonzero gap fill greatly slows down print speed compared to autospeed w/ gapfill = 0

#### STL/Config (.ZIP) where problem occurs
[3DBenchyTest.zip](https://github.com/alexrj/Slic3r/files/229265/3DBenchyTest.zip)

#### GCode diff screenshot ####
![image](https://cloud.githubusercontent.com/assets/1557031/14699271/799f72ee-0764-11e6-880b-e03833929ef9.png)

lordofhyphens:
While I'm not that surprised as the autospeed logic is trying to maintain pressure, could you try it with current -dev? There was some changes w/r/t cooling and the timers involved.

jhnphm:
I was running 510ca9f which already seems to include that and is the latest from master that I can tell. Adjusting the gap fill speed does not seem to improve times (tried to slow it down to 2 in order to decrease required pressure during gap fill and allow the rest to speed up didn't work, and neither does increasing the speed.

lordofhyphens:
Gap fill speed = 0 disables gapfill, so yeah it'd be an interaction between the logic to maintain pressure.

lordofhyphens:
Okay, near as I can tell it's calculating the minimum mm^3 for the different extrusion entities and those calculations look like they're including gapfill. The referenced pull request removes gapfill amounts from calculation in all cases. 

Please test and see if this helps you.
https://github.com/lordofhyphens/slic3r/tree/fix-3311-no-gapfill-autospeed-effects

lordofhyphens:
Yeah, it's fixed, at least with my quick test with your config and settings.
With the new PR the time is now 2hrs as guesstimated by Pronterface, down from 10hrs.

alexrj:
Wait, what's the point in using autospeed if pressure is changed all the time because of gap fill?

lordofhyphens:
Since pressure would be lost anyway on a layer change, if gap fill was delayed until the last print moves then it would not matter.

Additionally the ridiculously slow speeds causes heat damage to the print in practice.

hyperair:
[...]

PR #3387, opened 2016-06-30
min_mm3_per_mm() ignores gapfill extrusions. 

lordofhyphens:
Addresses #3311 
ExtrusionEntities with the gapfill roll aren't useful for comparing to much, and from what I can tell it's affecting the calculation for autospeed.

From what I can tell it doesn't change the extrusion amount and may also fix negative extrusion rates or similar problems. Please test :)


TTN-:
Compiles and builds fine for me.

lordofhyphens:
Yes, that is the current state with this PR attached.  Print tests I think are important atm, although nothing looks awry with the 3d preview.

lordofhyphens:
#3394 is a better solution I believe.

Issue #3390, opened 2016-06-30
Proposal: Conditional Gcode Syntax for Custom Gcode


lordofhyphens:
So with a simple post-processing script like the one below, I'm looking at proposing a simple grammar for some conditional statements in gcode. What I was thinking about was coming up with a simple grammar with references that can be added by the post-processor. It'd be based around parsing the numerical result and parsing everything around the statement and variable/value and running the comparison in perl. If the condition is false, the line is removed. Starting with a semicolon means that if it breaks the gcode itself isn't run. 

No consideration for else is currently proposed, and also planning to limit to single-variable, although I think the grammar could be extended to arbitrary numbers of evaluations.

Just in case, I want to avoid just calling eval() and do some basic sensitization to avoid someone doing weird things to your system through this mechanism.

[variable] is some slic3r variable and the value will be filled in. We'll use ; to delineate the blocks. 

```
;_if [variable] == value; gcode
```

Example, set the speed factor to 50% if the current_extruder is 2 (for toolchange or layer change or even start/stop gcode). 

`;_if [current_extruder] == 2; M220 S50`

if [current_extruder] is 2, then the processor would see:
`;_if 2 == 2; M220 S50`
and print:`M220 S50` in the resulting gcode.

I believe the leading _ in the keyword is important to make it less likely that an actual comment doesn't trigger the logic; additionally the leading ; (indicates the start of a comment for most, if not all, gcode parsers )  means that the default state of execution is no execution.

jaredabrandt001:
Conditional statements in gcode would be wonderful!  I think Cura supports IF statements, where when true, the line is included and if false the line is excluded.  I use Slic3r because it processes thin walls better.  I'd suspect this simple conditional argument could take care of 99% of the circumstances printer would encounter.  It would also greatly simplify wipe and prime code, and could easily be used to trigger certain operations (such as bed temp adjustments) at certain layers.  Something like:    IF ([layernum] == 25)  M190 S45;
It's my experience that bed heat always needs to be turned down after the first few layers; it makes the print less likely to pull off the bed and it saves a lot of electricity.

Is there any way to get slic3r to process these without the post-processing step?  It would be ideal if this were built-in, but I understand I might be asking too much.

Overall though, thanks for looking into this.  I think it will make Slic3r better!

lordofhyphens:
A post processor is the cleanest and easiest to debug. Slic3r's post
process facilities are decent and most information is all that information
is available.

[...]

PR #3391, opened 2016-07-01
added post-processing script to allow for conditional inclusion of gcode

lordofhyphens:
Addresses #3390

### DIFF
--- a/utils/slic3r-conditional.pl
+++ b/utils/slic3r-conditional.pl
@@ -1,0 +1,23 @@
+#!/usr/bin/perl -i
+# Conditional gcode post-processor
+# Written by Joseph Lenox
+# Executes statements beginning with ;_if
+# the condition must be evaluatable Perl. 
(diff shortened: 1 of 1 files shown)


lordofhyphens:
I need to do some more research, but here's a basic script in the utils directory to perform the conditional gcode. I need to do some more research to figure out what is considered to be in path at all times for Slic3r w/r/t packaging; which would let me add this the default list of post-processing scripts. 

I think additionally the interface for post-processing scripts needs to be cleaned up a bit, with a nicer UI for saving them and a clean way to reorder ones that are being applied.

lordofhyphens:
Syntax is going to change and this is instead going into the C++/XS side of the fence re: preferences @alexrj (and I agree enough to not waste time arguing about it). Closing; this PR does represent a large security hole in Slic3r; a maliciously crafted config.ini could execute code on your system.

Issue #3385, opened 2016-06-29
Lift Z option only works for first extruder


platsch:
### Version

1.2.9-257-g08a2775
### Operating system

x86_64 GNU/Linux
### Behavior

The `Lift Z` option always applies for the first extruder (id 0), values for `Lift Z` and `Only lift Z` for any other extruder are ignored. 
This behaviour is hard-coded in GCodeWriter.cpp: 
`target_lift = this->config.retract_lift.get_at(0);`

Is this a bug or is there any reason I overlooked?

lordofhyphens:
Suspect that it is a bug.

lordofhyphens:
I usually don't run into this problem myself as I run only a single extruder and use firmware retraction, which has a lift baked in as an option.
I suspect that it's usually not a problem for most people because the two hotends are close enough in requirements w/r/t z that it's hard to tell the difference as well. 

The other thing I can think of though to do it like this is to avoid head crashes, as on switch it would retract, lift for extruder 1, change tools to 2 and then use extruder 2's "lift amount" and restore. I'm not sure offhand how much impact actually fixing that to cover that case (which is pretty important :))

platsch:
I frequently use multiple extruders. For most plastic extruders this is not a real problem, but it is not exactly user friendly to have an option that only works in special cases.
However, for liquid extruders, i.e. syringe or pressure dispensers the different z offset is actually important. The obvious workaround is to increase the lift for both extruders, but since the plastic extruder often covers the largest share of printing time it is nice to reduce the movements for the slow z-axis as good as possible.

I locally fixed this by providing the correct extruder id to `get_at(0)` and didn't observe the head-crashing scenario so far. The `lift()` method stores the lift-distance and restores the old value even after switching the extruder. `unlift()` doesn't use any config values.

I can submit a pull request for this, but it's a simple replacement of the static extruder id to `get_at(this->_extruder->id)` in 3 lines.

lordofhyphens:
Go ahead and make the PR. I would rather you have the credit in the log.

We can put some tests together and see if it can be broken as well.

lordofhyphens:
@alexrj any particular reason lift always did the first extruder or cam this be merged in?

alexrj:
Code-wide it looks fine, and I appreciate the test case.

[...]

PR #3392, opened 2016-07-01
Bugfix: use Lift-z option for 2. extruder #3385

platsch:
Bugfix for #3385


lordofhyphens:
Looks good to me, earlier concern seems to be a non-issue w/r/t accidental head crashes. 

Could you augment https://github.com/alexrj/Slic3r/blob/master/t/retraction.t#L204 and add a regression test to make sure that it uses the selected extruder? Probably just a copy/paste with two different extruders defined instead of just one define the current extruder to be 2 and check to make sure it matches :)

platsch:
Yes, it's not forgotten, just give me a few days to find the time :-)

lordofhyphens:
Thanks :) I know how precious time is.

lordofhyphens:
Also, could you rebase this back onto master? It'd be nice to have the travis build passing :)

platsch:
Ok, I added the testcase you suggested as an additional case. Since this doesn't check for the actual values, I also modified the 1. test https://github.com/alexrj/Slic3r/blob/master/t/retraction.t#L52 to keep track of the lift value between extruder switches.

Rebased to master, compiling and testing runs fine, but I'm not able to execute Slic3r, fails with `Caught C++ exception of unknown type at /Slic3r/lib/Slic3r/Config.pm line 92.` looks like 30139fd64746f0a52ce489eb7f5e60965925f9fb breaks something...

lordofhyphens:
@alexrj had a commit that broke handling of config entries that aren't
enumerated in PrintConfig.hpp. The load/save was ported to C++ and
exceptions thrown weren't caught. You can rebase to the branch in #3432 to
figure out what is in your config file that Slic3r is choking on (then
rebase back to master when when it is sorted out).

platsch:
Ah, it was an option in my config, leftover from testing a different branch. Looks good, think you can merge the PR.

lordofhyphens:
Any comments/concerns, @alexrj, about this PR?

lordofhyphens:
Had forgotten to check the changes in my editor to make sure that spaces were used instead of tabs, everything looks good+clean.

@alexrj Unless there's a specific reason to use the old behavior (all retraction settings using the first extruder's settings) that hasn't been considered, I recommend that this get pulled at the earliest opportunity.

t/retraction.t
```diff
@@ -228,7 +228,7 @@ use Slic3r::Test qw(_eq);
     ok !!@lifted_at, 'lift takes place when above/below == 0';
     
     $config->set('retract_lift_above', [5, 6]);
-    $config->set('retract_lift_below', [15, 16]);
```
lordofhyphens: Any reason why this changed in the PR branch?

platsch:
Yes. With the old parameter set `[15, 16]`the `retract_lift_below` test doesn't fail for the second extruder even if only the first extruder config is used incorrectly. The test value for the second extruder must be lower for the `below` test and higher for the `above` test to test for the correct behaviour of the second extruder.

PR #3394, opened 2016-07-03
Separate fills and gapfills

### LINKED ISSUE #3311: Enabling gap fill greatly increases print time w/ autospeed

hyperair:
Separate fills and gapfills to allow excluding gapfill moves in the calculation of `min_mm3_per_mm` for autospeed. This fixes #3311.

Another approach to fix this issue is https://github.com/alexrj/Slic3r/pull/3387, but this PR attempts to avoid putting the gapfill filtering into `ExtrusionEntityCollection::min_mm3_per_mm`.


lordofhyphens:
Could you rebase this onto current master? Makes me feel better seeing the checks pass :)

I'd also appreciate a .t file; maybe add "gap fill" to the verbose comment output and check for it?

hyperair:
Yay it's green. As for the test, I'm thinking something more along the lines of testing gapfill + autospeed and ensuring that it doesn't slow down too much.

I also found, from my recent testing of autospeed on this branch, that thin-walls can also cause the same sort of autospeed slowdown. Not sure what the right course of action is for that though.

lordofhyphens:
I'm not surprised, given that autospeed calculates how fast it can go from the smallest area (everything else must be slower). I suspect external perimeters have an additional slowdown attached to them, but I haven't looked.

hyperair:
As far as I know, external perimeters only get a slowdown if set to a percentage of perimeter speed.

lordofhyphens:
This, by the way, is why I don't immediately merge my own PRs :) Sometimes other people find a better solution than the one I came up with.

alexrj:
Wait before merging this, let's first discuss #3311 better...

bubnikv:
The gap fill feature is problematic in regard to the constant nozzle pressure algorithms. I partially developed a feature, which does not keep the pressure constant, but allows a slow change of pressure. This sounds great, but it seems to break in case of gap fills with quick changes in the extrusion rate. I am not quite sure what to do now, the feature needs to optimize multiple opposing constraints, like the low jerk, high velocity, constant pressure in the nozzle, quick travel etc. My current impression is one shall apply some low pass filtering of the gap fill extrusion rates, but this needs to be verified with real prints first. 

So just disabling the gap fills in the constant volumetric feed algorithm will produce blobs in the perimeter after a gap fill, which is why the constant volumetric feed algorithm was developed in the first place.

PR #3395, opened 2016-07-04
Fix ExtUtils::CppGuess installation on travis

hyperair:
`ExtUtils::CppGuess` fails in its `t/011_makemaker.t` test because it passes `-lstdc++` before the `CppGuessTest.o` file instead of after, causing `libstdc++` to be completely excluded from the `DT_NEEDED` section of the generated `CppGuessTest.so` file, resulting in the error shown here: https://travis-ci.org/hyperair/Slic3r/jobs/142048019#L2781

See http://stackoverflow.com/questions/45135/why-does-the-order-in-which-libraries-are-linked-sometimes-cause-errors-in-gcc for more information on this.

This PR sets `LDLOADLIBS=-lstdc++` as a hack during the cpanm installation run to ensure that the flag gets passed after the .o argument.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ perl:
 install:
   # LDLOADLIBS hack to fix ExtUtils::CppGuess compilation (-lstdc++ needs to be
   # placed after CppGuessTest.o)
-  - LDLOADLIBS=-lstdc++ cpanm --installdeps --force --notest --verbose .
+  - LDLOADLIBS=-lstdc++ cpanm --installdeps --verbose .
 
 script: perl ./Build.PL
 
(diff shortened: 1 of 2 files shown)


lordofhyphens:
Mostly wish I knew why this has recently started to become a problem, but suspect it needs to be reported upstream as well.

lordofhyphens:
https://travis-ci.org/alexrj/Slic3r/jobs/142268952

It builds but Travis is interpreting the environmental variable as a command to be run apparently.

hyperair:
Huh, how odd. It was working before..

lordofhyphens:
Replaced by #3396.

PR #3402, opened 2016-07-07
Build + packaging script for Win32

lordofhyphens:
Provide Powershell scripts to permit automatic packaging of Slic3r for Win32 systems and enable automatic build/deployment of development branches.

This should make testing features easier for people who aren't particularly savvy with building on Windows and only use Windows systems.


lordofhyphens:
This will be squash-merged.

Issue #2693, opened 2015-02-25
Honeycomb infill is offset - 1.2.6


BJoey:
didn't find any similar issue so i'll open one:
since using Slic3r 1.2.6, the honeycomb infill layers do not always align correctly, as seen in the picture. Does not connect with perimeters as well and seems to get worse with smaller layer heights. this causes a great loss of stability and top layers are not properly filled due to the lack of material in the layers below.
Did I change any option to make this happen or is it a real bug?
![img_20150225_145450](https://cloud.githubusercontent.com/assets/11194145/6371886/4f59a2e8-bd00-11e4-9bb4-164f7c2a0bde.jpg)

a4jp-com:
I have the same problem.

alexrj:
@BJoey, see the guidelines. We need to reproduce your problem in order to help or take any action.

a4jp-com:
If BJoey can't post a file I can try and make a 3D model again with this problem. I deleted all the files I had before because they just looked ugly. I just had a land model with a honeycomb infill set at about 20 or 30 percent. Kind of a box with the bottom turned off.

BJoey:
Hey sry I can't figure out how to upload a file, but that proplem does not depend on a specific STL file. The picture above is a print of this one: http://www.thingiverse.com/thing:65168
I can post the GCode of one of those prints or Slic3r settings if it helps:


# generated by Slic3r 1.2.6 on Sat Feb 28 11:09:23 2015
avoid_crossing_perimeters = 0
bottom_solid_layers = 4
bridge_acceleration = 0
bridge_flow_ratio = 1
bridge_speed = 30
brim_width = 0
complete_objects = 0
default_acceleration = 0
dont_support_bridges = 0
external_fill_pattern = rectilinear
external_perimeter_extrusion_width = 0
external_perimeter_speed = 50%
external_perimeters_first = 1
extra_perimeters = 1
extruder_clearance_height = 20
extruder_clearance_radius = 20
extrusion_width = 0
fill_angle = 45
fill_density = 50%
fill_pattern = honeycomb
first_layer_acceleration = 0
first_layer_extrusion_width = 200%
first_layer_height = 0.3
first_layer_speed = 50%
gap_fill_speed = 20
gcode_comments = 0
infill_acceleration = 0
infill_every_layers = 2
infill_extruder = 1
infill_extrusion_width = 0
infill_first = 0
infill_only_where_needed = 0
infill_overlap = 15%
infill_speed = 60
[...]

PR #3410, opened 2016-07-11
Calculate actual flow for later inclusion as a ExtrusionPath for internal flow.

### LINKED ISSUE #1783: Infill way too sparse when using "Combine infill every"
Sorry, I originally posted this in the wrong thread:
https://github.com/alexrj/Slic3r/issues/1288

Combine infill every tooltip text:
"This feature allows to combine infill and speed up your print by extruding thicker infill layers while preserving thin perimeters, thus accuracy. (default: 1)"

Combine infill every actual behavior:
"Will produce sparse vertical infill by skipping a set number of layers. This can be used to speed up print times where the missing infill is acceptable."

To test issue: select 0.100 mm layer height, combine infill every 3 => two layers are skipped and every third is printed at 0.100 mm layer height

To fix issue: select 0.100 mm layer height, combine infill every 3, infill density 300 % => three layers of infill are combined into one 300 % infill layer

wp_20140216_002
Figure 1. Layer height 0.100, Combine infill every 3, Infill 100 %

wp_20140216_007
Figure 2. Layer height 0.100, Combine infill every 3, Infill 300 %

Slic3r 1.0.0.RC1 via Repetier-Host V0.95F Windows 7 64x

lordofhyphens:
Uses the correct layer height for the surface.

Fixes #1783 #2693

### DIFF
--- a/lib/Slic3r/Fill.pm
+++ b/lib/Slic3r/Fill.pm
@@ -229,6 +229,13 @@ sub make_fill {
             );
             $f->spacing($internal_flow->spacing);
             $using_internal_flow = 1;
+            # create the actual flow for internal flow that is used later.
+            $flow = Slic3r::Flow->new_from_spacing(
+                spacing         => $internal_flow->spacing,
+                nozzle_diameter => $flow->nozzle_diameter,
+                layer_height    => $h,
+                bridge          => 0,
+            );
         } else {
             $f->spacing($flow->spacing);
         }


lordofhyphens:
Flow rates before:
mm3: 0.218711 0.35
mm3: 0.125664 0.3
mm3: 0.125664 0.1

Flow rates after:
mm3: 0.439823 0.35
mm3: 0.376991 0.3
mm3: 0.125664 0.1

lordofhyphens:
Looking through the code as it is, it seemed appropriate to set the flow up in Fill.pm instead of trying to fix it in XS; the flow for other infills is also configured there as well.

Issue #2700, opened 2015-03-01
printer "has heated bed" option.


whosawhatsis:
It would be nice to have an option in the printer profile to specify whether the machine has a heated bed. With this option unchecked, all M140/M190 codes would be left out of the gcode, but the same  filament profiles could still be used for a different machine that does have a heated bed to produce those codes.

alexrj:
Will do.  :)

jonaskuehling:
Apologize my dumb question :) but does it hurt at all, if M140/M190 commands are sent to a non-heatbed-printer? Does it try to heat up infinitely then?

Never tried actually :)

alexrj:
Unfortunately yes!  (Or so it was last time it happened to me)

whosawhatsis:
M140 doesn't hurt, but M190 makes it wait for something that will never happen before proceeding, effectively locking up the printer.

kefir-:
Shouldn't the firmware handle (ignore) M190 if there's no heatbed?

jonaskuehling:
That's at least what I would have expected...

lordofhyphens:
There you go.

alexrj:
Merged, we have a "Has heatbed option".

PR #3415, opened 2016-07-13
Adds new has_heatbed option to capabilities

lordofhyphens:
Disables autogeneration of M140/M190 if it is unchecked. Implements #2700

Implementation of feature is done in Gcode.pm, with the other do/don't logic for setting bed temp.


lordofhyphens:
I realized that this needs a test, will update tonight.

lordofhyphens:
Test is added to verify that if has_heatbed is unselected, no M190/M140 gcodes are generated (and vice-versa).

alexrj:
Merged!

Issue #3360, opened 2016-06-03
Slic3r crashes on Windows when the Bonjour dialog is open / closed with OK and there are no Bonjour devices found


bubnikv:
### Version

Slic3r-1.3.0-dev
### Operating system type + version

Windows
### Behavior

Open the Bonjour dialog on windows. Nothing is found, even if there is a Bonjour capable device around (I have octoprint running). Click ok, Slic3r crashes.

Please cherry pick from github/prusa3d/slic3r
5573faec6b878a3f618618441eef070f4e7e8e95

Moved the Bonjour discovery to Tab.pm,
when no Bonjour devices found, a simple message box is shown.
Otherwise the Bonjour selection dialog with an empty list crashes on windows.

Thanks, Vojtech

PR #3419, opened 2016-07-13
Moved the Bonjour discovery to Tab.pm,

lordofhyphens:
when no Bonjour devices found, a simple message box is shown.
Otherwise the Bonjour selection dialog with an empty list crashes on windows.

Addresses #3360

### DIFF
--- a/lib/Slic3r/GUI/BonjourBrowser.pm
+++ b/lib/Slic3r/GUI/BonjourBrowser.pm
@@ -9,14 +9,10 @@ use base 'Wx::Dialog';
 
 sub new {
     my $class = shift;
-    my ($parent) = @_;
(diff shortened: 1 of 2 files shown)


lordofhyphens:
Code looks pretty clean. While I would recommend merging this, I'm curious if the search had to be moved into Tab.pm.

alexrj:
Hm, no, this makes Slic3r::GUI::BonjourBrowser not reusable as component. I'd rather investigate why an empty list crashes, as it shouldn't AFAIK...

lordofhyphens:
Alright, then. I'll close this and generate a Win32 build based on the branch and master (so that people who have this problem have a workaround).

Issue #3426, opened 2016-07-16
Repetier-firmware implements acceleration as M201/M202 seperately, not M204


lordofhyphens:
Repetier-firmware has a different gcode flavor than other Reprap firmwares (primarily Marlin). 
Definitely a case of https://xkcd.com/927/

This means that adjusting acceleration on Repetier firmware instead (possibly) changes PID values. 

Since https://github.com/repetier/Repetier-Firmware/issues/290 indicates that they aren't changing, we'll need to account for this hot mess.

lordofhyphens:
@alexrj this should probably be sorted out before the next release, as the current state of changing acceleration on Repetier FW is broken. I have a PR open for it but if you want to do something additional, say something along the lines of what you want and I'll add it to the PR.

lordofhyphens:
@repetier is there anything else that you know about that it handles differently than other firmwares that Slic3r might set? The attached PR covers acceleration.

repetier:
No I think only the acceleration stuff was changed some time ago in Marlin. M201/M202 date back to Sprinter I think. When marlin introduced M204 I was already using it for PID. Just be sure to use M201 which is for printing moves. I guess you never change travel acceleration.

- M201 - Set max acceleration in units/s^2 for print moves (M201 X1000 Y1000)
- M202 - Set max acceleration in units/s^2 for travel moves (M202 X1000 Y1000)

lordofhyphens:
The patch I have implemented currently sets both. I could probably do a
better job if I added a new field for travel moves. That doesn't do the
code cleanliness any favors though.

alexrj:
This was merged.

PR #3427, opened 2016-07-16
Added Repetier FW flavor, fixed accel gcode generation for it.

lordofhyphens:
Addresses #3426

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,7 +133,7 @@ The author of the Silk icon set is Mark James.
                             (default: 100,100)
         --z-offset          Additional height in mm to add to vertical coordinates
                             (+/-, default: 0)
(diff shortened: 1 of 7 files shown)


lordofhyphens:
X and Y will be set to the same value because we don't differentiate in the GUI settings between the X/Y accelerations.

lordofhyphens:
Still needs a short test to ensure that it generates the correct gcode output in the Repetier flavor.

I think the minimum test would load a default config, change the flavor to Repetier and set the accel to a distinctive value and then generate gcode. The validation would be that the number shows up in gcode and the line where it does show up is `M201 X### Y###`

lordofhyphens:
There, test also accounts for regression as well and passes :)

Issue #3430, opened 2016-07-18
1.3.0-dev crashes when loading 1.1.7 config bundles.


lordofhyphens:
### Version
1.3.0-dev 30139fd64746f0a52ce489eb7f5e60965925f9fb

### Operating system type + version
Linux Debian Jessie. 

### Behavior

Slic3r crashes with an exception when loading a config bundle from 1.1.7 with the following error:
````
Caught C++ exception of unknown type at /mnt/fiver/slic3r/lib/Slic3r/Config.pm line 92.
````
This renders it difficult, if not impossible, to follow up on old bug reports. I suspect it is crashing whenever it finds a key that it doesn't know about. 

Configuration file.
[config-117.zip](https://github.com/alexrj/Slic3r/files/368301/config-117.zip)

The problem was introduced in the referenced commit (or right before) when --load and --save were ported to XS. At minimum it should not be crashing Slic3r (previous behavior was that if a loaded config didn't have something it defaulted it).

lordofhyphens:
This is going to be considered "not a bug" then documentation will need to be written to "correct" user expectations.

lordofhyphens:
Okay, the exception thrown during  set_deserialize (and lower) isn't being caught at the load code, which is getting passed up to the Perl side of things. Because strings are being thrown (and not C++ exceptions), Perl has no additional information to give.

I don't think that an "old/invalid" configuration option should be a crash-level event in Slic3r. From a code-cleanup perspective, I also think all of those thrown exceptions should actually be exceptions derived from std::runtime_error (or something similar), as you can't reliably catch a string because it can throw too.

lordofhyphens:
Closed via #3432

PR #3432, opened 2016-07-19
Catch exception thrown when loading config files.

lordofhyphens:
Thrown exception has been changed to be std::runtime_error, includes regression test (program does not crash when loading unknown keys).
Sends error/diagnostic message to std::cerr.

Addresses #3430


alexrj:
Oh, I had actually fixed this in my local working copy but yours is nicely coded and has regression test which is lovely to have. Thank you.

lordofhyphens:
Thanks for merging, I would appreciate any comments you have for the
related issue I tagged you in re:exceptions ("I don't care so long as it is
clean" is a valid and helpful comment :) )

PR #3437, opened 2016-07-21
Add Ubuntu installation instructions

tareko:
Add installation instructions including easy to copy-paste commands for Ubuntu apt

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,18 @@ $ perl Build.PL --sudo --gui
 $ ./slic3r.pl
 ```
 
+### Ubuntu installation instructions
+If you wish to compile the source from within Ubuntu, here are some instructions tested on Ubuntu 16.04:
+
+```
+$ sudo apt install build-essential libgtk2.0-dev libwxgtk2.8-dev libwx-perl libmodule-build-perl git cpanminus libextutils-cppguess-perl libboost-all-dev libxmu-dev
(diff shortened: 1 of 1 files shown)


tareko:
I always have to go back and figure out how to do this each time I try to re-compile Slic3r on a new computer, so here are my compiled directions.

lordofhyphens:
I can't merge this, but it's feasible to add to the wiki: https://github.com/alexrj/Slic3r/wiki/Running-Slic3r-from-git-on-GNU-Linux/

I also disagree strongly with not using local::lib on Debian-based architectures, as it invariably breaks the related packages. This is been my experience time and again when building on package-based distros.

lordofhyphens:
For Debian and Ubuntu if you don't want to use local::lib imho is to unpack the Ubuntu package source (the debian directory) for slic3r 1.2.9 over a fresh checkout of the repository, and then installing the package.

lordofhyphens:
libwxgtx2.8-dev is also becoming rarer on modern Ubuntu distros, ran across someone who didn't have it on their xubuntu distro.

lordofhyphens:
The good news is that this prompted me to reorganize the wiki page on building on linux. The bad news is that I'm closing this pull request because it can't be merged.

tareko:
Thank you for doing this. I think having it in the wiki works well enough.

tarek : )

Issue #1986, opened 2014-04-28
Feature request: make bridging over infill selectablebased on infill density.


wolfmanjm:
Currently the first solid layer on top of an infill layer is always done as a bridge at bridging speed and bridging flow rate.

This is unnecessary if the infill density is around 30% and really slows the print down if the surface is large.

A nice feature would be to use normal solid layer speed and flow rate as the first solid layer if the infill was over a certain selectable percentage.

lordofhyphens:
Yeah, I can see where this would be useful, and it looks straightforward to implement. I'll have a branch + pull request up shortly if someone wants to give a hand testing :)

alexrj:
Hm, I had actually addressed this issue in 38a9e32a287c4a601e256ebd96533f68c0c8dcc0 (after this issue report). And I have one more improvement for making this logic even stricter.

PR #3468, opened 2016-08-19
Advanced option: do not use bridging infill above a specified fill density

lordofhyphens:
Addresses #1986 #3483
If someone can think of a good auto test for this I'd be appreciative of it. I verified it by looking at the output for the generated layers (visually).


lordofhyphens:
Not breaking existing expectations.

bubnikv:
This is certainly a good thing. Another extension would be to make the bridging sparse. This is what the kisslicer does.

Now we are getting more and more configuration values, which makes the UI cluttered. The "new" cura handles the UI clutter by a configuration of configuration UI. I wonder where Slic3r is going in this regard.

alexrj:
I had actually addressed this issue in 38a9e32a287c4a601e256ebd96533f68c0c8dcc0 (after this issue reports), and I have one more improvement for making this logic even stricter.

alexrj:
I'm proposing an alternative thing in #3627, comments very welcome.

lordofhyphens:
Alternative merged, closing :)

Issue #2979, opened 2015-07-05
Feature Request: Force Full Reslice


the-edge01:
It would be useful to have a button to force a full reslice of an object if something goes wrong and you want to restart the entire slicing process.

pyrophreek:
I agree, and really what I would like to have is a button to press to view the sliced gcode before exporting it, but without having to turn on background processing.  I love the idea of background processing, but if you are planning to adjust a few parameters, sometimes it slows everything down so much as it tries to reslice each time you change any parameters or load a part.  I would prefer to be able to set everything up, then hit slice.

alexrj:
@the-edge01, what do you mean by *"if something goes wrong"*? Can you elaborate more on your scenario?

the-edge01:
I was referring to something such as #2980 or #2958 where a forced reslice would probably have fixed the problem.

MartinMajewski:
I support that request and want to add, that it would be good for slic3r if the slicing action would be more controllable by the user, i.e. if one could deactivate the automatic invocation and have a "perform slicing" button in a convenient place.

Especially when I have a high detailed, large scale model and just want to tune a few options slic3r tries to immediately perform the slicing on every single changed option, leaving me with an unresponsive application and unnecessary computation.

gege2b:
it's (almost) already the case.

go to prefs, disable background processing, and then slicing will occurs only when "export gcode" is clicked

MartinMajewski:
I use slic3r with the direct uplink to OctoPrint and don't want to send my object every time I slice my object for review within slic3r. So no, it is not "already the case", it's a workaround and not convenient from the usability point of view.

lordofhyphens:
Wouldn't background processing already slice it and give you the update
preview? Else you could just export gcode normally--not like it HAS to go
to octoprint.

"Nothing unreal exists." - Kiri-kin-tha's First Law of Metaphysics.

MartinMajewski:
Please read my previous post. I want an option to trigger the processing on demand, because background processing is not always the optimal behavior.

kefir-:
Perhaps some hotkeys to control behaviour, like:
- Enable/disable autoslice
- Slice manually
- Save gcode
- Jump to various config/preview screens

I'm a big fan of keyboard shortcuts, and I think this could solve the issues without stuffing more stuff into the gui.

a4jp-com:
[...]

PR #3501, opened 2016-09-11
Background slice now

### LINKED ISSUE #3071: FEATURE REQUEST: Add a "Slice now" button
Using Slic3r on a daily basis, I often have to change many settings for specific print jobs, which trigger the background slicing every time. On larger pieces, it crashes the engine when changing the options and the background slicing has not completed it's run.

My request would be to have a "Slice Now" action button so you can disable background slicing and manually trigger the slicing, see the preview and make adjustments. I know you can export the GCode to trigger the slicing but exporting the GCode is unnecessary for me since I send it the the OctoPrint server (which is a very nice feature btw).

Thanks for the consideration.

lordofhyphens:
Fixes #3071


lordofhyphens:
Win32 build will appear at https://bintray.com/lordofhyphens/Slic3r/Slic3r_Branches

bubnikv:
The patch is fine, I have two comments though:
1) The new button shall be hidden if the background processing is enabled. The button has no meaning then, as the background processing triggers automatically anyway.
2) The button avoids crashes if the background processing triggers when a value is entered into the Slic3r confugration fields only partially. But the button causes Slic3r crash in many cases the same way Slic3r crashes with the background processing enabled, which defeats the purpose of the button. So the idea is sound, but we should concentrate on fixing the background processing crashes.

bubnikv:
Thanks, I extended your implementation
https://github.com/prusa3d/Slic3r/commit/34fab1566f7a99dd40d9856cb5185c8d20d1ddec
Namely, we added an image to the button and we hide the button when the background processing is enabled. Also we allow the "cancel" button on the status bar.

lordofhyphens:
I've pulled the extended implementation into this PR, so it can be merged now.

alexrj:
Merged!  (I ignored the changes to xs/src/libslic3r/PrintConfig.cpp while merging).

alexrj:
Hm. I still want to try the other way described in #2979, without a button.

PR #3502, opened 2016-09-12
Buildscripts - allow custom path for boost and detect mingw on win32

Brandano:
I have added a few lines to the build scripts to allow Win32 users to set the location of the boost libraries with a BOOST_DIR environment variable instead of forcing them to install in c:\dev. I think this is more consistent with the message generated by the build script when the libraries are not found. The second commit, still on the build scripts, is to attempt and locate Mingw by iterating through the PATH environment variable entries. MinGW should be in PATH both for Citrus and Strawberry perl, but in any case the build won't fail if it's not found, so it's more kind of a "nice to have" feature.

### DIFF
--- a/Build.PL
+++ b/Build.PL
@@ -128,8 +128,14 @@ EOF
         
         push @cmd, "$module~$version";
         if ($module eq 'XML::SAX::ExpatXS' && $^O eq 'MSWin32') {
-            my $mingw = 'C:\dev\CitrusPerl\mingw64';
-            $mingw = 'C:\dev\CitrusPerl\mingw32' if !-d $mingw;
+            my $mingw;
+            foreach my $path (split(/;/,$ENV{'PATH'})) {
+                $path =~ s/bin$//;
+                if (-e $path . 'include\expat.h') {  # don't use -x because it fails on Windows
+                    $mingw = $path;
+                    last;
+                }
+            }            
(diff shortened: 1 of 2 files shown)


alexrj:
Thank you @Brandano. I think the Boost part is superseded by the recent merge of another pull request.
Regarding the MinGW change, did you test it personally on a system using C:\dev\CitrusPerl\mingw64? In other words, can you confirm you tested personally it's backwards compatible?

Brandano:
I tested it on a Win32 machine running CitrusPerl (and on this machine I also have other precedent MinGW installs) and on a Win64 machine under StrawberryPerl. The script just scans every instance of MinGW mentioned in the %PATH% variable for the specific library header that is then needed to compile, returning the first that will match the criteria. This works because both Perl distributions will extend the environment prepending their libraries to the rest. Slic3r will compile without Mingw, though, so even if the script failed it shouldn't affect the build. It obviously only searches for MinGW installs on Windows.

lordofhyphens:
@Brandano I think this has been fixed in another way now. Could you double-check current master branch and see? I'd like to close this PR (as it's got conflicts now too).

Brandano:
Yes, this can be closed and the pull request discarded, the boost path can be specified by an environment variable and mingw doesn't seem to be used anymore. Thanks

PR #3525, opened 2016-10-04
A new feature "support_material_buildplate_only" implemented.

### LINKED ISSUE #2757: Feature request: optional printing of the "outside" support material only
When I print complex mechanical parts often need a support material under them. It prevent a bend and vibration of tall and inclined models. (I prefer a tilt angle near 45 degrees) But when I adjust appropriate "overhang threshold" for generation the support from printbed up to the model it also generated unnecessary support in internal hollow volumes of the model. It will be a good selectable option to generate support when it can reach printbed directly and disable support for other cases (internal and shielded by the model body).

### LINKED ISSUE #3310: Feature Request - Generate support only attached to bed.
Slic3r 1.2.9

please add option to generate support material only attached to bed.
Example would be pipes (or tubes) in + combination. Support should be generated only from bed to outer surface and not any support generated inside pipes.

lordofhyphens:
Also a bug has been fixed for zero interface layers. Before
slic3r would put infinite number of interface layers over top surfaces,
if the number of interface layers was set to zero.

Fixes #3310 
Fixes #2757
Thanks to @bubnikv for this work.

Win32 automatic development build: https://bintray.com/lordofhyphens/Slic3r/Slic3r_Branches/1.3.0-infill-over-plate-192#files

### DIFF
--- a/lib/Slic3r/GUI/PresetEditor.pm
+++ b/lib/Slic3r/GUI/PresetEditor.pm
@@ -428,7 +428,7 @@ sub options {
         raft_layers
         support_material_pattern support_material_spacing support_material_angle 
         support_material_interface_layers support_material_interface_spacing
-        support_material_contact_distance dont_support_bridges
+        support_material_contact_distance support_material_buildplate_only dont_support_bridges
         notes
         complete_objects extruder_clearance_radius extruder_clearance_height
         gcode_comments output_filename_format
@@ -578,6 +578,7 @@ sub build {
             $optgroup->append_single_option_line('support_material_angle');
             $optgroup->append_single_option_line('support_material_interface_layers');
             $optgroup->append_single_option_line('support_material_interface_spacing');
+            $optgroup->append_single_option_line('support_material_buildplate_only');
             $optgroup->append_single_option_line('dont_support_bridges');
         }
     }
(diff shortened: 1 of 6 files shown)


lordofhyphens:
Also, I have no idea what I was thinking when I named the branch.

VanessaE:
After some light testing with this feature (and a fair bit of normal use of Slic3r since then), this seems safe to merge.  It seems to work, and doesn't appear to break anything.

Caveat:  it can only build supports straight up from the build plate to the object - it doesn't attempt to reach the object from an angle (e.g. the way Meshmaker does.)

lordofhyphens:
Rebased to master.

lordofhyphens:
https://github.com/prusa3d/Slic3r/issues/66 denotes a reported issue with the code in this PR (or at least related).

lordofhyphens:
Rebased over master.

lordofhyphens:
This should handle a few current issues with generated supports until we can get something better implemented (with this feature still present, as it is generally useful). 

Recommend merge.

lordofhyphens:
So that's new.  
````
Global symbol "@p" requires explicit package name at /mnt/fiver/slic3r/lib/Slic3r/Print/SupportMaterial.pm line 766.
````

lordofhyphens:
There, works now.

PR #3556, opened 2016-11-02
3D Navigation improvement features

uclaros:
I have faced the following problems while navigating in the 3d view:

- I can only move the center of rotation on the X-Y plane so inspecting tall objects is strange.
- I cannot see the center of rotation.
- Sometimes when I want to rotate the view I accidentally move around objects, especially when zoomed in and there is not much free space in the view.
- While moving and rotating the view I have set the center of rotation far far away from the platter and cannot find it any more! Adding an object to the platter resets the view.

This PR adds the following functionality:

- Using the Alt key as modifier one can click and move the camera on the Z axis. This allows for rotating the view around any point in 3d space.
- Added a small indicator for the 3 axes that appears only when rotating or translating the view and is always positioned on the center of rotation. The axes are thin lines when they are hidden behind an object on the platter and thicker lines otherwise.

image

- When the shift key is pressed while the user clicks on an object, the object is not selected and the view is allowed to rotate-translate. This allows for easier rotation of the view when zoomed in or there is too little free space on the platter.
- Double clicking the middle mouse button zooms to extents of objects, or to extents of platter when there are no objects.

### DIFF
--- a/lib/Slic3r/GUI/3DScene.pm
+++ b/lib/Slic3r/GUI/3DScene.pm
@@ -148,7 +148,7 @@ sub mouse_event {
     } elsif ($e->LeftDClick) {
         $self->on_double_click->()
             if $self->on_double_click;
-    } elsif ($e->LeftDown || $e->RightDown) {
+    } elsif (($e->LeftDown || $e->RightDown) && not $e->ShiftDown) {
         # If user pressed left or right button we first check whether this happened
         # on a volume or not.
         my $volume_idx = $self->_hover_volume_idx // -1;
--- a/lib/Slic3r/GUI/3DScene.pm
+++ b/lib/Slic3r/GUI/3DScene.pm
@@ -148,6 +148,14 @@ sub mouse_event {
     } elsif ($e->LeftDClick) {
         $self->on_double_click->()
             if $self->on_double_click;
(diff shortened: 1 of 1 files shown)


lordofhyphens:
So what's the modifier if you're on OSX? :)

uclaros:
Well I guess for those macs with an alt key it will still be the alt key!
I guess it will be the option key on other macs...

Regarding the axes marker, I left the always visible marker untouched since it shows the coordinates origin on the platform and I guess it is useful to have. It's resizing algorithm though, makes no sense to me... Why do we need the axes resized if there are more objects on the table? The whole block that currently draws the static axes could be replaced with a single call to `sub draw_axes `.
Should I do it and issue a PR?

uclaros:
I just saw that PI is obviously used in other parts of slic3r too, and more accurately computed as 4*arctan(1),
so my [definition of PI](https://github.com/alexrj/Slic3r/pull/3556/commits/d7edc08287fb00cde3ca0d29d5101c950e048b20#diff-8cf21c60fb92537ecfc7bb0f527027abR49) should better change from 
`use constant PI             => 3.1415927;` 
to
`use constant PI => 4 * atan2(1, 1);`

alexrj:
Looks nice, thank you!

PR #3567, opened 2016-11-08
Add informational version macros to libslic3r.

lordofhyphens:
Added two debug and informational macros for FORK name and debug output path prefix.

Necessary for incorporating useful debugging enhancements from Prusa3d fork, and generally a good idea on their own to enhance log output (and more easily tell the difference between upstream logs and forks).

Shamelessly borrowed from work done by @bubnikv

### DIFF
--- a/xs/src/libslic3r/libslic3r.h
+++ b/xs/src/libslic3r/libslic3r.h
@@ -8,6 +8,12 @@
 
 #define SLIC3R_VERSION "1.3.0-dev"
 
+// Name of this fork. This is upstream, so we get to be just plain "Slic3r"
+#define SLIC3R_FORK_NAME "Slic3r"
+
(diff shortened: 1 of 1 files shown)


alexrj:
Hm.  I don't want to encourage forks, they should actually contribute back to the main repository. It's unusual to have a "fork name" constant, I never saw it in any project...
Also, where are those two constants supposed to be used?

lordofhyphens:
@bubnikv uses them in debugging output. He adds quite a bit of debug code
and profiling code to the work he does.

What it means for us is not including it is that anything he does will
probably take more work to use.

alexrj:
I prefer not to merge this. We'll pull any constants as soon as we need to use them... :)

lordofhyphens:
Sounds fine.

PR #3569, opened 2016-11-09
Fixed a bug with the "Window" menu items.

uclaros:
If the "Controller" tab was not visible (disabled from preferences), then the menu items and their shortcuts opened the wrong tab.
Also, the controller tab was not visible if the plater tab was not visible too.


alexrj:
Thank you!

Issue #3540, opened 2016-10-21
Ubuntu 16.10 segmentation fault when plater is displayed on 1.3.0-dev


mloebl:
### Version

_1.3.0-dev_

Git describe:
_1.2.9-367-g8910b49_
### Operating system type + version

_Ubuntu 16.10 x64_
### Behavior
- Segmentation fault launching with plater 
- Running in any mode that shows the plater
  - if I start with --no-plater it starts up ok
  - I think it may be an opengl related issue 
  -  Precompiled linux binary of 1.2.9 works fine 
  -  Worked fine on 16.04 before I upgraded to 16.10 and installed latest Nvidia driver 370.28 (done at same time)
  -  no errors during build, here's gui build output: 

App::cpanminus is up to date. (1.7042)
Class::Accessor is up to date. (0.34)
Growl::GNTP is up to date. (0.21)
LWP::UserAgent is up to date. (6.15)
Net::Bonjour is up to date. (0.96)
OpenGL is up to date. (0.70)
Socket is up to date. (2.024)
Wx is up to date. (0.9928)
Wx::GLCanvas is up to date. (0.09)
- End of output from trace:

> > (eval 254)[/usr/local/lib/x86_64-linux-gnu/perl/5.22.2/OpenGL.pm:6194]:1: sub OpenGL::GL_TRIANGLES { 4 }
> > /usr/share/perl5/Class/Accessor.pm:175:         my $self = shift;
> > /usr/share/perl5/Class/Accessor.pm:177:         if(@_) {
> > /usr/share/perl5/Class/Accessor.pm:180:             return $self->get($field);
> > /usr/share/perl5/Class/Accessor.pm:158:     my $self = shift;
> > /usr/share/perl5/Class/Accessor.pm:160:     if(@_ == 1) {
> > /usr/share/perl5/Class/Accessor.pm:161:         return $self->{$_[0]};

This is where it segfaults

mloebl:
Thanks for any help!

mac-miller:
I have exact the same Problem, but with  Ubuntu 16.04 and Nvidia driver 370.28.

mloebl:
Thanks for following up; I'd go back a version of the driver, but I believe there's a bug with the NVidia drivers on 16.10 prior to this version.   It's weird how the precompiled version works.

mac-miller:
I do not know anything about OpenGL-programming, but i did a litte search on the web and found this statement: 
[...]

PR #3575, opened 2016-11-17
Fix for OpenGL.pm 0.70 segfault

lordofhyphens:
Addresses #3540

The actual crash case appears to be the bed_triangles and bed_grid_lines draw code. I didn't dive into the exact nature of the segfault, but it can be avoided by explicitly generating buffer objects and doing the necessary binds before the draw command.

Going to do another round of code cleanup and refactoring to make things less messy, PR is intended for a squash+merge at the end. 

TODO: abstract out the draw code slightly and add branching based on whether or not we have Vertex Buffer Objects or not (which is probably the case)

### DIFF
--- a/lib/Slic3r/GUI/3DScene.pm
+++ b/lib/Slic3r/GUI/3DScene.pm
@@ -46,6 +46,10 @@ use constant GROUND_Z       => -0.02;
 use constant DEFAULT_COLOR  => [1,1,0];
 use constant SELECTED_COLOR => [0,1,0,1];
 use constant HOVER_COLOR    => [0.4,0.9,0,1];
+
+# Constant to determine if Vertex Buffer objects are used to draw
+# bed grid and the cut plane for object separation.
+# Old Perl (5.10.x) should set to 0.
 use constant HAS_VBO        => 1;
 
 
@@ -755,8 +759,6 @@ sub Render {
     # draw ground
     my $ground_z = GROUND_Z;
     if ($self->bed_triangles) {
-        my $norms = OpenGL::Array->new_list(GL_FLOAT,(0.0, 0.0, 1.0)); # common normal for both
-
         glDisable(GL_DEPTH_TEST);
         
         glEnable(GL_BLEND);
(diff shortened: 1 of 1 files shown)


lordofhyphens:
There's still some bugs in it, the cut dialog causes a segfault, probably something related to line drawing.

anoved:
Just tried this fix and can confirm slic3r now runs without segfault using OpenGL 0.70. (Although indeed the cut dialog still segfaults.) Thanks.

lordofhyphens:
Yeah, I think it is the other draw in the cut that causes a crash. Will
check the bed shape as well to make sure (and advanced property list).

lordofhyphens:
Alright, the cut plane no longer crashes. I also removed some dead code from my refactor and documented the HAS_VBO constant. 

Set it to 0 to get the old behavior. I don't think that it should be necessary on any modern system, but I figured it would be okay. Another option would be to add an advanced GUI option, but since you can't get at the GUI with the option in place... Could also add a CLI-only option instead to toggle to the old behavior, but I'm not that crazy about adding yet another CLI switch.

Any thoughts?

bubnikv:
Thanks for the effort.

In the future, my personal goal (if the time allows) is to migrate the
OpenGL code to C++, so this would allow us to use all the modern OpenGL
functionality. The Prusa3D fork has the GLEW library integrated, which
handles resolution of OpenGL functions. OSX and Linux do not need that.

Also on the same note, I am in the process of moving the AMF XML parser to
C++ as it is extremely slow now and it often causes compilation issues.

Patola:
+bubnikv Will this make using/porting to OpenGL/ES easier? I have a big interest in being able to run slic3r's interface in embedded devices and very few of them have actual desktop OpenGL working, whilst almost all have OpenGL/ES.

bubnikv:
> Will this make using/porting to OpenGL/ES easier? I have a big interest
in being able to run slic3r's interface in embedded devices and very few of
them have actual desktop OpenGL working, whilst almost all have OpenGL/ES.

I have no experience with OpenGL/ES, but AFAIK it is a stripped down OpenGL
2.0, where using the shaders is mandatory. Therefore having the code in C
seems to make the porting to OpenGL/ES easier.

alexrj:
Thank you! I confirm this fixes it for me as well. I'm not sure we need to expose a configurable VBO flag. Let's keep it this way and add it only when a test case arises.

bubnikv:
I am not sure, but I think this fix leaks VBO objects. The VBOs are getting allocated and they never get deallocated. I believe you will allocate a bit of the GPU memory at every screen refresh. I think you are missing the glDeleteBuffers call after each drawing operation.

I think this fix is simpler:
https://github.com/prusa3d/Slic3r/commit/c0b3de6248f6e6d50624b3d2132c823fb65af95e
I did not test it with Perl OpenGL 0.7 though, I only tested, that it works on my system. I think thought that it will work and that the culprit was the glVertexPointer_p call of the new OpenGL library.

bubnikv:
By the way, I find it great that nobody complained about the vertex buffer object extension :-) As it seemingly works on all systems tested, this opens a possibility for us to use it for the print path preview, where it should significantly speed up the visualization of large objects.

lordofhyphens:
I am a noob when it comes to using VBOs, I will take a look at your commit
and compare to doing it right with VBOs.

There is an issue I know about  in opengl.pm 0.7 that caused the segfault
with the legacy allocation method.

VBOs are first class with modern OpenGL.

lordofhyphens:
I will probably still produce a branch with your fix in it to see if it
solves some of the scattered issues I have seen with some distros.

bubnikv:
I am not sure what scattered issues you mean, but as far as I am know, the
alexrj master has following two issues in regard to the path preview:

1) Too large buffers are passed over to the OpenGL driver.
2) Maybe intel specific, maybe just on my computer, but I had to add
following to the end of 3DScene.pm, sub Render:
    # Calling glFinish has a performance penalty, but it seems to fix some
OpenGL driver hang-up with extremely large scenes.
    glFinish();

lordofhyphens:
I think those issues are distro specific (autoloader, etc).

lordofhyphens:
I would certainly like to modernize the GL draw code.

lordofhyphens:
It looks like the second thing you mentioned was already in the Render routine.

PR #3586, opened 2016-11-24
Quick compile fixes

lordofhyphens:
Two fixes to allow libslic3r to compile under windows.

1) GCC wants a flag to indicate that c++11 is in use, or else it complains loudly about the use of static_assert().

2) gcc 4.9.2 can't find TriangleMesh::swap with both hands and a radio telescope. Added an explicit reference so it knows where to go.


lordofhyphens:
Unnecessary now as @alexrj caught the same issue and fixed it :)

PR #3587, opened 2016-11-24
Tell GCC to use C++11 if we are using it.

lordofhyphens:
GCC defaults to not using new language features if we don't request them. This adds a compiler check with CppGuess to add the switch if the compiler is GCC.


lordofhyphens:
Right, so apparently the build server has an old gcc too. Joy. -.-

alexrj:
`-std=c++11` is not available before GCC 4.8.1. We should check for version...

lordofhyphens:
Yeah, I have been trying to figure out how to do it with the Perl modules
available to us, but haven't been able to figure it out yet  (Module::Build
is close to black magic to me).

Something else that would probably help is to check for system wide
environment variables and add those.

alexrj:
Also, compiling on MacOS X 10.8.5 (clang 4.0) doesn't work because the supplied libc++ misses many things like type_traits, std::move etc.

lordofhyphens:
This is probably something that should me mentioned in the wiki for people using a modern GCC.

PR #3590, opened 2016-11-27
Lambda modifier mesh

lordofhyphens:
Two related items. One is the ability to slide modifier meshes around in the Settings dialog, the other is the ability to make generic/anonymous (or lambda) meshes available to the user. Currently there's just a box available. Anything from Slic3r::Test is available for inclusion through appending them to the ComboBox list. 

UI needs some more work for it (porting to use the UI builder code).


lordofhyphens:
I did notice a small bug with updating of the main plater, the view isn't currently updating on the plater view (but it's correct in the Settings view and the slice is  correct).

lordofhyphens:
Revised the menu code to be nicer to look at (and largely use OptionGroup). I'm unsure how to get a combobox that isn't driven by the XS config going, couldn't find any Perl examples in the current codebase. 

@alexrj if you remember how it'd be great if you could whip up a snippet I could learn from.

alexrj:
Hey, I do really appreciate this effort - that's something needed.
But I don't think we should use Slic3r::Test at all. Those are really meant just for the test suite... What about a simple button "Add Cube Modifier" which prompts for parallelepiped sizes (can be three Wx::GetNumberFromUser dialogs)? I think that's the only useful thing for starters. What do you think?

lordofhyphens:
Works for me, I used Slic3r::Test because it was there and I didn't feel like duplicating code. I can pull that out either into its own package or otherwise. I think it could go into its own package primarily to facilitate a few other simple solids (sphere, cylinder) getting added in. 

I made a point of committing the movement code separately from the generics code if you want to grab just that via cherry-pick.

bubnikv:
One more thing. IMHO there should be a new set of transformation attributes (transformation, scale, rotation) stored for the modifier mesh, so when you open the dialog, you could reset the transformation by filling in the zeros. Your implementation just moves the coordinates of the mesh. When implementing the transformation attributes of the modifier meshes, these need to be stored into the AMF file and restored from the AMF file to be really useful.

lordofhyphens:
I'm not sure about putting it in TriangleMesh myself, although the basic code would probably be cribbable from ReadFromPerl in the xsp if you wanted to push it into XS. 

Perhaps a specialized Model instead?

lordofhyphens:
I'm not trying to implement scale or rotation. I agree that being able to store/recall the modifier mesh setup into formats that support it is useful, but the rest of it isn't quite there yet.

I'm more in favor of getting something more limited like this in while leaving it flexible enough to support more work later.

lordofhyphens:
@bubnikv Unless you were talking about putting make_cube and such into the xsp interface and not libslic3r's TriangleMesh.cpp?

bubnikv:
> Unless you were talking about putting make_cube and such into the xsp interface and not libslic3r's TriangleMesh.cpp?

That is really a matter of taste. A free function at TriangleMesh.cpp/hpp or a static function of TriangleMesh class would be possible. Then I would bind the static or free functions to the TriangleMesh XS object.

alexrj:
I agree with @bubnikv about creating TriangleMesh constructors for specific geometric shapes. Why don't you like that @lordofhyphens?

lordofhyphens:
@alexrj I mostly was trying to avoid non-specific "kitchen sink" functions, aka functions that do one thing and then sit around in the interface for all time (see: PHP). It's mostly an aesthetics thing to me, but if we made some specific constructors for it I'd be alright with it.

One thing I'd like to see in a constructor is the ability to pass in a STL vector of STL tuples for the facets and vertices as a generic, from there we can have static functions that make/return the simple shapes and then bind them to the interface via the XSP or w/e. If someone's calling it through libslic3r that isn't us then I think that would be cleaner and a bit easier to use.

lordofhyphens:
Actually, we don't need tuples, we have Point3 and Pointf3 for facet IDs and positions, currently putting together a TriangleMesh constructor that takes a std::vector of Pointf3 and Point3 for the vertices and facets and generates a mesh from that.

lordofhyphens:
While I'm in here, is there any objection to overloading [] for Point (and their respective descendent classes) to refer to x/y/z numerically? Or maybe something we can use with std::get<>?

lordofhyphens:
There's the new constructor and make_cube utility function.

lordofhyphens:
Cylinder mesh has been added (via make_cylinder in TriangleMesh.cpp and exposed as a clone function in the XSP). 

I've updated the generation UI as well; do we want to be able to generate these anonymous objects on the main plater as well? Seems like it'd be handy for a purge/prime tower (at least for making the tower).

lordofhyphens:
I'll check to see if I can work out the canvas size generated for the parts dialog, in which case slab will be fairly straightforward (take a Z from the dialog, etc).

lordofhyphens:
Behold, for it is a sphere. I defaulted the step angle to 1 degree (actually 2pi / 360).

Rotation (and preserving it) only makes a lot of sense if you expect to be able to (ab)use a file format to save the result in a way that only Slic3r knows how to replicate. I think that instead performing any translation/rotation directly and exporting is a better idea as it lets the user pull the mesh out with the object and manipulate it in an exterior program without breaking something.

I know that @bubnikv (or someone else he works with) is working on extending the AMF reader and making it suitable for a project file through special comments.

lordofhyphens:
Slab is done as a UI function, the same elements get the bounding box for model_object and call TriangleMesh::Cube() with those attributes and then shifts it so that it is centered properly.

lordofhyphens:
I wasn't calling repair() on the produced mesh, nor was I marking it as fixed when I generated it, hence the call to repair().

lordofhyphens:
I feel this is feature complete at this point, if someone else wants to do a PR later to add more functionality fine.

alexrj:
Yay! I'm going to merge this.

lordofhyphens:
If you feel like messing with it as you do so, feel free :)

alexrj:
I'm doing some minor adjustments, but this new feature is really great to have :)

lordofhyphens:
Go nuts. ;) I know that @bubnikv had rotation on his wishlist as well as
full absolute coordinates with save/load.

PR #3591, opened 2016-11-27
Avoid recentering viewport if the bounding box is max size 0

### LINKED ISSUE #3589: Division by zero when chaning to spiralize mode
### Version
current master head:  dfd2466a724acf12a0e48d0645f381b01d45680a

Ubuntu

### Behavior
Illegal division by zero at .Slic3r/lib/Slic3r/GUI/3DScene.pm line 373.

* _Steps needed to reproduce the problem_
Enable spiralize mode
* _Expected Results_
no crash
* _Actual Results_
crash

lordofhyphens:
Defensive coding - fix provided by @jreinam

Fixes #3589

### DIFF
--- a/lib/Slic3r/GUI/3DScene.pm
+++ b/lib/Slic3r/GUI/3DScene.pm
@@ -370,12 +370,15 @@ sub zoom_to_bounding_box {
     # bounding box
     my $max_size = max(@{$bb->size}) * 2;
     my $min_viewport_size = min($self->GetSizeWH);
-    $self->_zoom($min_viewport_size / $max_size);
+    if ($max_size != 0) {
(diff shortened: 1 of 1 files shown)


lordofhyphens:
Works for me, I'll close the related issue then.

lordofhyphens:
Perhaps we should merge this in to cover the situation? We could be chasing down marginal cases for weeks.

lordofhyphens:
Apparently it still is in a few cases.

lordofhyphens:
I think this is necessary for defensive coding (avoid crashes). Alternative I think is to invent a flag to not trigger the draw/update until slicing is complete, but if your BB is zero/invalid it amounts to the same thing, doesn't it?

PR #3610, opened 2016-12-12
New Rectilinear implementation

alexrj:
@bubnikv did a great work with his Rectilinear2 implementation, so I strongly considered porting that work upstream. While I couldn't find documentation or test cases for that code, he told that the key points are:
* faster processing compared to the current code;
* connections between lines follow the polygon perimeters instead of being straight lines;
* paths are more continuous because of this reason (they are not interrupted when contours are concave) and also because the sweepline goes only in one direction.

However I was worried by code readability and maintainability, so I wrote a more compact implementation which has these advantages:
* more compact (~470 lines -100 being comments- vs. ~1,700)
* probably faster - I didn't compare but it has O(n) complexity and never calls Clipper
* numerically robust (no fudge factors, epsilons or safety offsets)

Unless I missed anything, I ported @bubnikv's functionality.

I'm parking this change in a pull request in order to collect comments from @bubnikv, @lordofhyphens and anyone else.


alexrj:
The downside of this is that the Line infill pattern gets removed. I think nobody will miss it. Also, I'm proposing to use the new "Stars" pattern as the default one instead of Honeycomb.

xs/src/libslic3r/Fill/Fill.cpp
```diff
@@ -80,26 +76,27 @@ Fill::adjust_solid_spacing(const coord_t width, const coord_t distance)
...
+    
+    coord_t distance_new = (width / number_of_intervals);
     
     const coordf_t factor = coordf_t(distance_new) / coordf_t(distance);
     assert(factor > 1. - 1e-5);
     
     // How much could the extrusion width be increased? By 20%.
     const coordf_t factor_max = 1.2;
-    if (factor > factor_max)
+    if (false && factor > factor_max)
```
lordofhyphens: This means the following line is never called, why is it here?

xs/src/libslic3r/Point.cpp
```diff
@@ -18,6 +18,12 @@ Point::operator==(const Point& rhs) const
     return this->coincides_with(rhs);
 }
 
+bool
+Point::operator<(const Point& rhs) const
+{
+    return this->x < rhs.x || this->y < rhs.y;
```
lordofhyphens: Is there an external constraint/property in mind when comparing Points in this manner? Sorting with this seems particularly arbitrary.

My first thought on "is Point X < Point Y" was to compare the length of their distance vectors to some third point (such as the origin).

lordofhyphens:
Looks like the test needs to be updated:

xs/t/15_config.t need to read "stars" instead of "line" in the relevant areas.

xs/src/libslic3r/Fill/Fill.cpp
alexrj: Ah right, @bubnikv added that and I disabled it temporarily while debugging. Will remove that `false`.

xs/src/libslic3r/Point.cpp
alexrj: This is needed by std::map. It's for ordering, not for comparing size. I'll move this to a more narrow scope.

bubnikv:
+Point::operator<(const Point& rhs) const
+{
+    return this->x < rhs.x || this->y < rhs.y;


This is not what one intuitively expects from the < operator. It
usually is a lexicoraphical sorting,

return this->x < rhs.x || (this->x == rhs.x && this->y < rhs.y);

xs/src/libslic3r/Point.cpp
alexrj: Even better, actually, It was not needed. I just left it when deciding to go for another solution. :)

alexrj:
I pushed a fix for those three minor things.

lordofhyphens:
My sense of pessimism says that as soon as this goes it we get three issues asking where the Line infill went. :(

alexrj:
Definitely... but I think it makes no sense anymore. What do you think?

lordofhyphens:
I never quite understood what the Lines infill was for in the first place, so I'm quite alright with axing it.

RafaelEstevamReis:
@lordofhyphens I think (about your pessimism) a GUI note should be fine and maybe a special catch at command line to say "hey, lines do not exist anymore" avoid the issues

lordofhyphens:
I wouldn't waste the GUI screen space on it tbh. Put it in the release notes and deal with it if someone complains.

lordofhyphens:
UI bug that may be related; the "Hey should I switch to rectilinear for 100% infill" box goes to Stars instead.

alexrj:
Crossing fingers, let's go.

alexrj:
@lordofhyphens, good spot, I'll fix that right after the merge.

PR #3614, opened 2016-12-13
Extra perimeters detection includes extra perimeters.

### LINKED ISSUE #3613: Slic3r hangs with "Processing triangulated mesh..."
### Version
1.2.9

Kubuntu 16.04.1 LTS

### Behavior
* add 'g-doesntwork.stl' to the plater
* expected: slic3er slices the STL
* observed: slic3er stays in "Processing triangualted mesh..."

This happens with 'g-doesntwork.stl', but not with 'g.stl'. The difference of the two files is the depth of the emboss of the "G": in 'g.stl' it is larger than on 'g-doesntwork.stl'.

As a side mark: the first layer above the G uses the longer axis. Is there a way to turn it 90°, so that it has shorter bridges?

lordofhyphens:
Fixes #3613 

Would appreciate it if @bubnikv or @alexrj would check the one-line edit to see if I'm missing something critical about the extra perimeter detection.

### DIFF
--- a/xs/src/libslic3r/PrintObject.cpp
+++ b/xs/src/libslic3r/PrintObject.cpp
@@ -662,7 +662,7 @@ PrintObject::_make_perimeters()
         this->typed_slices = false;
         this->state.invalidate(posPrepareInfill);
     }
-    
+    std::cerr << "Done merging slices" << std::endl;
     // compare each layer to the one below, and mark those slices needing
     // one additional inner perimeter, like the top of domed objects-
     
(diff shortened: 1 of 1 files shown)


xs/src/libslic3r/PrintObject.cpp
```diff
@@ -662,7 +662,7 @@ PrintObject::_make_perimeters()
         this->typed_slices = false;
         this->state.invalidate(posPrepareInfill);
     }
-    
+    std::cerr << "Done merging slices" << std::endl;
```
alexrj: whoops ;)

xs/src/libslic3r/PrintObject.cpp
```diff
@@ -695,13 +695,12 @@ PrintObject::_make_perimeters()
...
             const coord_t ext_perimeter_width   = ext_perimeter_flow.scaled_width();
             const coord_t ext_perimeter_spacing = ext_perimeter_flow.scaled_spacing();
-            
             for (Surfaces::iterator slice = layerm.slices.surfaces.begin();
                 slice != layerm.slices.surfaces.end(); ++slice) {
                 while (true) {
                     // compute the total thickness of perimeters
                     const coord_t perimeters_thickness = ext_perimeter_width/2 + ext_perimeter_spacing/2
-                        + (region.config.perimeters-1 + region.config.extra_perimeters) * perimeter_spacing;
+                        + (region.config.perimeters-1 + region.config.extra_perimeters + slice->extra_perimeters) * perimeter_spacing;
```
alexrj: Oh, that's right!
And I clearly see my error at porting time: `region.config.extra_perimeters` is a boolean: it should not be summed in that expression!  I wrote `region.config.extra_perimeters` instead of `slice->extra_perimeters`.

lordofhyphens:
Oops indeed. I am away from my computers until probably Sunday or Monday;
you should have access though to change the commit yourself, or fix it
separately and close this :).

alexrj:
I fixed this directly, quicker... But gave you credits in the message ;)

PR #3618, opened 2016-12-15
Retract when there is no a lower layer, like in the first layer.

### LINKED ISSUE #3053: Feature request: retract on first layer
Using Slic3r 1.2.9: In order to reduce the number of retracts, I enabled the option "Only retract when crossing perimeters". However, I'd like to see an option to disable this function for the first layer, so that no ooze is visible in the bottom of a printed object. Would this be possible? If required, I can try to make a photo of the issue.

xoan:
Not sure if this small addition is all the code needed to fix #3053, but it seems to work.


alexrj:
I pushed a different change (more correct). But thank you for the input!

PR #3619, opened 2016-12-15
Use --local-lib when calling cpanm

alexrj:
I propose to supply `--local-lib local-lib` to cpanm so that every module is installed in the local-lib directory. Advantages of this:

* no more need to tell users to install and initialize local::lib themselves
* no more need for `sudo`
* easier troubleshooting for users here, no more permission issues
* ability to keep several Slic3r versions or branches (instead of recompiling every time I switch between branches)

@lordofhyphens, what do you think?
I'm sure @henrikbrixandersen would appreciate this :-)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ MANIFEST.bak
 xs/MANIFEST.bak
 xs/assertlib*
 .init_bundle.ini
+local-lib
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,3 +20,4 @@ addons:
     - libboost-thread1.55-dev
     - libboost-system1.55-dev
     - libboost-filesystem1.55-dev
+    - liblocal-lib-perl
--- a/Build.PL
+++ b/Build.PL
@@ -107,7 +107,16 @@ EOF
         if !$cpanm;
     my @cpanm_args = ();
     push @cpanm_args, "--sudo" if $sudo;
-
+    
+    # install local::lib without --local-lib otherwise it's not usable afterwards
+    if (!eval "use local::lib; 1") {
+        my $res = system $cpanm, @cpanm_args, 'local::lib';
+        warn "Warning: local::lib is required. You might need to run the `cpanm --sudo local::lib` command in order to install it.\n"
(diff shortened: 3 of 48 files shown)


alexrj:
Heh, of course this commit is incomplete.

lordofhyphens:
Works for me, probably something we should have done a long time ago :). 
This will probably impact (of course!) distro packagers - @hyperair. Just making sure someone sees it coming.

lordofhyphens:
So are we putting the local::lib path to a subdir of the slic3r directory then? The other thing I can think of is how this impacts Windows users.

Side-by-side branches would be checked out in multiple directories then?

alexrj:
Yeah, nothing will be installed outside of the Slic3r directory.  So you can have multiple copies of the repository, each one with its compiled and working stuff.

I think nothing special happens on Windows, it just works... :)

lordofhyphens:
Looks like we need to have local::lib in the install for the travis server, all of the tests are failing.

alexrj:
You are right.  Let's see with this commit...

lordofhyphens:
Looks like FindBin is giving the test dir when running tests, Not really sure why, although I guess it has something to do with prove (which would be run from that dir)?

alexrj:
You're right. Stupid me, I didn't test completely...

lordofhyphens:
These things happen :). One of the reasons why I prefer to park most anything in a PR before merging. Lets the build server catch egregious issues so I don't have to remember to.

alexrj:
Okay, fixed.

Patola:
Thank you for this feature. I had so much trouble with libraries that this option will effectively resolve.

hyperair:
@lordofhyphens Thanks for the ping. I'll just patch this out in the Debian package if need be.

swiftgeek:
find . -iregex '.*\.\(pl\|pm\|t\)' -print0 |  xargs -0 -l sed -i -e '/use local::lib/d'
Seems to be enough to make it possible to package again

alexrj:
(Sorry for breaking packages)

PR #3627, opened 2016-12-19
Limit bridge over infill

alexrj:
I always prefer to avoid adding new options, so I'd rather try to find a decent heuristic for avoiding bridge over infill when not necessary.

Overstuffing was mostly handled by 38a9e32a287c4a601e256ebd96533f68c0c8dcc0 and 4724f0fb99abee2fa1b3252c4137e738f760e7d6.
The main issue now is to avoid bridging in those cases when it's physically possible but not necessary.

I'm proposing a new logic which compares infill density to a density threshold, above which we don't bridge. We calculate this density threshold by comparing the bridge flow to the normal solid infill flow we would use otherwise. If those flows are almost equal, we set the threshold to 25% (i.e. we only bridge when printing at low densities). If normal flow would be much lower, thus very poor (this happens when printing thin layers), we raise this density up to 50% in order to bridge more often. Intermediate cases are handled by interpolation.

For example, for a 0.4mm nozzle and default extrusion widths, no bridging will take place in these cases:

* when density > 50%
* when layer height = 0.1 and density > 49%
* ...
* when layer height = 0.15 and density > 35%
* ...
* when layer height >= 0.2 and density > 25%

Any thoughts?

### DIFF
--- a/xs/src/libslic3r/PrintObject.cpp
+++ b/xs/src/libslic3r/PrintObject.cpp
@@ -2,6 +2,7 @@
 #include "BoundingBox.hpp"
 #include "ClipperUtils.hpp"
 #include "Geometry.hpp"
+#include <algorithm>
 
 namespace Slic3r {
 
@@ -542,7 +543,8 @@ PrintObject::bridge_over_infill()
         );
         
         // get the average extrusion volume per surface unit
-        const double mm3_per_mm2 = bridge_flow.mm3_per_mm() / bridge_flow.width;
+        const double mm3_per_mm  = bridge_flow.mm3_per_mm();
+        const double mm3_per_mm2 = mm3_per_mm / bridge_flow.width;
         
         FOREACH_LAYER(this, layer_it) {
             // skip first layer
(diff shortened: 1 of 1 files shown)


lordofhyphens:
If I recall, didn't we pull in the debug output code from the Prusa fork
that @bubnik did? Might be useful to show before/after pics with that.

lordofhyphens:
As a heuristic that seems fine to me, I guess the "real" proof will be print tests with some of the known problem cases.

alexrj:
Well, you can enable my SectionCut code in Print.pm, but there's not much to see here. This pull request is about reducing the cases when bridging is applied. Nothing changes in the bridging itself, when applied...

lordofhyphens:
https://github.com/prusa3d/Slic3r/issues/66 looks like it's going to have an issue with this.

alexrj:
@lordofhyphens I don't understand your comment, how's that related to this change?

lordofhyphens:
@alexrj I linked the wrong PR. I meant for it to go in the support touching build plate only PR.

PR #3646, opened 2017-01-01
Cleanup of Win32 packaging build script

lordofhyphens:
With local::lib, most of the used dependencies are now in local::lib and thus don't need to be called out separately. Removed extra modules called out.


lordofhyphens:
Merging this in with a squash to cleanly merge it. Tested with my own system and the basic install was tested.

This should be a bit more future-proof now.

Issue #3655, opened 2017-01-05
Choose "no" on "auto-start" print for Octoprint always start the print


gege2b:
hi all and happy new year :)

### Version
_Version of Slic3r used goes here_
Current dev ( 586d95b28a8e473bdcbd959f1e37d215ae531df5 )

### Operating system type + version

Kubuntu 16.60 x64

### Behavior
* _Describe the problem_
Sending Gcode to Octoprint, when prompted if the print should start as soon as the file is uploaded, I choose "no" but the print start anyway. It also start if I choose yes (as expected).

So basically, the print always start, no matter what I choose when prompted

* _Steps needed to reproduce the problem_

Load a model, slice it, send to printer, click 'no' for auto-start the print, look the print begin

Another question related to this new octoprint stuff : is there a way to disable this (prompting for the filename AND for autostart) ?

mloebl:
Same problem here too on Linux.  

I also wouldn't mind an option to disable that prompt to ask about auto starting.

gege2b:
OK at least, I'm not the only one who got this issue :)
For the option, I suggested something in #3657

mloebl:
@gege2b Nope :)  Meant to log this last week when I hit the issue, but forgot... remembered today, and saw you beat me to it.

gege2b:
aha :)

after digging a bit into the code, I think I got the culprit
Plater.pm line 1402 : 
 ```            print => $self->{send_gcode_file} ? 1 : 0,
should be : 
```            print => $self->{send_gcode_file_print} ? 1 : 0,

Still not tested, my octoprint server is not reachable right now

gege2b:
OK I can confirm (tested right now)
@alexrj or @lordofhyphens do you want me to do a PR for this tiny fix ?

lordofhyphens:
Go ahead.

Issue #3657, opened 2017-01-06
Feature request : slight UI change for octoprint stuff


gege2b:
Hi all

Due to #3655 , I came with an idea to keep the new octoprint upload things (change the name, run the print after upload...) but still keep the old behaviour

I would suggest something like this screenshot, this will give the user the choice, and - IMHO - would not clutter too much the UI.

![image](https://cloud.githubusercontent.com/assets/2814966/21719897/6203b760-d420-11e6-9fbb-e78ca2cd426d.png)
(edit the wrong screenshot)

It's also possible to replace the "send to printer as..." button with an "Edit filename before upload" checkbox

Also, giving octoprint stuff its own place could be usefull in the future, if some other related features are added (i.e. basic control of the printer, or feedback of the current print, or anything else allowed by the octoprint API) - but that's not  the subject here

I know @alexrj is kind of reticent to add more options to the UI, but what do you think about that ? :)

cheers

gege2b:
Hi there
I'm just doing some UI test here : 

https://github.com/gege2b/Slic3r/tree/sendToPrinterUI

Right now it's not working (I still have to figure out how to take these checkboxes into account when clicking the send gcode button)
But it gave an idea of how it could look

I'm a real noob at perl programming, so please be indulgent :p

dwillmore:
I would welcome some setting somewhere that allowed the 'send to printer' to be customized to the extent it could optionally prompt for a new filename and/or start printing once uploaded.

Specifically, I would like the ability to *not* get a dialog for either of those features and just upload the file to the printer.

So, I see it as two check options.
1) Do/don't prompt for filename
2) Radio button:
 a) Do prompt for 'start printing after upload'
 b) Do not prompt for 'start printing after upload' and don't start printing anyway
 c) Do not prompt for 'start printing after upload' and start printing anyway.

I think that covers the normal cases.

gege2b:
Hi  @dwillmore 

that's exactly what the mod I made on my "sentToPrinterUI" branch does, but with only one checkbox ("print [or don't] after upload") and two buttons ("send" and "send as..."). 

Doing this, there isn't any popup (except if you click on "send as" OR if a file with the same name already exists on Octoprint), it also covers the normal cases

I think there is less overhead in the UI this way (one checkbox and two buttons instead of one button and 4 check/radio), but its something pretty subjective I admitt

still waiting to know the dev's opinion on this feature request/piece of crappy code :)

dwillmore:
[...]

PR #3659, opened 2017-01-09
Fix for #3655 - always print after gcode send

gege2b:


lib/Slic3r/GUI/Plater.pm
```diff
@@ -1399,7 +1399,7 @@ sub send_gcode {
             # OctoPrint doesn't like Windows paths so we use basename()
             # Also, since we need to read from filesystem we process it through encode_path()
             file => [ $path, basename($path) ],
-            print => $self->{send_gcode_file} ? 1 : 0,
+            print => $self->{send_gcode_file_print} ? 1 : 0,
```
lordofhyphens: Basically a typo?

gege2b: basically, yeah :)

alexrj:
Great job! :-)))))))))

lib/Slic3r/GUI/Plater.pm
lordofhyphens: Yeah, I said you should put it in so your name is in the commit log for it ;) 

I would have merged it in myself if I'd had time to test it locally.

dwillmore:
Does this make the "send to printer" button always send & print?  Because that would break my workflow.

gege2b:
no, it just make the confirmation box act as intended

I feature request in #3657 to make the "send & print" feature really optionnal (just a checkbox and no confirmation dialog)
Still waiting for comments

Issue #1766, opened 2014-02-12
Estimate build time with material needed


RenaeMacLeod:
It would be very helpful to have Slic3r calculate an estimated build time, along with the amount of material needed, and save it to the GCode file above or below the 'filament used' comment. It would be a bonus to see this at the bottom of the program beside the 'GCode exported' message.

lordofhyphens:
@RenaeMacLeod Yes, but frankly it's going to be inaccurate anyway and most gcode host programs (like pronterface) already do this.

bubnikv:
Cura contains a simplified Planner code from the Marlin firmware, albeit with the hard coded acceleration values :-) Still the time estimates given by Cura are quite precise.
One can actually build a trimmed down Marlin firmware into the Slic3r just for the time estimates. In my opinion it makes sense.

lordofhyphens:
@bubnikv I would rather not put any code from Marlin into Slic3r myself. ;)

thiagopeixoto16:
Hi @lordofhyphens , I'm just to check what I'm doing wrong since my UI is not showing the estimated time after gcode export.
► Slic3r 1.3.1 dev 2019.02.10.2284.05a2f902.64bit
![image](https://user-images.githubusercontent.com/4227271/53736456-8cb98480-3e68-11e9-821b-edb49419147a.png)

thiagopeixoto16:
Updating: I tried different things on dev version and noticed that it is only being displayed at **Controller** tab when printing via USB (queue frame).
Can this be moved to the Print Summary screen for those don't print via USB?

Another suggestion:
Repetier Host has a field that allows calibration of this estimated time, it is a percentage field that allows the user to manually fix the real machine print time deviation, screenshot below:
![image](https://user-images.githubusercontent.com/4227271/54034453-a5da7200-4195-11e9-9527-d99326ecacc3.png)

lordofhyphens:
@thiagopeixoto16 move it and open a pull request :)

Should be easy to find the code that does the display.

One feature at a time though; I have been known to implement the top line feature request and ignore additional asks (closing the issue).

thiagopeixoto16:
Thanks for you reply!

@lordofhyphens You're completely correct, in this way we can have different issues for different requests.

About your "DIY" recommendation, I'll ask someone who know coding near me, as I told you before I'm not that guy and certainly will create more problems than solutions trying to change something :)

saintofinternet:
[...]

PR #3669, opened 2017-01-16
Add weight/cost output to gcode. On the way to #647

### LINKED ISSUE #647: Feature request;  GUI presentation of verbose report on processed plate.
As feature request [644#](https://github.com/alexrj/Slic3r/issues/644#) was well received; a related GUI feature request.

It would be really nice if the calculated cost, length and weight of the filament required and maybe estimated print-time for the 'plate' one has processed, would maybe(?) appear beneath the object placement window.

Thanks again for sharing your awesomeness with the rest of mankind!

### LINKED ISSUE #2250: [feature request] info about material used upfront
It would be nice if 'filament used' info would be displayed somehwere after 'export gcode'.

data is there, after one opens up file, total filament lenght and volume is displayed at the very bottom.
it is just very inconvinient to open up large gcode file in text editor and scroll to the bottom (btw. noticed openscad does this job faster than leafpad and nano )

lordofhyphens:
Parking PR here to get feedback, this has been asked for a couple times.

Fixes #647
Fixes #2250 

#1766

### DIFF
--- a/lib/Slic3r/GUI/Plater.pm
+++ b/lib/Slic3r/GUI/Plater.pm
@@ -444,8 +444,8 @@ sub new {
             $grid_sizer->AddGrowableCol(3, 1);
             $print_info_sizer->Add($grid_sizer, 0, wxEXPAND);
             my @info = (
-                fil_mm => "Used Filament (mm)",
-                fil_mm3 => "Used Filament (mm^3)",
+                fil_cm  => "Used Filament (cm)",
+                fil_cm3 => "Used Filament (cm^3)",
                 fil_g   => "Used Filament (g)",
                 cost    => "Cost",
             );
@@ -1127,6 +1127,7 @@ sub async_apply_config {
         # remove the sliced statistics box because something changed.
         if ($self->{"right_sizer"}) { 
             $self->{"right_sizer"}->Hide($self->{"sliced_info_box"});
+            $self->{"right_sizer"}->Layout;
         }
     } else {
         $self->resume_background_process;
@@ -1302,6 +1303,7 @@ sub export_gcode {
     # this updates buttons status
     $self->object_list_changed;
     $self->{"right_sizer"}->Show($self->{"sliced_info_box"});
+    $self->{"right_sizer"}->Layout;
     
     return $self->{export_gcode_output_file};
 }
@@ -1398,8 +1400,8 @@ sub on_export_completed {
     $self->{send_gcode_file} = undef;
     $self->{"print_info_cost"}->SetLabel(sprintf("%.2f" , $self->{print}->total_cost));
     $self->{"print_info_fil_g"}->SetLabel(sprintf("%.2f" , $self->{print}->total_weight));
-    $self->{"print_info_fil_mm3"}->SetLabel(sprintf("%.2f" , $self->{print}->total_extruded_volume));
-    $self->{"print_info_fil_mm"}->SetLabel(sprintf("%.2f" , $self->{print}->total_used_filament));
+    $self->{"print_info_fil_cm3"}->SetLabel(sprintf("%.2f" , $self->{print}->total_extruded_volume) / 1000);
+    $self->{"print_info_fil_cm"}->SetLabel(sprintf("%.2f" , $self->{print}->total_used_filament) / 10);
     
     # this updates buttons status
     $self->object_list_changed;
@@ -1598,6 +1600,7 @@ sub update {
     }
     if ($self->{"right_sizer"}) { 
         $self->{"right_sizer"}->Hide($self->{"sliced_info_box"});
+        $self->{"right_sizer"}->Layout;
     }
     $self->refresh_canvases;
 }
@@ -1674,6 +1677,7 @@ sub on_config_change {
     }
     if ($self->{"right_sizer"}) { 
         $self->{"right_sizer"}->Hide($self->{"sliced_info_box"});
+        $self->{"right_sizer"}->Layout;
     }
     
     return if !$self->GetFrame->is_loaded;
--- a/lib/Slic3r/GUI/Plater.pm
+++ b/lib/Slic3r/GUI/Plater.pm
@@ -444,6 +444,7 @@ sub new {
             $grid_sizer->AddGrowableCol(3, 1);
             $print_info_sizer->Add($grid_sizer, 0, wxEXPAND);
             my @info = (
+                fil_mm => "Used Filament (mm)",
                 fil_mm3 => "Used Filament (mm^3)",
                 fil_g   => "Used Filament (g)",
                 cost    => "Cost",
@@ -1123,6 +1124,10 @@ sub async_apply_config {
     if ($invalidated) {
         # kill current thread if any
         $self->stop_background_process;
+        # remove the sliced statistics box because something changed.
+        if ($self->{"right_sizer"}) { 
+            $self->{"right_sizer"}->Hide($self->{"sliced_info_box"});
+        }
     } else {
         $self->resume_background_process;
     }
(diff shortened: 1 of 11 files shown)


lordofhyphens:
The parameters default to 0, which turns off the extra output. Cost is defined as per-kg and as such you need the material density to make it work.

Patola:
Congratulations, that is a thing that everybody I know eagerly awaits for in Slic3r. Little question: from the code it seems that the price is set by extruder, wouldn't it be best if it was set by filament?

lordofhyphens:
It is, but you assign filament to extruders. The code is grabbing all of
the filament used by each extruder.

I need a better spot to put those statistics. Right now they are being
appended to the status bar message when you do an export. I think that area
could do with a refactor anyway, as it takes a full gcode export presently
to populate the information. What I think people want is for that
information to be present in the preview tab only.

Patola:
Status bar message? Why not the "Info" area? Or is that the area you are referring to?
image

lordofhyphens:
The info area is refreshed per object, not for the entire thing.

It would take a lot more work to count used filament per object.

RafaelEstevamReis:
And about rename that one to "object info" to make it more specific and a
new one below that named  "job info" with total filament and cost info?

lordofhyphens:
@RafaelEstevamReis Fair point there; I really should ask a friend of mine who is good at UX to take a once-over of Slic3r's UI.

RafaelEstevamReis:
I tried to make some UI changes but I have never seen PERL neither wx.
You helped me a lot by IRC (display total filament), but my free time
vanishes and I have to stop.

lordofhyphens:
There's a box now with the sliced statistics, generated post slice. I think I have the math wrong for cost, but I'm kinda spacing out on the unit conversions.

lordofhyphens:
Build should show up at https://bintray.com/lordofhyphens/Slic3r/Slic3r_Branches

lordofhyphens:
Win32 build: https://bintray.com/lordofhyphens/Slic3r/download_file?file_path=slic3r-slice_statistics.2017.01.21.337.716806e.zip

I don't know what's up with Travis. I think I have the weight calculated properly from the volume.

lordofhyphens:
I've done some rudimentary testing, this looks feature-complete. I'd appreciate other people giving it a whirl.

Patola:
I tried it, configured one filament accordingly, imported one object, no deal. Two objects, no deal. It does not seem to be popularing the "Sliced Info" frame.
slic3r_521
selection_520

Patola:
Got it to populate the frame after saving the G-Code. I understand the reason why this is this way, but it might be confusing to users. 
slic3r_522

Patola:
Just to report some additional tests with material with cost defined and cost not defined: the algorithm worked as intended, in this case I set the bigger object to be "ABS_rosa_175mm" which has no cost or density defined. The Slic3r Info frame considered only the filament with density and cost to perform the calculation.

slic3r_524

forReason:
Really Nice!
How about some advanced Calculations like Energycost (time to print the thing) and maybe for more professional users a general Maintenance/printer startup cost?

Patola:
I just noticed that the calculations seem to be three orders of magnitude off. I used 1.24 g/mm³ for the filament (which is the actual density of PLA) but it said in the Sliced Info that the tiny smiling bishop would weight 4,5 kg.

lordofhyphens:
Are you sure you didn't derp and use the value in units of g/cm^3 ? I know
I did that at least once myself.

1.24g/cm^3 = 1.24x10^-3 g/mm^3.

Patola:
my apologies. PLA's density is 1.24 g/cm³, not mm³. That means it is 0.00124 g/mm³. In slicers that have a field for cost I am used to input it in g/cm³, so I entered that into slic3r without even thinking about units. Now the calculation is working

lordofhyphens:
@forReason I don't see any value added in putting static costs into the interface, especially since it has very little practical use for most people and is yet another option to clutter the interface  (and we are trying to limit those).

(Aka you need to convince me)

@Patola regarding the frame not populating until after gcode export, that is something I might add a tooltip to the box to cover. While I would love to have that information available on preview, I think it would require a lot of refactoring to get it.

So what to do about user error in the case of mixed settings? It should be treating the weight for the part without settings as 0.

lordofhyphens:
Yeah, I have the unit there for that reason ;)

I have considered entering it as g/cm^3. Currently weighing the pros and
cons.

I probably will change it to avoid that particular bit of confusion, but
there is no conversion necessary if you are directly calculating from
weight/volume as the tooltip indicates.

lordofhyphens:
I also don't much care for the (money) placeholder. I think that it's okay for now, but should be replaced with the correct locale symbol when we have that working correctly.

forReason:
@lordofhyphens You are probably right the static value is unnessesary. BUT I really would consider energy costs. by Print-time

Patola:
@lordofhyphens not refactoring is better then. There's things that are much more important right now, I don't think it is a great hassle to wait for the G-Code export to have this data.

lordofhyphens:
If I add a print time estimate, three hours later someone will complain
about it being inaccurate, no matter how much documentation I throw in for
it being wildly optimistic. And it will be. I am not going to clutter the
interface with all of the minutae required to even approximate printing
times.

That is the source of me not wanting to deal with print times: it never
works anyway and it is really complicated. Very much not worth it to me.

lordofhyphens:
I've taken a cue from @bubnikv and the box now hides itself if you don't have gcode exported. It goes away if you move the model in such away that invalidates things.

lordofhyphens:
@alexrj I would like to merge this unless you have an objection.

lordofhyphens:
https://ci.appveyor.com/api/buildjobs/wypaif3a28jokd8t/artifacts/slic3r-master.2017.03.07.457.82581b3.zip
artifact path for the windows build btw.

platsch:
Tested and works so far. Two remarks / suggestions:
is there a reason not to set useful default values? It feels like a rough estimation is better than no estimation...
I'd like to have the length of filament used, didn't check your code, isn't that already calculated to get the volume? if not, it should be easy to do since the filament diameter is known already...?

Downside: even more clutter

lordofhyphens:
Added the filament length (which was already exposed) and ensured that the box is hidden on config change or if invalidation happens for background processing.

lordofhyphens:
I feel it is better to return something that is obviously wrong so the user can set the appropriate values for their filament (since density of PLA and ABS differ significantly) than provide some "default" which is generally always wrong (and basically silently wrong).

platsch:
Ok, lenght works now, empty defaults is fine. But I noticed 3 other issues:

- the units are inconsistent. The Density (input) is g/cm³ while the output volume is mm³
- The UI doesn't update after exporting i.e. the Sliced Info frame is invisible. Only when I click somewhere it appears. No idea what's going on.
- If the Sliced Info was displayed correctly and I change a setting to trigger re-slicing I end up with this:
filament_cost_screenshot

lordofhyphens:
Odd, on my system it re-updated as it should. Might need to call something
else explicitly.

The density unit was originally in g/mm^3 and was changed to let people
plug in values from Wikipedia (which I think is a good thing). I don't see
how this "inconsistency" is a problem, though. All of the outputs are
consistent with each other. The volume isn't *that* critical of a item
(leaving it out of the display wouldn't hurt anything).

lordofhyphens:
The issue with the box doing weird things should be gone now, and I've changed the display units to cm and cm^3. If someone else wants it to be mm/mm^3, too bad for them. ;p

lordofhyphens:
image

Here it is after I changed a configuration option: 
image

lordofhyphens:
Rebuild this PR now

xoan:
Nice feature!

Some things about it:

* Floats should be probably reduced to 2
* Units should be after value (Used filament: X.YZ cm)

PR #3670, opened 2017-01-16
Disable background processing by default.

lordofhyphens:
It causes more problems than it solves, especially with the new slice-on-preview code available.

Most of the weird crash bugs in Slic3r can usually be traced to background processing and strange timing issues that are very very difficult to reproduce.


lordofhyphens:
:D

PR #3671, opened 2017-01-16
Remove main opt

lordofhyphens:
Been seeing an undefined "entry" error 

`Use of uninitialized value in subroutine entry at /mnt/fiver/slic3r/lib/Slic3r/GUI/Plater.pm line 1228.`

Tracked it to the use of $main::opt{output}. Removed it and didn't see any changed behavior.

### DIFF
--- a/lib/Slic3r/GUI/Plater.pm
+++ b/lib/Slic3r/GUI/Plater.pm
@@ -1488,7 +1488,7 @@ sub _get_export_file {
     
     my $suffix = $format eq 'STL' ? '.stl' : '.amf.xml';
     
-    my $output_file = $main::opt{output};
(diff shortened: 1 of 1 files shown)


lordofhyphens:
I think main::opt is actually supposed to reference "remember output directory" setting.

alexrj:
`$main::opt{output}` holds the contents of the CLI `--output` argument, which is used as default value for the export dialog. It's a pretty hidden feature, but I think it's there for a good reason.

Better to add `|| ''` after it in order to turn the undef value to an empty string and suppress the warning.

lordofhyphens:
Looks like this was fixed by @alexrj. Closing.

PR #3677, opened 2017-01-21
OpenGL.pm fixes

lordofhyphens:
Replace glVertexPointer_p() call with glVertexPointer_c(), 
Delete VBO buffers after they've been drawn. Both ways work on OpenGL.pm 0.7

Credit to @bubnikv from prusa3d fork, drawn from commit prusa3d@c0b3de6248f6e6d50624b3d2132c823fb65af95e.

Switch HAS_VBO = 0 in 3DScene.pm to not use VBOs.


lordofhyphens:
This should avoid leaking GPU memory.

PR #3718, opened 2017-02-20
Fix octoprint ssl

### LINKED ISSUE #3203: Can't connect to octoprint over SSL
I have my octoprint server exposed via an SSL port so I can access it remotely, but slic3r can't use it.  Is this something that can be trivially fixed? Would be cool =]

Awesome tool, keep up the good work.

Salandora:
This is my attempt to fix #3203.

For me it was necessary to disable the hostname check, no matter what I did perl was not able to verify the hostname, not even if I specify the path to the certificates... maybe someone can take a look at this part.

Next was to remove the hard coded http:// and replace it with a regex check, so basically I check if the url starts with http if not http:// will be appended.
But now you can add a https:// in front of the url so that it now gets recognized as a secure connection.

### DIFF
--- a/lib/Slic3r/GUI/Plater.pm
+++ b/lib/Slic3r/GUI/Plater.pm
@@ -42,6 +42,10 @@ our $PROCESS_COMPLETED_EVENT : shared = Wx::NewEventType;
 use constant FILAMENT_CHOOSERS_SPACING => 0;
 use constant PROCESS_DELAY => 0.5 * 1000; # milliseconds
 
+use Net::SSL;
+$ENV{HTTPS_VERSION} = 3;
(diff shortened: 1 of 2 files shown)


alexrj:
Net::SSL needs to be added as a dependency in Build.PL and this means we need to include openssl in the builds. Since we're in the process of reworking our builds, let's not add one more dependency for now.

Salandora:
Ahh I see okay, well any Idea how to get around Net::SSL? 

I will try a few things the next days and report back when I found a solution.

alexrj:
We're going C++ soon. We can use boost::asio with OpenSSL, but we need to make this dependency optional (i.e. allow people to compile without SSL support).

lordofhyphens:
Incorporated into #4795

PR #3726, opened 2017-02-22
Add option to specify infill/perimeter overlay for bridge areas

tcm0116:
I've been working on a part that is hollow, and thus has a bridging layer as the bottom layer of the top of the part. I noticed that sometimes the infill wouldn't be pressed enough into the perimeter to ensure that it didn't just fall down into the void in the part. I tried adjusting the infill/perimeter overlap, which helped with the bridges but made a mess of the non-bridge infill layers due to the over-extrusion.

This change adds an option to specify the infill/perimeter overlap for bridges. I'm mostly placing this here to get some feedback to see if this is a desirable option, or if there are other techniques in place to address this particular issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -364,6 +364,8 @@ The author of the Silk icon set is Mark James.
         --support-material-extrusion-width
                             Set a different extrusion width for support material
         --infill-overlap    Overlap between infill and perimeters (default: 15%)
+        --bridge-infill-overlap
+                            Overlap between bridge infill and perimeters (default: 15%)
         --bridge-flow-ratio Multiplier for extrusion when bridging (> 0, default: 1)
     
        Multiple extruder options:
--- a/lib/Slic3r/GUI/Tab.pm
+++ b/lib/Slic3r/GUI/Tab.pm
@@ -496,7 +496,7 @@ sub build {
         extrusion_width first_layer_extrusion_width perimeter_extrusion_width 
         external_perimeter_extrusion_width infill_extrusion_width solid_infill_extrusion_width 
         top_infill_extrusion_width support_material_extrusion_width
-        infill_overlap bridge_flow_ratio
+        infill_overlap bridge_infill_overlap bridge_flow_ratio
         xy_size_compensation threads resolution
     ));
     $self->{config}->set('print_settings_id', '');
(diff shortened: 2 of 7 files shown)


alexrj:
You need to share and discuss a full test case here.   Most of the time the endpoints of bridge infill do NOT hit a perimeter, they just end within internal infill.

tcm0116:
I started this discussion in prusa3d's fork, but I determined that your re-worked infill code had not been accepted into their fork. Because it was infinitely easier to add this change to your re-worked infill code, I decided to create this pull request. That original discussion can be found at: https://github.com/prusa3d/Slic3r/issues/134

What's the best way to pickup that discussion as part of this pull request?

lordofhyphens:
I'd suggest summarizing the salient points from the discussion (at minimum,
it should help distill the essence of the request) as a new issue here.

tcm0116:
Consider a part with the following cross section:

cross section

Due to the fact that the bottom layer above the void is a bridging layer, I was running into issues with the strings not attaching to the perimeter as can be see in the image below. I tried increasing the overall infill/perimeter overlap from 15% to 25%, but that made the non-bridge top/bottom layers very messy. Going back own to 20% was a pretty reasonable compromise as the bottom and top layers don't look too bad and most of the bridge strings attached.

fail

My proposal (and this pull request) is to add an option for specifying the infill/perimeter overlap for bridges. This way, I can use 15% for the top and bottom layers and 25% for the bridges to ensure they attach to the perimeter.

lordofhyphens:
@tcm0116 Any other updates to this?

tcm0116:
@lordofhyphens not that I'm aware of. I provided the requested overview of the problem and my solution, but then never heard anything back from you or @alexrj. 

If I remember correctly, this ended up only really being an issue if the total number of perimeters was less than the perimeters setting (i.e. the wall is only 3 perimeters in total with a perimeters setting of 2). However, I don't really remember as it's been quite a while, and I'm not currently in a position to be able to go back and look at this.

lordofhyphens:
Sorry about that, it fell off my list of things to look over and Alex has disappeared into the ether again.

@foreachthing or @VanessaE could you take a look over the problem statement as well? Is there a general case solution available that is not +1 tweak option?

VanessaE:
@tcm0116  maybe you refer to https://github.com/slic3r/Slic3r/issues/3916 ?

foreachthing:
@lordofhyphens I am not really sure what you mean with "not +1 tweak option".
I got this [part.zip](https://github.com/slic3r/Slic3r/files/1937304/part.zip) where the cap needs bridging and the strings are hardly attached. Set to 1 perimeter, the bridging looks as expected (on screen). Set to 2; it might still work. Set to 3 perimeters; back to the issue.

PR #3731, opened 2017-03-02
Cpp11

lordofhyphens:


lordofhyphens:
Merging as the current codebase state breaks, well, everything. Feel free to revert and solve a different way if desired.

Issue #3733, opened 2017-03-03
Add support for custom gcode for triggering solenoid


seanam:
I'm working on a bioprinting project which uses a solenoid for extruding material. The process of modifying the g-code after slicing isn't very easy and am looking for ways to automate / improve it.

1. How difficult would it be to add "Starting extrusion" and "Stopping extrusion" sections to the Printer Settings -> Custom G-code area? This would be really useful for triggering the solenoid on / off.
2. Is there a way to prevent Slic3r from adding the E parameters to the g-code? I'm not using a motor so don't need this. Would it be possible to add a selection in the Filament settings to switch between motor extrusion and solenoid extrusion?
3. Another approach I thought of is using a post-processing script to delete the E parameters (not using a motor) and then and add the solenoid on/off code by check if movements (without E parameter) occurred before or after the code.

Any advice on how to implement this would be very appreciated. I don't have much experience with the slic3r code but can try to adding this if someone points me to the areas that need to be modified.

lordofhyphens:
#848 for a discussion of on/off. 

A "custom constant gcode extrusion" flavor would probably be appropriate for pumps/solenoids. It would need to enable a field to enter in custom on/off gcode that the _extrude() function would use to place the on/off. 

You wouldn't get the benefit of acceleration of course with what I have in mind.

lordofhyphens:
https://bintray.com/lordofhyphens/Slic3r/Slic3r_Branches will have a copy of this implemented for testing purposes.

lordofhyphens:
https://bintray.com/lordofhyphens/Slic3r/Slic3r_Branches should have a test implementation for this when it gets off of the build server, branch name `custom-constant-extrusion`

I've added a switch in the Extruders options, along with a couple gcode entry text boxes to use whatever command you want. 

Currently, the user is assumed to be able to match the flow rate from their weird extruder to the print speed.  Acceleration is not being set for print moves (although it is being set for travel afaik).

A possible refinement would be to override print speed with advanced knowledge of the flow rate instead. 

Please leave feedback in the PR thread.

lordofhyphens:
@seanam if you want to mess around with it check to see if the way I implemented it isn't good for that use case (too much pump action), there's an alternative which I can merge instead: `constant-extruder-fewer-pumps`

It will also be on the bintray link above.

seanam:
@lordofhyphens Thank you for your help! I will test both of these and leave feedback in the PR thread.

[...]

PR #3734, opened 2017-03-03
Implemented simplistic extrusion for constant rate extruders

### LINKED ISSUE #848: Feature Request: Replacing Extruder G-Code Commands with Pump On/Off
I am building a 3D printer that uses a pressurized pump instead of a stepper extruder to print material. Consequently I have no need of the fourth-axis A G-Code commands as there is no filament nor stepper-based extruder. Instead I need a switch-like on/off command when the printhead is either printing or moving to a new location. Basically I just need to turn the pump on and off. I would like to use the M7 and M8 commands (traditionally for the coolant pump) within the G-Code to control this, but I'm not sure where or how to begin modifying Slic3r (or any other useful slicing program) to output M7/M8 instead of the stepper commands on the A axis. Basically this is what the final G-Code code should look like: 

M7 (turns pump on) 
(X and Y movements for perimeter and infill until 1st layer is finished) 
M8 (turns pump off) 
(Printhead moves to new location for the start for 2nd layer) 
M7 (pump on) 
(2nd layer) 

etc. 

Any ideas?

lordofhyphens:
On/off commands only.

User is assumed to match flow rate to print speed.
Fixes #848 #3733

Currently a pump on/off is done for each arc/extrude move. May be more appropriate to push this up into _extrude in Gcode instead of in GcodeWriter.

### DIFF
--- a/xs/src/libslic3r/GCode.cpp
+++ b/xs/src/libslic3r/GCode.cpp
@@ -508,9 +508,12 @@ GCode::_extrude(ExtrusionPath path, std::string description, double speed)
     }
     
     // compensate retraction
-    gcode += this->unretract();
+	if (!this->writer.extruder()->is_constant_rate())
+		gcode += this->unretract();
     
     // adjust acceleration
+	// Don't adjust if using a constant extruder
+	if (!this->writer.extruder()->is_constant_rate())
     {
         double acceleration;
         if (this->config.first_layer_acceleration.value > 0 && this->first_layer) {
--- a/xs/src/libslic3r/GCode.cpp
+++ b/xs/src/libslic3r/GCode.cpp
@@ -586,16 +586,20 @@ GCode::_extrude(ExtrusionPath path, std::string description, double speed)
     {
         std::string comment = this->config.gcode_comments ? description : "";
         Lines lines = path.polyline.lines();
+		
+        if (this->writer.extruder()->is_constant_rate())
+            gcode += this->writer.extruder()->start_extrusion_gcode() + ";\n";
         for (Lines::const_iterator line = lines.begin(); line != lines.end(); ++line) {
             const double line_length = line->length() * SCALING_FACTOR;
             path_length += line_length;
-            
             gcode += this->writer.extrude_to_xy(
                 this->point_to_gcode(line->b),
                 e_per_mm * line_length,
                 comment
             );
         }
+        if (this->writer.extruder()->is_constant_rate())
+            gcode += this->writer.extruder()->stop_extrusion_gcode() + ";\n";
     }
     if (this->wipe.enable) {
         this->wipe.path = path.polyline;
--- a/xs/src/libslic3r/GCode.cpp
+++ b/xs/src/libslic3r/GCode.cpp
@@ -508,12 +508,12 @@ GCode::_extrude(ExtrusionPath path, std::string description, double speed)
     }
     
     // compensate retraction
-	if (!this->writer.extruder()->is_constant_rate())
-		gcode += this->unretract();
+    if (!this->writer.extruder()->is_constant_rate())
+        gcode += this->unretract();
     
     // adjust acceleration
-	// Don't adjust if using a constant extruder
-	if (!this->writer.extruder()->is_constant_rate())
+    // Don't adjust if using a constant extruder
+    if (!this->writer.extruder()->is_constant_rate())
     {
         double acceleration;
         if (this->config.first_layer_acceleration.value > 0 && this->first_layer) {
@@ -586,7 +586,7 @@ GCode::_extrude(ExtrusionPath path, std::string description, double speed)
     {
         std::string comment = this->config.gcode_comments ? description : "";
         Lines lines = path.polyline.lines();
-		
+        
         if (this->writer.extruder()->is_constant_rate())
             gcode += this->writer.extruder()->start_extrusion_gcode() + ";\n";
         for (Lines::const_iterator line = lines.begin(); line != lines.end(); ++line) {
--- a/README.md
+++ b/README.md
@@ -4,9 +4,8 @@ A: Yes.
 
 Slic3r [![Build Status](https://travis-ci.org/alexrj/Slic3r.svg?branch=master)](https://travis-ci.org/alexrj/Slic3r) [![Build status](https://ci.appveyor.com/api/projects/status/8iqmeat6cj158vo6?svg=true)](https://ci.appveyor.com/project/lordofhyphens/slic3r) [![Build Status](http://osx-build.slic3r.org:8080/buildStatus/icon?job=Slic3r)](http://osx-build.slic3r.org:8080/job/Slic3r)
 ======
-Prebuilt Win32 builds:
+Prebuilt Windows (64-bit) and OSX (>10.7) builds:
 * https://bintray.com/lordofhyphens/Slic3r/slic3r_dev/view (from build server)
-* https://bintray.com/lordofhyphens/Slic3r/slic3r_dev/1.3.0-dev (manually packaged)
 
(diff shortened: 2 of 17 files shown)


lordofhyphens:
https://github.com/lordofhyphens/slic3r/tree/constant-extruder-fewer-pumps is a derivative of this PR that reduces the number of on/off to one per planned print move series (loops, etc).

seanam:
I've test out both of the branches now and both turn the solenoid on/off more frequently than I'd like. The `constant-extrusion-fewer-pumps` branch has the same output as `custom-constant-extrusion` except that it also includes the stepper motor E parameters for G1 commands. They also both have the G92 E0 commands which isn't necessary if the "Custom Constant rate extruder" button is selected (although doesn't hurt anything).

Here is some sample g-code to clarify this:

;current Slic3r output for cube
G92 E0;
G0 X40.000 Y40.000 Z0.200 ; move to starting point
M126 ; ON
G1 X40.000 Y60.000 ; extrusion path 1
M127 ; OFF
M126 ; ON
G1 X60.000 Y60.000 ; extrusion path 2
M127 ; OFF
M126 ; ON
G1 X60.000 Y40.000 ; extrusion path 3
M127 ; OFF
M126 ; ON
G1 X40.000 Y40.000 ; extrusion path 4
M127 ; OFF

The `constant-extrusion-fewer-pumps` branch has the same output, except it also includes the E parameters. Both have the G92 E0 code.

;desired Slic3r output for cube
; remove G92 E0
G1 X40.000 Y40.000 Z0.200 ; move to starting point
M126 ; ON
G1 X40.000 Y60.000 ; extrusion path 1
G1 X60.000 Y60.000 ; extrusion path 2
G1 X60.000 Y40.000 ; extrusion path 3
G1 X40.000 Y40.000 ; extrusion path 4
M127 ; OFF

I'm also attaching the g-code files. I've renamed the SLIC3R_VERSION in /xs/src/libslic3r/libslic3r.h to make it clearer which versions were used for slicing. 
[cube_constant-extrusion-fewer-pumps.txt](https://github.com/alexrj/Slic3r/files/826724/cube_constant-extrusion-fewer-pumps.txt)
[cube_custom-constant-extrusion.txt](https://github.com/alexrj/Slic3r/files/826725/cube_custom-constant-extrusion.txt)

lordofhyphens:
Weird, the E axis stuff shouldn't have been written out in the  fewer pumps
branch for starters. Fixed it in GcodeWriter.cpp

Oops. I didn't quite put the on/offs where I thought I did (between loops),
they went between each extrusion move. That's been fixed.

There's a variant I thought up using negative logic -- expect extruder to
be on all the time except on travel moves (negative logic).
That's something I would suggest experimenting with.  On/off is removed
from the extrusion move, and put in the travel move generation code. Let
the first travel move turn the pump on, put a pump off command at the start
of the end gcode to clean up.

I've updated the fewer-pumps branch to reduce even more and added another
branch to do the negative logic, constant-extruder-pump-off-travel. I
should have also removed the retraction moves. I don't think the E0 calls
have been removed yet (some of them would be because I disabled retraction
too).

lordofhyphens:
If you do any work yourself on it you can submit a PR to one of my branches.

seanam:
Both the new `constant-extrude-fewer-pumps` and `constant-extrusion-pump-off-travel` variants work nicely. Here is the output code from each:

`constant-extrusion-fewer-pumps`:
G92 E0
G1 Z0.300 F7800.000
G1 X40.000 Y40.000 F7800.000
G1 F1800
M126;
G1 X40.000 Y60.000
G1 X60.000 Y60.000
G1 X60.000 Y40.000
G1 X40.000 Y40.000
M127; 

If I wanted to remove the G92 E0, would adding an if statement to GCodeWriter.cpp (~line 205) break anything?

`constant-extrusion-pump-off travel`:
G1 Z0.300 F7800.000
G1 X40.000 Y40.000 F7800.000
M126;
; 2nd layer temp also inserted here
G1 F1800
G1 X40.000 Y60.000
G1 X60.000 Y60.000
G1 X60.000 Y40.000
G1 X40.000 Y40.000
M127; 

The `constant-extrusion-pump-off-travel` also looks nice. It adds the speed (F) parameters after the on (M126) command whereas the fewer-pumps method has this before. This shouldn't effect anything though since it is instantaneous. If you specify a different temperature for the second layer though, it does this while the solenoid is open. Not sure if there is a delay or it will keep printing while the temperature adjusts. This version doesn't have the G92 E0 commands which is a bit cleaner.

If I had to choose between the two, I slightly prefer the fewer-pumps approach since the output code seems more well defined. My application is pretty simple but perhaps the solenoid would remain on for more complex things (maybe z-hop but this requires retraction and you mentioned its disabled).

[cube_constant-extruder-fewer-pumps.txt](https://github.com/alexrj/Slic3r/files/827487/cube_constant-extruder-fewer-pumps.txt)
[cube_constant-extruder-pump-off-travel.txt](https://github.com/alexrj/Slic3r/files/827486/cube_constant-extruder-pump-off-travel.txt)

lordofhyphens:
It is disabled as it doesn't make sense (constant velocity extrusion,
retraction is a short move to relieve pressure). I disabled retraction via
appending a case to the needs retraction function in GCode.cpp.

Since you are the person who has the use case, go nuts with tweaking
things. As can be seen the changes got made to GCode and GCodeWriter.cpp.
Worst case scenario your changes don't get to master without refactoring.
But that means you still have access to it for your needs. Overriding the
print speed to match flow rate may something you want to look at as well.

Odds are so long as you are switching based on whether or not the current
version extruder is constant velocity it should be safe. At least it won't
break things for anyone else.

lordofhyphens:
I've merged the "fewer pumps" branch into the main custom-constant-extrusion branch. 
@jtemple3 would you be interested in taking a look at this as well?

alexrj:
Watch out for tabs, use 4 spaces instead.

xs/src/libslic3r/GCode.cpp
```diff
@@ -508,9 +508,12 @@ GCode::_extrude(ExtrusionPath path, std::string description, double speed)
     }
     
     // compensate retraction
-    gcode += this->unretract();
+	if (!this->writer.extruder()->is_constant_rate())
+		gcode += this->unretract();
     
     // adjust acceleration
+	// Don't adjust if using a constant extruder
+	if (!this->writer.extruder()->is_constant_rate())
```
alexrj: Is this actually needed? User can just leave acceleration settings to 0 and Slic3r will not emit any acceleration command.

lordofhyphens: I wanted it to be more broadly compatible with print settings.

lordofhyphens:
@seanam any more reports with this? 

You should turn on autospeed and set the volumetric speed to your flow rate from your special extruder. It *should* work to limit things as you need it to be.

lordofhyphens:
@alexrj Any interest in getting this merged into master, assuming I fix it to use the new gui conventions?

seanam:
@lordofhyphens Sorry I missed the earlier messages. We have been using this slic3r branch and it creates the on/off commands correctly. There is an issue that the xy paths are not sliced correctly when changing the infill, nozzle size and z height parameters. I think this is related to the flow rate calculations that slic3r does and was wondering if it is possible to turn the calculations off and just let the user calibrate it themselves externally (we measure the widths for different speeds on a microscope).

For example, the speed and nozzle size we are using prints line paths 1mm thick and 0.2mm tall. When I set this as the nozzle diameter to 1mm and 100% infill at 0.2mm layer height the print infill does not look correct. I think there are some calculations that associate the nozzle width with the layer height. Is this possible to add a button to turn this off (constant extrusion ON/OFF button)?

Let me know if you have additional thoughts about what is happening

PR #3738, opened 2017-03-05
package Slic3r for OSX post-build.

lordofhyphens:
Thanks to @bubnikv who provided his sample scripts to build and package. The new local-lib arrangement obviated the need for most of the complexity but the dmg generation was a good model to write my own. 

Should work anywhere that has a working, statically-linked perl. Build server output is OSX 10.9, perl 5.24 (perlbrew).

I don't currently have a test case on a system that isn't the build server.
(next part is getting the build server itself going automagically)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@ xs/MANIFEST.bak
 xs/assertlib*
 .init_bundle.ini
 local-lib
+package/osx/Slic3r*.app
+*.dmg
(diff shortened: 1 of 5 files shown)


lordofhyphens:
Okay, *now* it should work. Uses PAR::Packer to pull the few core modules that don't get built in local-lib.

alexrj:
Looks like an excellent work there.

PR #3747, opened 2017-03-10
Optimistic print time estimator

lordofhyphens:
Add an optimistic print time estimator to Slic3r. Considers just the print and travel moves under ideal circumstances. The information is collected during gcode generation per-extruder and then summed at the end. 

This PR requires that #3669 be merged first (adds slicing statistic box)

Implements #1766 (along with #3669).


lordofhyphens:
Obviously it will be even less accurate if you're not telling Slic3r to control acceleration.

lordofhyphens:
I am very okay with that. Perhaps we can roll it into a more general plugin
arrangement instead? Something I have been mulling over.

PR #3751, opened 2017-03-11
Added some 'send to printer' options in plater tab ( related to #3657 )

gege2b:
This PR is for allowing a smooth workflow (i.e. without any dialog - see #3657 for the related issue/feature request ) when using the Send to Printer button, restoring the old behavior, but still keep the new features if needed

Things to know : 
- options are in plater tab to allow for a "per print" settings on the fly. 
- options aren't saved accros restart (I'm still not sure how to save settings from plater tab, using the global settings)

feel free to comment / ask for change or whatever ^^

### DIFF
--- a/lib/Slic3r/GUI/Plater.pm
+++ b/lib/Slic3r/GUI/Plater.pm
@@ -11,7 +11,7 @@ use Slic3r::Geometry qw(X Y Z MIN MAX scale unscale deg2rad rad2deg);
 use LWP::UserAgent;
 use threads::shared qw(shared_clone);
 use Wx qw(:button :cursor :dialog :filedialog :keycode :icon :font :id :listctrl :misc 
-    :panel :sizer :toolbar :window wxTheApp :notebook :combobox);
+    :panel :sizer :toolbar :window wxTheApp :notebook :combobox :checkbox);
 use Wx::Event qw(EVT_BUTTON EVT_COMMAND EVT_KEY_DOWN EVT_LIST_ITEM_ACTIVATED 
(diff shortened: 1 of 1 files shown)


alexrj:
I do appreciate this, however I'm very worried about cluttering the plater window. Slic3r has still too many things in it compared to more user-friendly software out there. I'd prefer to keep the dialog, which is consistent with what happens when clicking on the "Export G-code..." button for saving to a SD, and with the print dialog of any normal application. Having an always visible "Rename" checkbox that eventually shows a dialog is not a common UX pattern...
So I'd prefer to turn the filename dialog into a more complete dialog with the other options you're proposing.

gege2b:
Hi Alessandro
I understand your point (and to be honest, Slic3r is the only slicer I use since a while, so I don't really know how the other slicers looks like)

My goal with these options is to get rid of any dialog (even more complete) that can show up when sending a gcode to the printer, for people like me who don't care about renaming gcode (filename output settings are great), overriding them (timestamp rules) or start the print after upload (using octoprint for printer control)

So, what about putting these options in the printer tab OR preference dialog instead (printer tab makes more sense, IMHO) ?

So I can do another PR with 3 checkboxes : 

"when sending file to printer, ask for :
- renaming the gcode
- override existing file
- print after upload"

Default options can be set to mimic the actual behaviour (i.e. all checked).
Losing the "per print" usage is something I can live with :)

gege2b:
@alexrj  I did this on another branch (I'm not a git/hub pro so I don't want to confuse things in this PR)
https://github.com/gege2b/Slic3r/tree/OctoprintOptions-printerTab

Do you want me to make another PR or to give up ? :D

gege2b:
I'll close this PR and open a new one, with options in printer tab (I'm not brave enough to figure out how to customize the filename dialog)

and then, you'll decide what you want to do with this :-)

PR #3765, opened 2017-03-15
Added options to printer tab to manage 'Send to Printer' related dialogs

gege2b:
Replace PR #3751 

Moved options from plater to printer settings

One thing I don't really know how to handle is the CLI parameters (actually I leave them empty, because I don't think they make sense, but correct me if I'm wrong )

### DIFF
--- a/lib/Slic3r/GUI/Plater.pm
+++ b/lib/Slic3r/GUI/Plater.pm
@@ -50,7 +50,7 @@ sub new {
     my $self = $class->SUPER::new($parent, -1, wxDefaultPosition, wxDefaultSize, wxTAB_TRAVERSAL);
     $self->{config} = Slic3r::Config->new_from_defaults(qw(
         bed_shape complete_objects extruder_clearance_radius skirts skirt_distance brim_width
-        serial_port serial_speed octoprint_host octoprint_apikey
+        serial_port serial_speed octoprint_host octoprint_apikey octoprint_ask_rename octoprint_ask_override octoprint_ask_print
     ));
     $self->{model} = Slic3r::Model->new;
     $self->{print} = Slic3r::Print->new;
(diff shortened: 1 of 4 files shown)


lordofhyphens:
CLI parameters accessible from config are basically the same as any other
parameter (if you copypasta from another configuration item the items are
available on the CLI).


gege2b:
yep! but does it make sense to have CLI parameters allowing the display of dialogs ? ;) 

That's the point of these options : allow (or not) to show dialogs when sending to octoprint, make sense in a GUI...

Note : I've never used CLI version

lordofhyphens:
https://github.com/alexrj/Slic3r/blob/master/lib/Slic3r/GUI/Preferences.pm ?

gege2b:
After ripping off these options from the plater, I considered adding them to the preference dialog, but it would make no sense for people who don't use Octoprint.
And we would loose the per printer settings, witch are important I think

But again, correct me if I'm wrong

Does every options presents in print/filament/printer tabs should have a CLI parameter ?

lordofhyphens:
Not if it does not make sense from a CLI context. I don't know if assigning
the CLI option is compulsory from a code standpoint.

alexrj:
@gege2b, I know you'll hate me and you don't agree with this. However I always try not to add too many options, and this is really overkill. The usual print/export pattern for any application is: click on button -> get one single dialog with the print/export options -> click OK.
I now merged all the options in a single dialog. I understand it's one more click, but it's exactly the same number of clicks people have to do when saving the G-code file somewhere, and it's a very common UX pattern. Much more than moving options outside of the flow.
But there's a hidden feature just for you: if you hold the Alt key when clicking on the "Send to printer..." button the dialog will not be shown (and last settings are used) :-)
Thank you for understanding.

gege2b:
at the begining of your post yeah I kinda hate you, and then I read the ALT key cheatcode... ;) 
It's so simple... I don't know why I didn't think about it myself !

And you're wrong, I totally understand your point, and at least I tried and proposed something (and it gave me the opportunity to dig into the UI code)

so, well, thank you a lot !

lordofhyphens:
Issue a pull request against The Manual for that information and I think I
will be able to accept it ;)

gege2b:
OK,  will do this, later this week ^_^

PR #3796, opened 2017-03-23
Ported parts of Slice to xs

platsch:
Ported a large portion of the slice method in Print/Object.pm to C++ as mentioned in https://github.com/alexrj/Slic3r/pull/1386#issuecomment-284928987. I wanted this to have the adaptive slicing stuff entirely in C++, but think it is generally useful.

Based on @bubnikv's code, but without the `SlicingParameters` object and the different approach of layer generation.


alexrj:
Thank you a lot for this!
So this is a 1:1 direct port of the current code in master? No additions or changes, just language porting?

platsch:
I hope so...
Of course I had to make some changes to get it compatible. Some things are:

- I added a check in adjust_layer_height to avoid a result of 0 if z_steps_per_mm is too small
- The generation of a vector containing all layer positions is moved to a separate method, it feels cleaner this way and is easier to integrate with things like adaptive slicing
- The matching of the top layer is done during the generation of layer positions, but otherwise untouched. While I think the feature is generally good idea but I don't like the current implementation because it doesn't work in some cases and is pretty obscure for the user.
- I had to pull over some extensions to `Surface.hpp` and `SurfaceCollection.hpp` to use @bubnikv's code, but everything looks straight forward.

Otherwise it should be 1:1

alexrj:
Okay, thank you for clarifying. I'm definitely going to merge this, but I need to wait until @lordofhyphens manages to upgrade the Windows build server. We have an older compiler there which doesn't support `emplace()`, so merging this would break builds.

platsch:
Ok, I didn't even notice the emplace-call, that's @bubnikv's code. If it's a problem we could probably avoid the call (it's only one occurrence), but @lordofhyphens seems to be working a lot on the build servers :-)

lordofhyphens:
Yeah, I will know tonight if it works (we will have gcc 6.3.0 on the build
server on Windows if it does).

lordofhyphens:
Build server is upgraded.

PR #3797, opened 2017-03-23
working on rc file to define manifest and filetype

lordofhyphens:


lordofhyphens:
Tested, it does what I expect it to now.

PR #3804, opened 2017-03-24
force --gui for shell during compilation if -DFORCE_GUI is set

### LINKED ISSUE #3803: Slic3r not opening when executing stl file
### Version
Prusa Version

Windows 8.1

### Behavior
When I asociated .stl files into executing Slic3r, is expected to open Slicer with the .stl file already loaded, but it creates a .gcode output file without opening the software. Is it a common problem?

lordofhyphens:
Fixes #3803


lordofhyphens:
Done and tested on my local system.

PR #3820, opened 2017-03-29
Added translation string markers to libslic3r PrintConfig text.

lordofhyphens:
Recommend merging if only to avoid headaches later (and give a heads-up to people adding options that they need to wrap strings in libslic3r with the macro for things to work). 

I think that the branch should stay up to collect translations (provide another PR when they are known to work). 

I am unsure if we should include the compiled files here in the repository or run a script over them to generate them during the build process.


lordofhyphens:
The macro calls have no other effects and I used them rather than something wx (or gettext) specific for portability reasons.

PR #3823, opened 2017-03-30
add opengl to installed packages to make opengl.pm happy

lordofhyphens:
PR to bang on server.


lordofhyphens:
Nope, not helping at all.

PR #3825, opened 2017-03-30
Visivalingam simplification algorithm

fuchstraumer:
Howdy - Slic3r has been super helpful to me in the process of writing a custom slicer at work (it should be open-sourced eventually: my contract is for a well-liked govt. institution in the US). I'd like to pay back the debt (in any way I can, even as small as this), so I quickly implemented/transcribed the simplification algorithm I use in my slicer. It's significantly faster than Douglas-Puecker, doesn't require recursion, and tends to do a better job at preserving details while still managing to simplify just as much (if not more). A wonderfully succinct description of this algorithm can be found here, along with a neat demo:

https://bost.ocks.org/mike/simplify/

I'll also link a gist containing the approximate test setup I ran to initially compare it to the Douglas-Puecker algorithm. This gist was something I just quickly adapted a few minutes ago - it might contain errors that are down to differences between my project and Slic3r, but I tried my best to rectify those. 

gist link: https://gist.github.com/fuchstraumer/95fa4d21147d83a9a4bd220aa29ad5b2

I've tested it with a ton of models in my slicer and haven't had a single issue yet, and usually find it able to cut the point cut in half even on highly curved segments (like the legs of the octopus test print, or highly detailed text on the bottom of a print). These were after I just started rolling with the algorithm as the default simplifier, though, so I didn't include them in the tests. I've kept images and performance-test datasets around, if those would help, though. Tuning the "tolerance" amount will probably have to happen, as this is a different type of tolerance than what's used in Douglas-Puecker.

This implementation also isn't mine alone, so I've linked the location I got it from. I adapted it slightly to just use std::algorithm and be a bit more "Modern C++"-y, namely by using std::make_heap and lambdas to avoid adding single-use methods to the slic3r namespace. The vis_node struct can also probably be moved into the scope of the method, to further remove clutter from the general namespace.

Lastly, I have a concern and some thoughts for improvements. I'm worried about possible memory leaks due to exceptions in the method itself, around "node_list". I also think I could use a forward_list to make this even faster, but its already speedy enough that this wasn't a priority. Any fixes or bugs I encounter in my project will be added to my fork/branch, and I'll submit another PR then (assuming this one is even accepted).

Thanks! I really hope this might prove useful, and I'll be keeping an eye out for any other ways I can contribute. This is only my second PR, and my first one didn't end so well...

### DIFF
--- a/xs/src/libslic3r/MultiPoint.cpp
+++ b/xs/src/libslic3r/MultiPoint.cpp
@@ -212,4 +212,151 @@ MultiPoint::_douglas_peucker(const Points &points, const double tolerance)
     return results;
 }
 
+/*
+
+    struct - vis_node
+
+    Used with the visivalignam simplification algorithm, which needs to be able to find a points
+    successors and predecessors to operate succesfully. Since this struct is only used in one
+    location, it could probably be dropped into a namespace to avoid polluting the slic3r namespace.
+
+    Source: https://github.com/shortsleeves/visvalingam_simplify
+
+    ^ Provided original algorithm implementation. I've only changed things a bit to "clean" them up
+    (i.e be more like my personal style), and managed to do this without requiring a binheap implementation
+
+*/
+
+struct vis_node{
+    vis_node(const size_t& idx, const size_t& prev_idx, const size_t& next_idx, const double& _area)
+    // Indices into a Points container, from which this object was constructed
+    size_t pt_idx, prev_idx, next_idx;
+    // Effective area of this "node"
+    double area;
+    // Overloaded operator used to sort the binheap
+    // Greater area = "more important" node. So, this node is less than the 
+    // other node if it's area is less than the other node's area
+    bool operator<(const vis_node& other){
+        return this->area < other.area);
+    }
+}
+
+Points
+MultiPoint::visivalingam(const Points& pts, const double& tolerance)
+{
+    // Make sure there's enough points in "pts" to bother with simplification.
+    assert(pts.size() >= 2);
+
+    // Result object
+    Points results;
+
+    // Lambda to calculate effective area spanned by a point and its immediate 
+    // successor + predecessor.
+    auto effective_area = [pts](const size_t& curr_pt_idx, const size_t& prev_pt_idx, const size_t& next_pt_idx)->coordf_t{
+        const Point& curr = line[curr_pt_idx];
+        const Point& prev = line[prev_pt_idx];
+        const Point& next = line[next_pt_idx];
+        // Use point objects as vector-distances
+        const Point curr_to_next = next - curr;
+        const Point prev_to_next = prev - curr;
+        // Take cross product of these two vector distances
+        const coordf_t cross_product = static_cast<coordf_t>(curr_to_next.x * prev_to_next.y) - static_cast<coordf_t>(curr_to_next.y * prev_to_next.x);
+        return 0.50 * abs(cross_product);
+    }
+
+    // We store the effective areas for each node
+    std::vector<coordf_t> areas;
+    areas.reserve(pts.size());
+
+    // Construct the initial set of nodes. We will make a heap out of the "heap" vector using 
+    // std::make_heap. node_list is used later.
+    std::vector<vis_node*> node_list;
+    node_list.resize(pts.size());
+    std::vector<vis_node*> heap;
+    heap.reserve(pts.size());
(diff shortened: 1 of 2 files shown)


lordofhyphens:
I'll take a look at it later, it certainly looks a promising. 

Could you write up a few unit tests for it to verify that it is operating correctly? At least replace the usage or modify the function that uses it to optionally use it (and update the test and xsp interface to test with it too) to verify that our regular test suite passes?

We do support C++11 in the codebase now (and it's recommended), so go nuts.

lordofhyphens:
@fuchstraumer Whoops! https://travis-ci.org/alexrj/Slic3r/jobs/216945632

lordofhyphens:
If you want to chat in realtime, I'm usualy hanging around on at least one of my nicks in #slic3r on freenode IRC.

fuchstraumer:
Oops! I just wrote this up in VS Code, not an actual IDE, so those small formatting errors don't get caught. 

I'm about to head home from work, but I'll make those changes when I get home (along with fixing that silly mistake bug, lol).  Will the auto-checks run once I commit the changes to my branch, or is there some way you or I will have to queue them?

Excited to help though, and glad to know that it looks promising at least. This is already better than my last PR. I can hope on freenode once home as well - handle there will be fuchstraumer, most likely. I'd be glad to chat there as well, especially if it helps.

lordofhyphens:
Travis and Appveyor will trigger every time you update/push the branch the pull request is sourced from. It's automatic, so you don't need to do anything else there. 

If you want to make using this simplifcation method a config option, here's a short list of the files that probably need to be touched (copy/paste an existing configuration item). That should let you add the widgets necessary to run the relevant parts of the test suite with this algorithm.
https://github.com/alexrj/Slic3r/wiki/Code:-Adding-new-option-to-GUI-CLI

I am interested in having this in Slic3r. If it's equivalent in quality to Douglas-Puecker and faster for our use cases then I suspect it will help a lot with over-detailed models.

lordofhyphens:
````
src/libslic3r/MultiPoint.cpp:256:29: error: ‘line’ was not declared in this scope
         const Point& curr = line[curr_pt_idx];
````

lordofhyphens:
Any updates to get this PR building?

If you need to test locally you can grab the docker image we use, etc. Might be faster for checking for compile issues than setting up a full native build environment.

fuchstraumer:
Sorry - was busy all this weekend doing stuff mostly away from the computer. I'm trying to just build it on the windows subsystem for Linux right now, since that seems like it'll be easier to setup. I'll comment again when I get it building.

fuchstraumer:
Hey - sorry that I seemingly forgot about this. I was just never able to get Slic3r to build on any of my systems, not even Ubuntu on my dual-boot surface.

Regardless, I've found an algorithm that has _completely_ blown me away with how well it simplifies. I'm now writing GCode in my own slicer, and the files were coming out in the tens of MB - ew. I found an iterative, not recursive, version of the Douglas-Peucker-Ramer algorithm and cut my filesize down by 500% in most test cases. It does quite well with reducing the complexity of curving segments to something that's not a nightmare!

The code can be found in this gist, if y'all are curious: https://gist.github.com/fuchstraumer/9421573fc281b946e5f561758961212a

lordofhyphens:
@fuchstraumer willing to give a build another shot with the new algorithm? Also the gist died :(

bubnikv:
By the way, I think the vanilla Douglas-Peucker or Visivalingam will happily create self-intersections in the contour. There are modifications, which try to avoid them.

https://pdfs.semanticscholar.org/8130/b7b306175d9ae28bbf1aa1f252ef68750624.pdf

fuchstraumer:
Unfortunately I really don't have time to work on this anymore (honestly just so overwhelmed at work!). It came about because I was writing a slicing engine back then. I'd love to investigate the algorithm in that paper, but implementing those from papers is always a pretty big time sink (even if you get source code, you have to spend time optimizing and cleaning it up, 9/10 times)

I've updated the gist with the improved Douglas-Puecker algorithm. It shouldn't change how simplification works since it's effectively the same technique, but it's iterative instead of recursive so is a bit easier to debug and tweak, and should reduce RAM usage since we're not creating new storage during each iteration (as I believe that happens with the current impl?)

https://gist.github.com/fuchstraumer/fe399b249bd447e776a0ab5baa559a2d

PR #3826, opened 2017-03-31
Fix regression from 35fc1de

xoan:
which hides settings from PresetEditor but leaves Controller tab always visible


lordofhyphens:
Good catch.

PR #3830, opened 2017-04-01
[i18n] French translation

gege2b:
Hi there

My two cents
I read somewhere that we keep sticking with PR for now, so here it is

I'm not used to POedit so I hope I didn't mess too much


lordofhyphens:
Thanks!

PR #3831, opened 2017-04-01
Slic3r packaging for Linux via TravisCI

lordofhyphens:
Revised .yml file to grab dependencies and build Slic3r.


lordofhyphens:
It's not 100% complete, but it's worth merging in at this point and dealing with the appimage package later. 

@alexrj I've factored out a few of the utility functions (setting variables, etc) to package/common/util.sh if you want to take advantage of them in make_dmg.sh or add a short script to replace the upload to your personal directory. 

The one thing we "lose" at present is that it will only produce one build on travis. If you still want it to build on 3 different perl versions we'll need to set up a more involved caching system (rebuilding each perl version with gcc 4.9 with brewperl, archiving them and their dependencies for xsgui, etc)

PR #3856, opened 2017-04-05
I18n - fix for #3846

### LINKED ISSUE #3846: Tooltip for Filament Settings/Cooling/Disable fan
### Version
1.3.0 dev

Windows 10 1607

### Behavior
Tooltip for Filament Settings/Cooling/Disable fan for the first...________...layers
reads "You can set this to a positive value to disable fan at all during the first layers, so that is does not make adhesion worse. (default 3)"

Might be better worded.........
"This disables the fan completely for the first # of layers to aid in the adhesion of media to the bed. (default 3)"

ruggb:
this fixes #3846 - this is my very first attempt so check it out.......

### DIFF
--- a/translation/en_US.po
+++ b/translation/en_US.po
@@ -285,10 +285,10 @@ msgstr "Disable fan for the first"
 
 #: xs/src/libslic3r/PrintConfig.cpp:169
 msgid ""
-"This disables the fan completely for the first # of layers "
+"This disables the fan completely for the first N layers "
 "to aid in the adhesion of media to the bed. (default 3)"
 msgstr ""
-"This disables the fan completely for the first # of layers "
+"This disables the fan completely for the first N layers "
 "to aid in the adhesion of media to the bed. (default 3)"
 
 #: xs/src/libslic3r/PrintConfig.cpp:170
--- a/translation/en_US.po
+++ b/translation/en_US.po
@@ -285,11 +285,11 @@ msgstr "Disable fan for the first"
 
 #: xs/src/libslic3r/PrintConfig.cpp:169
 msgid ""
-"You can set this to a positive value to disable fan at all during the first "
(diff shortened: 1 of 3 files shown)


translation/en_US.po
```diff
@@ -285,11 +285,11 @@ msgstr "Disable fan for the first"
 
 #: xs/src/libslic3r/PrintConfig.cpp:169
 msgid ""
-"You can set this to a positive value to disable fan at all during the first "
-"layers, so that it does not make adhesion worse."
+"This disables the fan completely for the first # of layers "
```
lordofhyphens: I think you should avoid using symbols, perhaps "first N layers" instead?

ruggb: yea, I couldn't decide on that..........thx

gege2b:
Hi @ruggb 
When doing the french translation, I used the word "plateau" to translate "bed", maybe you should use the same word to keep coherence ?

And I would suggest, for the first sentence : 
"Cela désactive complètement le ventilateur pour les N premières couches"

lordofhyphens:
@gege2b I figured that translators made whatever editing choices as got the point across in their translations.  Not like I can tell the difference anyway ;) 

Feel free to update your translation in a separate PR if you think the French version should be updated (the English text, however, is the lookup key so any revisions to that text must be committed to the other translations).

I think this particular scenario highlights the need of having a separate repository that I can just give contributor access to translators.

gege2b:
@lordofhyphens yep, I think you're right for the separate repo, it should make translation integration smoothier 

Right now, I have no git/poedit or whatever else available to make the changes myself (and can't install them). So it was easier and faster to tell the "needed" changes

lordofhyphens:
It's all text, so even github webedit will do ;)

PR #3863, opened 2017-04-06
Update FillRectilinear.cpp

NateTG:
Changes cubic infill so that it forms closed parallelepiped cells and infill walls are flat.  The old version formed a mix of cells and long chambers.


alexrj:
Makes sense, thank you!

PR #3864, opened 2017-04-06
Adding Zoom

### LINKED ISSUE #3233: Feature request: Zooming without mousewheel
I have not found the possibility of zoom without using the mouse wheel in Slic3r interface. I think it would be useful for people who use a laptop with a trackpoint (eg. Lenovo) or without traditional mouse.

It would be useful to add the possibility of zoom via keyboard (eg. Ctrl - + / ctrl--) or with mouse and a key mod (eg. Ctrl, Shift).

Im using:
Linux 4.3.0-1-amd64 #1 SMP Debian 4.3.3-7 (2016-01-19) x86_64 GNU/Linux
and Slic3r 1.2.9

Samir55:
Fixes #3233

### DIFF
--- a/lib/Slic3r/GUI/3DScene.pm
+++ b/lib/Slic3r/GUI/3DScene.pm
@@ -330,6 +330,20 @@ sub set_viewport_from_scene {
     $self->_dirty(1);
 }
 
+sub zoom{
+    my ($self, $direction) = @_;
+    print ($direction);
+    if( $direction eq 'in'){
+        $self->_zoom($self->_zoom / (1-0.3));
+    }
+    elsif($direction eq 'out'){
+        $self->_zoom($self->_zoom / (1+0.3));
+    }
+    $self->on_viewport_changed->() if $self->on_viewport_changed;
+    $self->_dirty(1);
+    $self->Refresh;
+}
+
 # Set the camera to a default orientation,
 # zoom to volumes.
 sub select_view {
--- a/lib/Slic3r/GUI/3DScene.pm
+++ b/lib/Slic3r/GUI/3DScene.pm
@@ -332,7 +332,6 @@ sub set_viewport_from_scene {
 
 sub zoom{
     my ($self, $direction) = @_;
-    print ($direction);
     if( $direction eq 'in'){
         $self->_zoom($self->_zoom / (1-0.3));
     }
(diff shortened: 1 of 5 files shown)


lordofhyphens:
Needs a review from @alexrj as this is from a GSoC student.

lib/Slic3r/GUI/3DScene.pm
```diff
@@ -330,6 +330,20 @@ sub set_viewport_from_scene {
     $self->_dirty(1);
 }
 
+sub zoom{
+    my ($self, $direction) = @_;
+    print ($direction);
```
alexrj: Remove this debugging statement :)

lib/Slic3r/GUI/MainFrame.pm
```diff
@@ -631,6 +632,11 @@ sub select_view {
     $self->{plater}->select_view($direction);
 }
 
+sub zoom{
```
alexrj: This method is unneeded. Just call `$self->{plater}->zoom()` directly from the menu item handler.

lib/Slic3r/GUI/Plater.pm
```diff
@@ -2432,6 +2432,11 @@ sub select_view {
     }
 }
 
+sub zoom{
+    my ($self, $direction) = @_;
+    $self->{canvas3D}->zoom($direction);
```
alexrj: This will only zoom the 3D plater, and not the other tabs (2D plater, Preview, Toolpaths). It should zoom the currently active tab and leave the others untouched.

Samir55: I've updated my code. but there are two concerns:
* Applying any changes to 3d platter tab by Mouse or keyboard also affects the platter in the preview tab.
* Also There is no Mouse Zoom in the 2D platter tab.

slic3r.pl
```diff
@@ -6,7 +6,7 @@
 BEGIN {
     use FindBin;
     use lib "$FindBin::Bin/lib";
-    use local::lib '--no-create', "$FindBin::Bin/local-lib";
+    use local::lib  "$FindBin::Bin/local-lib";
```
alexrj: Why changing this line? This will break packaging.

Samir55: O.K I will remove that change but When I add '--no-create' It gives me this error (Unknown import argument: --no-create at /usr/share/perl5/local/lib.pm line 49.
BEGIN failed--compilation aborted at slic3r.pl line 9.
),I am on Ubuntu 14.04

lordofhyphens:
That is because your version of local::lib is ancient.

Issue #3862, opened 2017-04-06
Background processing don't trigger when new object is added


gege2b:
Hi all

There are some issue about background processing opened, but no one with these symptoms
my appologies if this is a duplicate 

### Version
04b59af (current build from bintray)

### Operating system type + version
_What OS are you using, and state any version #s_
Windows 10

### Behavior
when adding a new object, the background processing didn't trigger. 
If i'm on the preview tab when adding the object, it is emptied first (seems logic) but that's all

I need to change a whatever setting to trigger the background processing (as expected)

* _Steps needed to reproduce the problem_

- Open slic3r with background processing enabled
- Load an object
- Nothing is done
- Change a setting and see the BP begin

* _Expected Results_

Background processing to trigger when adding an object

* _Actual Results_

Nothing happens

#### STL/Config (.ZIP) where problem occurs
_Upload a zipped copy of an STL and your config (`File -> Export Config`)_

config here : 
[slic3r_config.zip](https://github.com/alexrj/Slic3r/files/903063/slic3r_config.zip)

STL is not important

lordofhyphens:
@gege2b just as a fyi we've migrated away from bintray, http://dl.slic3r.org/dev has everything new and it's sorted by date/time.

lordofhyphens:
Well, I can't reproduce with those steps on Linux or on my Windows 10 system with the default config or the provided configuration. 
04b59af

1. Started Slic3r.
1. Ensured background processing was on.
1. Loaded 1 STL. It started slicing. 
1. Loaded a different STL. It started slicing.

I changed no additional configuration options.

foreachthing:
I can reproduce this on my computer with Windows 7 Pro and slic3r-master.2017.04.05.885.66c7eb7 (1.3.0-dev).

1. Start Slic3r (Background processing is turned on).
2. Import STL.
3. .... nothing! Slice3r tells me: Loaded modelname.stl and just sits there until I change a setting.

By only clicking a rotate button, Slic3r starts slicing.

gege2b:
@lordofhyphens yep I know, I said bintray but I was thinking of the dev from the slic3r website, my bad

I tested right now on my Kubuntu 16.04 with a freshly pulled repo ( ab278f03e17873dc423e011b49da4a480fb806fd ) and I can reproduce it again with the same steps and default config

xoan:
I can reproduce it too.

[...]

PR #3873, opened 2017-04-10
single line fix for #3862

gege2b:
This line is present in all object related function (remove(), [de|in]crease(), mirror(), etc...) but seems to be missing in the add() function, fix the #3862 issue at least on my PC (I still need to test on my linuxes but I'm pretty confident)

And as long as I tested, it didn't introduce any unwanted behaviour.

### DIFF
--- a/lib/Slic3r/GUI/Plater.pm
+++ b/lib/Slic3r/GUI/Plater.pm
@@ -804,6 +804,7 @@ sub add {
     
     my @input_files = wxTheApp->open_model($self);
     $self->load_file($_) for @input_files;
+	$self->on_model_change;
 }
 
 sub add_tin {


lordofhyphens:
Well, I was able to reproduce the bug... with your patch applied.

gege2b:
ahahaa... well, there is something wrong either on my side or on yours...

with this patch applied, the problem is solved on all my linux PCs (I've never managed to get a working windows build)

I think we need other people to test this and see what behaviour they get

lordofhyphens:
I think this indicates that the problem is more complex than it seems, as
without your code I don't see the symptoms on Windows or my Debian system.

Pretty normal for background processing, IMHO.

gege2b:
seems legit
looks like this issue is now officially too complicated for my knowledges

alexrj:
`on_model_change()` is called by `load_model_object()` which is called by `add()`, so this change doesn't help. :)

gege2b:
OK at least I tried :)

PR #3877, opened 2017-04-13
Addressed logic of enforced supports. 

### LINKED ISSUE #3836: Enforce Supports not functioning
Fresh build on April 2, 2017: 1.2.9-901-g90623c4
OS: Debian Stable

Description:
"Support material -> Enforce support for the first N layers" doesn't seem to do anything.
I have Supports disabled, then any number entered in the Enforce option.

If I'm understanding correctly, this would allow me to have supports generated for N layers, and the rest of my object (above that) be support-less.

jaggzh:
There are inconsistent rare crashes which I cannot find as related to my changes.

Fixes #3836

### DIFF
--- a/lib/Slic3r/Print/Object.pm
+++ b/lib/Slic3r/Print/Object.pm
@@ -211,7 +211,11 @@ sub generate_support_material {
     
     $self->clear_support_layers;
     
-    if ((!$self->config->support_material && $self->config->raft_layers == 0) || scalar(@{$self->layers}) < 2) {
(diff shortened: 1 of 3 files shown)


lordofhyphens:
image

lordofhyphens:
image
This is with 50 layers of enforced support.

lordofhyphens:
I haven't been able to reproduce any of the crashes.

alexrj:
The proposed changes make sense to me.

PR #3883, opened 2017-04-15
Reversed order of cooling buffer and spiralize, as suggested by @bubnikv

### LINKED ISSUE #2371: slic3r 1.2.0, 1.2.1, 1.2.2 respect min print speed / min layer time when in Spiral Vase Mode
Win64, but I don't think that matters.  When using spiral vase mode, it appears that slic3r is using the cooling parameters of "min_print_speed" and "slowdown_below_layer_time", resulting in a print speed that's equal to "min_print_speed" for the layer > bottom_solid_layers..

lordofhyphens:
Fixes #2371

### DIFF
--- a/lib/Slic3r/Print/GCode.pm
+++ b/lib/Slic3r/Print/GCode.pm
@@ -642,13 +642,7 @@ sub process_layer {
         }
     }
     
-    # apply spiral vase post-processing if this layer contains suitable geometry
-    # (we must feed all the G-code into the post-processor, including the first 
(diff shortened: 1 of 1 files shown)


alexrj:
Er no, the original issue is not a bug, and this "fix" is based on the wrong assumption that CoolingBuffer considers the short Z increments, which is not true...

lordofhyphens:
I agree with your assessment, mostly forgot to close the PR myself.

Issue #1405, opened 2013-08-24
Feature Request: Allow to set shell thickness


zdar:
Hi,
I'd would be nice to be able to set arbitrary shell thicknes instead of setting number of layers/perimentrs.

For example to set horizontal shells to 1mm and vertical shells to 1.5mm --> Number of layers/perimetrs would be calculated by slic3r automatically depending on layer and extruder properties.

lordofhyphens:
What does this solve other than porting a issue Cura has with having to warn you when it's not cleanly divisible over to Slic3r? Also, unless you fix the perimeter width there's a ton of possible combinations and no good heuristic that I can think of offhand to select the correct one.

alexrj:
I tend to agree with considerations by @lordofhyphens, but I've seen that it might be useful to have this in addition to perimeters/top/bottom, defined as "Minimum Shell Thickness". Number of perimeters would be increased until that thickness is achieved, regardless of their divisibility. This option would be useful to display in the plater overrides/shortcuts.

curieos:
I'm looking into this. I've added the option in the UI and I'm working on some checks for spiral vase.

curieos:
I created a pull request for this feature

lordofhyphens:
Merged PR is incomplete; sets minimum shell only.

PR #3885, opened 2017-04-16
Implemented feature request #1405 minimum shell thickness option

curieos:
This adds a new print config option, minimum shell thickness. It works as described in the issue page by alexrj. The current configuration, when paired with the generate extra perimeters option, will only add perimeters when both the normal and extra perimeters combined don't meet the shell thickness requirements.

### DIFF
--- a/lib/Slic3r/GUI/PresetEditor.pm
+++ b/lib/Slic3r/GUI/PresetEditor.pm
@@ -407,7 +407,7 @@ sub options {
     return qw(
         layer_height first_layer_height
         perimeters spiral_vase
-        top_solid_layers bottom_solid_layers
+        top_solid_layers min_shell_thickness bottom_solid_layers
         extra_perimeters avoid_crossing_perimeters thin_walls overhangs
         seam_position external_perimeters_first
         fill_density fill_pattern top_infill_pattern bottom_infill_pattern fill_gaps
@@ -485,6 +485,7 @@ sub build {
         {
             my $optgroup = $page->new_optgroup('Vertical shells');
             $optgroup->append_single_option_line('perimeters');
+            $optgroup->append_single_option_line('min_shell_thickness');
             $optgroup->append_single_option_line('spiral_vase');
         }
         {
@@ -759,10 +760,11 @@ sub _update {
     
     my $config = $self->{config};
     
-    if ($config->spiral_vase && !($config->perimeters == 1 && $config->top_solid_layers == 0 && $config->fill_density == 0 && $config->support_material == 0)) {
+    if ($config->spiral_vase && !($config->perimeters == 1 && $config->min_shell_thickness <= $config->external_perimeter_extrusion_width && $config->top_solid_layers == 0 && $config->fill_density == 0 && $config->support_material == 0)) {
         my $dialog = Wx::MessageDialog->new($self,
             "The Spiral Vase mode requires:\n"
             . "- one perimeter\n"
+            . "- a shell thickness no larger than the extrusion width\n"
             . "- no top solid layers\n"
             . "- 0% fill density\n"
             . "- no support material\n"
@@ -771,6 +773,7 @@ sub _update {
         if ($dialog->ShowModal() == wxID_YES) {
             my $new_conf = Slic3r::Config->new;
             $new_conf->set("perimeters", 1);
+            $new_conf->set("min_shell_thickness", 0);
             $new_conf->set("top_solid_layers", 0);
             $new_conf->set("fill_density", 0);
             $new_conf->set("support_material", 0);
--- a/lib/Slic3r/GUI/PresetEditor.pm
+++ b/lib/Slic3r/GUI/PresetEditor.pm
@@ -760,11 +760,11 @@ sub _update {
     
     my $config = $self->{config};
     
-    if ($config->spiral_vase && !($config->perimeters == 1 && $config->min_shell_thickness <= $config->external_perimeter_extrusion_width && $config->top_solid_layers == 0 && $config->fill_density == 0 && $config->support_material == 0)) {
+    if ($config->spiral_vase && !($config->perimeters == 1 && $config->min_shell_thickness == 0  && $config->top_solid_layers == 0 && $config->fill_density == 0 && $config->support_material == 0)) {
         my $dialog = Wx::MessageDialog->new($self,
             "The Spiral Vase mode requires:\n"
             . "- one perimeter\n"
-            . "- a shell thickness no larger than the extrusion width\n"
+            . "- shell thickness to be 0\n"
             . "- no top solid layers\n"
             . "- 0% fill density\n"
             . "- no support material\n"
--- a/t/perimeters.t
+++ b/t/perimeters.t
@@ -1,4 +1,4 @@
-use Test::More tests => 59;
+use Test::More tests => 60;
 use strict;
 use warnings;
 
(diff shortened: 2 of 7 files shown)


xs/src/libslic3r/PerimeterGenerator.cpp
```diff
@@ -6,6 +6,23 @@
 
 namespace Slic3r {
 
+int
+PerimeterGenerator::num_loops(int perimeters, int extra_perimeters) {
+    int generated_perimeters = perimeters + extra_perimeters - 1;
+    long int min_shell_thickness, shell_thickness = 0;
```
lordofhyphens: ### Style
We don't use `long int`, just `long` (when we aren't using something that typedefs to `long`, like `coord_t`). 

Suggest `auto shell_thickness = this->ext_perimeter_flow.scaled_width()` (yay c++11).

xs/src/libslic3r/PrintConfig.cpp
```diff
@@ -775,6 +775,15 @@ PrintConfigDef::PrintConfigDef()
     def->max = 100;
     def->default_value = new ConfigOptionInt(35);
 
+    def = this->add("min_shell_thickness", coFloat);
+    def->label = "Minimum shell thickness";
+    def->category = "Layers and Perimeters";
+    def->sidetext = "mm";
+    def->tooltip = "Minimum shell thickness in mm";
+    def->cli = "min-shell-thickness=f";
```
lordofhyphens: ### Documentation
Update slic3r.pl and README.md to add your option to them.

xs/src/libslic3r/PerimeterGenerator.cpp
```diff
@@ -54,7 +71,7 @@ PerimeterGenerator::process()
     for (Surfaces::const_iterator surface = this->slices->surfaces.begin();
         surface != this->slices->surfaces.end(); ++surface) {
         // detect how many perimeters must be generated for this island
-        const int loop_number = this->config->perimeters + surface->extra_perimeters -1;  // 0-indexed loops
+        const int loop_number = num_loops(this->config->perimeters, surface->extra_perimeters);//this->config->perimeters + surface->extra_perimeters-1;  // 0-indexed loops
```
lordofhyphens: I like the way you extended this to increase the number of loops if necessary.

t/perimeters.t
```diff
@@ -315,6 +315,40 @@ use Slic3r::Test;
...
+        my ($self, $cmd, $args, $info) = @_;
+        
+        if ($info->{extruding} && $info->{dist_XY} > 0 && ($args->{F} // $self->F) == $config->perimeter_speed*60) {
+            $perimeters{$self->Z}++ if !$in_loop;
+            $in_loop = 1;
+        } else {
+            $in_loop = 0;
+        }
+    });
+    ok !(grep { $_ % $config->min_shell_thickness/$config->perimeter_extrusion_width } values %perimeters), 'should be 6 perimeters';
```
lordofhyphens: ### Robustness
Please add another couple test cases to check that it rounds up as expected if the perimeter width isn't a multiple of the min-width. Just in case @alexrj or I blunder in and break that condition somehow ;)

Take a look at the first test in t/support.t, it puts the actual test into a subroutine  and then calls that routine with different options. Here you'd just need to round up the division in the test itself to verify the number of perimeters that were generated.

lib/Slic3r/GUI/PresetEditor.pm
```diff
@@ -759,10 +760,11 @@ sub _update {
     
     my $config = $self->{config};
     
-    if ($config->spiral_vase && !($config->perimeters == 1 && $config->top_solid_layers == 0 && $config->fill_density == 0 && $config->support_material == 0)) {
+    if ($config->spiral_vase && !($config->perimeters == 1 && $config->min_shell_thickness == 0  && $config->top_solid_layers == 0 && $config->fill_density == 0 && $config->support_material == 0)) {
```
lordofhyphens: Good catch, I probably would have missed this myself.

curieos:
I just realized I forgot about the top and bottom of the model. This should also increase the amount of solid top/bottom layers yeah?

lordofhyphens:
I'm not so sure myself.

xs/src/libslic3r/PerimeterGenerator.cpp
```diff
@@ -6,6 +6,23 @@
 
 namespace Slic3r {
 
+int
+PerimeterGenerator::num_loops(int perimeters, int extra_perimeters)
+{
+    int generated_perimeters = perimeters + extra_perimeters - 1;
+    long min_shell_thickness = scale_(this->config->min_shell_thickness);
```
alexrj: Use `coord_t` instead of `long`

xs/src/libslic3r/PerimeterGenerator.cpp
alexrj: Just a note about semantics: `loop_number` holds the number of loops diminuished by 1. This is related to the implementation of the subsequent code, and I admit the variable could be named in a better way. But when you expose a `num_loops()` method, it should output the *actual* number of loops. Then you can subtract 1 in the caller method if it's needed by its implementation.

By the way, I don't see the need for a separate method just for this calculation. I'd embed the logic in the main method.

alexrj: By embedding it in the main method you can also cache `min_shell_thickness` and `this->ext_perimeter_flow.scaled_width()` without recalculating them every time.

xs/src/libslic3r/PerimeterGenerator.cpp
```diff
@@ -6,6 +6,23 @@
...
+    auto shell_thickness = this->ext_perimeter_flow.scaled_width();
+
+    shell_thickness += ((perimeters+extra_perimeters)-1) * this->perimeter_flow.scaled_width();
+    
+    if (shell_thickness < min_shell_thickness) {
+        shell_thickness = min_shell_thickness - this->ext_perimeter_flow.scaled_width();
+        generated_perimeters = ceil((float)shell_thickness/this->ext_perimeter_flow.scaled_width());
+    }
```
alexrj: Why calling `this->ext_perimeter_flow.scaled_width()` and `/this->ext_perimeter_flow.scaled_width()` twice?

By the way, I would simplify the logic:
1. Calculate how many loops are needed for the requested min_shell_thickness (considering one ext_perimeter followed by many normal perimeters, of course).
2. Increase loop number until min_shell_thickness is covered.

Also, write some comments.

alexrj: And you should just skip all of this math if min_shell_thickness is 0.

xs/src/libslic3r/PrintConfig.cpp
```diff
@@ -775,6 +775,15 @@ PrintConfigDef::PrintConfigDef()
     def->max = 100;
     def->default_value = new ConfigOptionInt(35);
 
+    def = this->add("min_shell_thickness", coFloat);
+    def->label = "Minimum shell thickness";
+    def->category = "Layers and Perimeters";
+    def->sidetext = "mm";
+    def->tooltip = "Minimum shell thickness in mm";
```
alexrj: This needs to be more clear. Minimum shell thickness is an alternative way for configuring perimeters and solid shells. If it's set to a non-zero value, they will be automatically increased until the configured value is reached.

alexrj:
Thank you for this. I added some comments to the code. Here's more:

* Solid layers (top/solid) shall be increased as well. Shell = vertical shell + horizontal shell.
* Option shall be added to the GUI (PresetEditor) as well.
* Option shall be added to the relevant place in `PrintRegion::invalidate_state_by_config()` as well.

xs/src/libslic3r/PerimeterGenerator.cpp
lordofhyphens: or `auto` (which is my preference as you are instantiating the value already). If @alexrj disagrees, go with what he said. ;)

curieos:
Thank you for the comments. I will get to these soon.

xs/src/libslic3r/PerimeterGenerator.cpp
curieos: loop_number is a constant. Any preference on how to set it? I thought a new method would be okay but you do bring up a good point.

xs/src/libslic3r/PerimeterGenerator.cpp
lordofhyphens: @alexrj what is the context  for `coord_t` ? What is the assumption/guarantee with using it?

curieos:
I made changes that should work. I'm running a build on my end real quick to make sure and I will then push the changes. I'll then get to working on the 3 other things you mentioned.

lordofhyphens:
We could break constness. Inlining the new code would make semantic sense
for a non-const number.

So would using another variable and assigning to the const loop_num

curieos:
I came up with a solution that I think you will like that I based on some of the semantics Alex talked about. Should be pushed in the next couple minutes.

curieos:
I forgot to change the tooltip. I will change it and push it with the next set of changes.

curieos:
I've been trying to find where slic3r creates the layer objects and determines whether or not they are solid and it seems like its in Object.pm. Does this sound right?

curieos:
The top and bottom solid layers implementation is in perl. I don't know much perl so I'm going to create another branch+pull request for a version of the method written in C++ before I finish this up.

foreachthing:
I hope this is no real issue! I get this, when Slic3r is started with `Slic3r-debug-console.exe`:
`Possible precedence problem on bitwise & operator at C:/Slic3r-master.2018.05.23.1703.f3b5909/lib/Slic3r/GUI/PresetEditor.pm line 806.`

Samir55:
@foreachthing I think it's because a missing '&' in line 806 
```Perl
$config->min_shell_thickness == 0 & $config->top_solid_layers == 0
# Instead of
$config->min_shell_thickness == 0 && $config->top_solid_layers == 0
```

PR #3958, opened 2017-05-17
Model.hpp Doxygen Documentation

Samir55:


xs/src/libslic3r/Model.hpp
```diff
@@ -28,221 +28,529 @@ typedef std::vector<ModelObject*> ModelObjectPtrs;
...
+    /// Invalidate the bounding box in the current ModelObject.
     void invalidate_bounding_box();
 
+    /// Repair all TriangleMesh objects found in each ModelVolume.
     void repair();
+
+    /// Flatten all volumes and instances into a single mesh and applying all the ModelInstances transformations.
     TriangleMesh mesh() const;
+
+    /// Flatten all volumes and instances into a single mesh.
```
alexrj: No! This just flattens all volumes *ignoring* instances.

xs/src/libslic3r/Model.hpp
```diff
@@ -28,221 +28,529 @@ typedef std::vector<ModelObject*> ModelObjectPtrs;
...
+    /// Get the raw bounding box
+    /// \return BoundingBoxf3
     BoundingBoxf3 raw_bounding_box() const;
+
+    /// Get the bounding box of the *transformed* given instance.
+    /// \param instance_idx size_t the index of the ModelInstance in the ModelInstance vector
+    /// \return BoundingBoxf3 the bounding box at the given index
     BoundingBoxf3 instance_bounding_box(size_t instance_idx) const;
+
+    /// Align the current ModelObject to ground by translating in the z axis the needed units.
```
alexrj: Specify whether this happens by changing the instances offsets or by translating volumes.

xs/src/libslic3r/Model.hpp
```diff
@@ -28,221 +28,529 @@ typedef std::vector<ModelObject*> ModelObjectPtrs;
...
     void align_to_ground();
+
+    /// Center the current ModelObject to origin.
```
alexrj: Specify whether this happens by changing the instances offsets or by translating volumes.

xs/src/libslic3r/Model.hpp
```diff
@@ -28,221 +28,529 @@ typedef std::vector<ModelObject*> ModelObjectPtrs;
...
     void center_around_origin();
+
+    /// Translate the current ModelObject with (x,y,z) units.
```
alexrj: Specify whether this happens by changing the instances offsets or by translating volumes.

xs/src/libslic3r/Model.hpp
```diff
@@ -28,221 +28,529 @@ typedef std::vector<ModelObject*> ModelObjectPtrs;
...
+    /// \param vector Vectorf3 the translation vector
     void translate(const Vectorf3 &vector);
+
+    /// Translate each TriangleMesh in every ModelVolume in this ModelObject.
```
alexrj: If this is just an alternative syntax for the other `translate()` method, it should have the same description. Different wording leads readers into thinking the methods actually behave differently.

xs/src/libslic3r/Model.hpp
```diff
@@ -28,221 +28,529 @@ typedef std::vector<ModelObject*> ModelObjectPtrs;
...
+    /// \param x coordf_t the x units
+    /// \param y coordf_t the y units
+    /// \param z coordf_t the z units
     void translate(coordf_t x, coordf_t y, coordf_t z);
+
+    /// Scale the current ModelObject
```
alexrj: Specify whether this happens by scaling the volume meshes or by altering the scaling factor of instances.

xs/src/libslic3r/Model.hpp
```diff
@@ -28,221 +28,529 @@ typedef std::vector<ModelObject*> ModelObjectPtrs;
...
+    /// \param factor float the scaling factor
     void scale(float factor);
+
+    /// Scale each TriangleMesh in every ModelVolume in this ModelObject.
```
alexrj: Same as above: it's different syntax for the same method. Use the same wording for the two.

xs/src/libslic3r/Model.hpp
```diff
@@ -28,221 +28,529 @@ typedef std::vector<ModelObject*> ModelObjectPtrs;
...
+    /// \param versor Pointf3 the scaling factor in a 3d vector.
     void scale(const Pointf3 &versor);
+
+    /// Scale the current ModelObject to fit.
```
alexrj: Specify whether this happens by scaling the volume meshes or by altering the scaling factor of instances. It's important to clarify this because it affects also the semantics of input values (i.e. should users specify a size relative to the raw volume meshes or to the instance-transformed meshes?)

alexrj: Since it's of course the second case, we should also mention that it operates on the total size computed by duplicating the object according to all the instances.

xs/src/libslic3r/Model.hpp
```diff
@@ -28,221 +28,529 @@ typedef std::vector<ModelObject*> ModelObjectPtrs;
...
+    /// \param size Sizef3 the size vector
     void scale_to_fit(const Sizef3 &size);
+
+    /// Rotate the current ModelObject.
```
alexrj: Specify whether this rotates the volume meshes or just alters the instance attributes.

xs/src/libslic3r/Model.hpp
```diff
@@ -28,221 +28,529 @@ typedef std::vector<ModelObject*> ModelObjectPtrs;
...
+    /// \param angle float the angle in radians
+    /// \param axis Axis the axis to be rotated around
     void rotate(float angle, const Axis &axis);
+
+    /// Mirror the current Model around a certain axis.
+    /// \param axis Axis enum member
     void mirror(const Axis &axis);
+
+    /// Transform the current ModelObject by a certain ModelInstance attributes.
```
alexrj: It's important to document that the inverse transformation is applied to all the object instances, so that the final size/position/rotation of the transformed objects doesn't change. This method is useful for flattening the instance transformations (of one instance) to the mesh.

xs/src/libslic3r/Model.hpp
```diff
@@ -28,221 +28,529 @@ typedef std::vector<ModelObject*> ModelObjectPtrs;
...
+
+    /// Get the number of the unique ModelMaterial objects in this ModelObject.
+    /// \return size_t the materials count
     size_t materials_count() const;
+
+    /// Get the number of the facets found in all ModelVolume objects in this ModelObject which are not modifier volumes.
+    /// \return size_t the facets count
     size_t facets_count() const;
+
+    /// Know whether there exists a TriangleMesh object that needs repair or not.
```
alexrj: ...that *needed* repair.

xs/src/libslic3r/Model.hpp
```diff
@@ -28,221 +28,529 @@ typedef std::vector<ModelObject*> ModelObjectPtrs;
...
+    /// \return bool
     bool needed_repair() const;
+
+    /// Cut (Slice) the current ModelObject at a certain axis at a certain magnitude.
```
alexrj: Just English notes: *along* a certain axis at a certain _coordinate_.

xs/src/libslic3r/Model.hpp
```diff
@@ -28,221 +28,529 @@ typedef std::vector<ModelObject*> ModelObjectPtrs;
...
+    /// \param axis Axis the axis to slice at (X = 0 or Y or Z)
+    /// \param z coordf_t the point at the certain axis to cut(slice) the Model at
+    /// \param model the owner Model
```
alexrj: This is the Model which will get the resulting objects added. It can be the same model as the input one, or another one.

lordofhyphens:
As a FYI, Markdown syntax is supported with Doxygen. Don't feel as if you have to use it.

xs/src/libslic3r/Model.cpp
```diff
@@ -212,16 +212,6 @@ Model::center_instances_around_point(const Pointf &point)
...
-Model::align_instances_to_origin()
-{
-    BoundingBoxf3 bb = this->bounding_box();
-    
-    Pointf new_center = (Pointf)bb.size();
-    new_center.translate(-new_center.x/2, -new_center.y/2);
-    this->center_instances_around_point(new_center);
-}
-
-void
```
Samir55: @alexrj told me to remove it from Model.

PR #4023, opened 2017-06-12
Feature: sequential printing without completing objects

### LINKED ISSUE #3739: Feature request: sequential printing without completing objects
This is a feature present in Simplify3D, described here: https://www.simplify3d.com/support/articles/multi-part-printing/ (Multiple Process, Sequential Printing Mode)

### Version
I am currently using 1.3.0-dev 1.2.9-627-g7d2884b

Ubuntu 16.04 64 bits

### Behavior
When using sequential printing, if objects are taller than the heigh clearance slic3r simply refuses to slice. But it should not be this way. It could behave like Simplify3D: since we are limited by this height, do a sequential printing for all objects up to a height of <height clearance>, then start over again at <height clearance+1> and do it sequentially again, then again at <height clearance x 2 + 1> and so on.

YES

Note: I did not include my configuration files because they aren't really needed, since this is a feature request. But I can of course provide them if deemed necessary.

slic3r_827

fivethreeo:
Fixes #3739

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: perl
 before_install:
 - sh package/linux/travis-decrypt-key
 install:
-- export LDLOADLIBS=-lstdc++
 - export BOOST_DIR=$HOME/boost_1_63_0
 - export SLIC3R_STATIC=1
 - export CXX=g++-4.9
--- a/package/deploy/sftp.sh
+++ b/package/deploy/sftp.sh
@@ -19,8 +19,16 @@ fi
 if [ -s $KEY ]; then
     for i in $FILES; do 
          filepath=$(readlink -f "$i")
-         echo put $filepath | sftp -i$KEY "${UPLOAD_USER}@dl.slic3r.org:$DIR/"
+         tmpfile=$(mktemp)
+         echo put $filepath > $tmpfile 
+         sftp -b $tmpfile -i$KEY "${UPLOAD_USER}@dl.slic3r.org:$DIR/"
+         result=$?
+         if [ $? -eq 1 ]; then 
+             echo "Error with SFTP"
+             exit $result; 
+         fi
     done
 else
     echo "$KEY is not available, not deploying." 
 fi
+exit $result
--- a/xs/Build.PL
+++ b/xs/Build.PL
@@ -10,6 +10,7 @@ use Module::Build::WithXSpp;
 
 my $cpp_guess = ExtUtils::CppGuess->new;
 my $mswin = $^O eq 'MSWin32';
+my $linux = $^O eq 'linux';
 
 # prevent an annoying concatenation warning by Devel::CheckLib
 $ENV{LD_RUN_PATH} //= "";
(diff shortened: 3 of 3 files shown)


fivethreeo:
Ref: #3739

fivethreeo:
Cannot find out why tests fail, needs more eyes.

fivethreeo:
All tests successful.

xs/src/libslic3r/Print.cpp
```diff
@@ -656,22 +656,6 @@ Print::validate() const
...
-                object_height.insert(object_height.end(), object->copies().size(), object->size.z);
-            }
-            std::sort(object_height.begin(), object_height.end());
-            // ignore the tallest *copy* (this is why we repeat height for all of them):
-            // it will be printed as last one so its height doesn't matter
-            object_height.pop_back();
-            if (!object_height.empty() && object_height.back() > scale_(this->config.extruder_clearance_height.value))
-                return "Some objects are too tall and cannot be printed without extruder collisions.";
-        }
-        */
```
lordofhyphens: Isn't this still generally necessary?

fivethreeo: Only if we add a extra option to choose between completely print individual objects sequentially  and partially print objects per clearance height sequentially. But that makes little sense since the object is too tall anyway. Another option would be to make it a warning that says "Some objects are higher than extruder clearance height. Objects will be be printed in stages, still sequentially but now per clearance height."

fivethreeo:
Should between objects gcode be added per clearance as well?

xs/src/libslic3r/Print.cpp
lordofhyphens: We should leave a warning in place that the fall-through logic is occurring. Operating on a principle of least surprises.

lordofhyphens:
We'd have to reconsider the semantics for between-objects gcode if we did that, and I'm unsure whether or not anyone can actually appropriately grok that in a static sense. Maybe with custom conditional gcode, but I think it's better to leave between-object gcode as between full objects (that is, this partial mode is not in effect). Thus we should also probably add a note to the tooltip for complete objects sequentially that notes this behavior.

lordofhyphens:
@Patola any comments? You're the original requester of the feature and as such probably have a decent grip of the use case.

lordofhyphens:
https://ci.appveyor.com/api/buildjobs/bxxdtkeccm3rm5nl/artifacts/Slic3r-master.2017.06.13.1196.0864140.64bit.zip

xs/src/libslic3r/Print.cpp
fivethreeo: Yes, but where would I add it as just a warning, not a fatal error as in validate?

lordofhyphens: Good question; I haven't actually looked around in the source for a notification from the xs side.

lordofhyphens:
@fivethreeo Could you make the tweaks to the validation and see if #4043 is affected by your work?

fivethreeo:
Will not be affected by my work.

lordofhyphens:
Thanks.

fivethreeo:
Made this feature a option in the printconfig. Adding link to artifact so it can be tested.

https://ci.appveyor.com/api/buildjobs/7fhbctsh0ijxotba/artifacts/Slic3r-master.2017.06.26.1257.93268f0.64bit.zip

lordofhyphens:
Could use some automatic tests if they haven't been done yet.

fivethreeo:
Yes, but what to test for? Validation, if so how?  If height is visited (n) times per clearance without going below previous clearance or something?

lordofhyphens:
Look at the functionality. You can plate arbitrary objects in the test
environment.

Check for the conditions that would trigger the new logic.

Issue #4022, opened 2017-06-11
feature request: generic alternative to octoprint print server


rob-miller:
### Operating system OS X El Capitan 10.11.6 (old iMac)
I used to use OctoPrint/OctoPi, but now I have a DuetWifi.  I wish I could still use the 'send to printer' button.  Duet supports a different POST method to upload files (see [DuetWebControl ReadMe](https://github.com/chrishamm/DuetWebControl/blob/master/README.md) and/or [forum posting](https://www.duet3d.com/forum/thread.php?pid=9032#p9032)) and apparently ftp although not certain ftp works across all versions.  In any event it is likely I could come up with a command line to implement the transfer.  As there will be other print servers to support with 'send to printer' a more generic option is desirable.

I think this could be implemented by adding a text box for the transfer command next to the OctoPrint API key box, and if this box is filled then use it in preference to the inbuilt OctoPrint transfer command.  More features like the 'overwrite existing file' and 'start print' would be great too, possibly tied to symbols in the command line, but that is outside the core functionality (a button to execute an external command with the output G-code).

Yes I could implement this as others have done by having a script watching the 'export G-code' output directory, but that still leaves me with extra dialogues for file name and overwriting existing files, and depending on implementation I would need to change more things when I actually want to look at the G-code before printing.

Apologies if I have missed some existing way to implement this.

lordofhyphens:
Going to consider this closed for now with the duet commit; I don't see a need to architect some master "use this on your firmware to enable print from Slic3r" at this time; we can refactor later as needed.

PR #4028, opened 2017-06-14
Octoprint settings functionality expanded to allow upload to RepRapFi…

emaker:
…rmware on a Duet controller.
Some way towards solving #4022 although I don't know about any other print servers at this time.

### DIFF
--- a/lib/Slic3r/GUI/Plater.pm
+++ b/lib/Slic3r/GUI/Plater.pm
@@ -49,7 +49,7 @@ sub new {
     my $self = $class->SUPER::new($parent, -1, wxDefaultPosition, wxDefaultSize, wxTAB_TRAVERSAL);
     $self->{config} = Slic3r::Config->new_from_defaults(qw(
         bed_shape complete_objects extruder_clearance_radius skirts skirt_distance brim_width
-        serial_port serial_speed host_type octoprint_host octoprint_apikey shortcuts filament_colour
+        serial_port serial_speed host_type print_host octoprint_apikey shortcuts filament_colour
     ));
     $self->{model} = Slic3r::Model->new;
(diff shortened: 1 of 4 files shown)


emaker:
I wasn't sure how to implement the aliases you mentioned, so I added octoprint_host as a legacy option in PrintConfig.cpp, using the Bed_Size option as a guide.

lordofhyphens:
Looks good, thanks for the commit. This looks generic enough that if someone wants to continue to add support for another host program that doesn't need more than what is provided we're still good to go.

lordofhyphens:
If it needs a refactor for the next one, then we'll deal with it then.

rob-miller:
This is great to have (!!!) 
But on 1.18.1 DuetWifi Slic3r automated build slic3r-1.3.0-dev-98fca90.dmg I find:
- if the file does not exist on the SD card already, Duet reports 
'GCode file "lsm1.gcode" not found
Failed to open file lsm1.gcode'
(or whatever file may be called)
- if a file with the same name exists, the existing file is queued for printing (but not the new one sent from slic3r)

this is with 'Overwrite existing file' and 'Start print' enabled.  Have tried other settings, all consistent with not getting file loaded on to SD card.

emaker:
@rob-miller I don't have a DuetWifi to play with, but this code was developed using a Duet 0.6 over ethernet, and with @chrishamm's firmware 1.18-b2. I didn't test with 1.19 as it didn't appear to be stable yet.

rob-miller:
Thanks, saw that on the duet forum.  Surprised the behaviour is different on mine - initially thought it worked until realised the file was the old one.  happy to debug this offline (email in my gh profile)

PR #4039, opened 2017-06-22
Produce AppImage from the leavings of the normal bz2.

lordofhyphens:
Experimental build for Linux AppImages.


lordofhyphens:
Well, it works on my system; I have other refactors that depend on this and I don't want to make a mess of the commit history so here it goes.

Issue #3934, opened 2017-05-04
[GSoC] Samir55 / 3MF file format


alexrj:
This issue is dedicated to the GSoC student @Samir55 who proposed to implement support for the 3MF file format in Slic3r:
https://summerofcode.withgoogle.com/projects/#6343944127381504

This topic was discussed in this original issue:
#2811

@Samir55 already contributed to the Slic3r codebase by adding zoom commands (pull request #3864, merged).

@Samir55, welcome to Slic3r! Would you please introduce yourself to the Slic3r community? Please mention any experience with 3D printing or similar technologies, and include your time zone and operating system.

Samir55:
I'm Samir, a computer engineering undergraduate student, I have some experience in CNC machines and G-code as I am working with my team this semester to make a machine like CNC machine (but we are using a touch screen stylus instead which draws any picture we send to it) . 3D printers are not so popular yet in our country :) so, I have little experience with 3D printers. I dealt with 3D model files (like: OBJ ) in the computer graphics game project .

Time Zone: Eastern European Time Zone (UTC+02:00)

Operating System : Ubuntu 14.04 and Mac OS Sierra .  
I am trying now to build Slic3r on mac OS.

Samir55:
Today's Task: 
* Studying different Formats ( AMF, OBJ, STL )
* Listing the differences between them and the information each carries. 
* Putting the generated document in the Slic3r Wiki
```NP: this task may not be finished today :)  ```

**Comment by NateTG** [comment]
If there are specs or references about the formats that you used, and those are available on-line, be sure to include links in the wiki.

**Comment by Samir55** [comment]
I have made a draft about what file formats Slic3r support document. [This is the link](https://docs.google.com/document/d/1ciqo8SBUPJGTuaqNNrQ1m0rTcV6APSY2l7YgLVXIjmw/edit?usp=sharing). I have some questions found in the comments. any help is really appreciated :)

**Comment by Samir55** [comment]
This Week Task (From 8-5):
* Improve and Finalize the wiki draft  [in this link](https://docs.google.com/document/d/1ciqo8SBUPJGTuaqNNrQ1m0rTcV6APSY2l7YgLVXIjmw/edit?usp=sharing)  (Still, needs feedback).
* Study Slic3r::IO.
* Study Slic3r::Model.
* See how to improve the proposal.

**Comment by Samir55** [comment]
Question #1: 
```
Does Slic3r use the loaded materials from OBJ files?
[...]

PR #4046, opened 2017-06-24
[GSoC 2017] Support 3MF Format Read/Write.

Samir55:
3MF is generally smaller than other comparable 3D formats where transforms and object references are supported. 3MF has an advantage over the STL format since STL doesn’t contain color or material/property and also it doesn’t have mesh topology which in turn results in larger file size.
**_This topic was discussed in this original issue: #2811_**

## Features:
* Import/Export to 3MF files command line options.
* Export plate with modifiers and print configs in .3MF (GUI).
* Export selected objects with modifiers print configs in .3MF (GUI).

**_The test suite is found in 23_3mf.t in xs/t/._**
**_The code is documented using Doxygen syntax._**

## Related Links:
* **The GSoC project blog (A dedicated issue)**: #3934 
* **[3MF Document Specification](https://github.com/alexrj/Slic3r/wiki/3MF-Document-Core-Specification) wiki page I've written summarizing the 3MF format core specification.**
* **[Project Plan](https://github.com/Samir55/Slic3r/wiki/Project-Plan) containing Deliverables and what Slic3r should support.**

## Other Pull Requests:
* Undo/Redo to rotate, miror, split, cut, etc (PR [# 4100](https://github.com/alexrj/Slic3r/pull/4100), Still working on it)
* Documentation for Slic3r::Model. [PR # 3958 (Merged)](https://github.com/alexrj/Slic3r/pull/3958)


slic3r.pl
```diff
@@ -6,7 +6,7 @@
 BEGIN {
     use FindBin;
     use lib "$FindBin::Bin/lib";
-    use local::lib '--no-create', "$FindBin::Bin/local-lib";
```
lordofhyphens: Did this file actually change? If not, please commit a reversion (probably a manual one) and leave your local patch uncommitted.

src/slic3r.cpp
```diff
@@ -180,4 +187,4 @@ main(int argc, char **argv)
     }
     
     return 0;
-}
+}
```
lordofhyphens: There should be a newline here.

xs/src/libslic3r/IO/TMF.cpp
```diff
@@ -0,0 +1,1066 @@
+#include "TMF.hpp"
+
+namespace Slic3r { namespace IO {
+
+TMFEditor::TMFEditor(std::string input_file, Model *model)
```
lordofhyphens: I recommend an initialize list here (also review for anything that would be better to have as a constant), instead of a constructor like this. You can stick it in the hpp as well if it's just an initialize list.

slic3r.pl
Samir55: @lordofhyphens  Sorry, this file I always forget to revert it back :) . I won't leave it till the end. I do that change when I code on ubuntu 14.04. I tried to update the perl modules but I couldn't.

xs/src/libslic3r/IO/TMF.hpp
```diff
@@ -0,0 +1,246 @@
...
+        s << number;
+        return s.str();
+    }
+
+};
+
+/// 3MF XML Document parser.
+struct TMFParserContext{
+
+    enum TMFNodeType {
```
lordofhyphens: C++11 note: use scoped enums where possible (enum class); this will scope them appropriately. 

Suggest pushing it up outside the struct (as it's public already); this is not a hard requirement (and subject to additional comment by @alexrj )

xs/src/libslic3r/Model.cpp
```diff
@@ -20,6 +20,9 @@ Model::Model(const Model &other)
     this->objects.reserve(other.objects.size());
     for (ModelObjectPtrs::const_iterator i = other.objects.begin(); i != other.objects.end(); ++i)
         this->add_object(**i, true);
+
+    // copy metadata
+    this->metadata = other.metadata;
```
lordofhyphens: Where is the initialize for metadata and the part number? I don't see a commit in Model.

lordofhyphens:
I like the comments that have been written so far in the headers.

xs/src/libslic3r/Model.cpp
lordofhyphens: The reasoning is that it should be initialized to something for a non-3mf load (I don't know if you tested this).

xs/t/23_3mf.t
```diff
@@ -0,0 +1,210 @@
...
+    is( $content_types_output, $expected_content_types, "[Content_Types].xml file matching");
+    # Check contents in _rels.xml.
+    my $relationships_output ;
+    unzip $tmf_output_file => \$relationships_output, Name => "_rels/.rels"
+        or die "unzip failed: $UnzipError\n";
+    is( $relationships_output, $expected_relationships, "_rels/.rels file matching");
+}
+
+# Finish finish test cases.
+done_testing();
```
lordofhyphens: A convert/from STL would be useful.

Samir55: @lordofhyphens  I will be adding them :) .

xs/src/libslic3r/Model.cpp
Samir55: @lordofhyphens  The metadata is a map. If it's empty I do nothing.
The part number is specific to the model object and I initialize it by -1 if it's not found.

lordofhyphens: Please make sure all data members/structures are initialized; this is good practice.

Samir55:
@alexrj @lordofhyphens I think it's ready to start merging if you are satisfied with current zip implementation. if not, I will implement it :) .

lordofhyphens:
I would like you to go through the code once and update to take advantage
of constexpr, auto variables, etc.

Basically a beauty pass over it.

Samir55:
@alexrj, @lordofhyphens I have updated the code according to the link [you have provided in IRC](https://hownot2code.com/2016/09/15/how-to-avoid-bugs-using-modern-c/) and I have used also -Weffc++ and -Wextra flags to see warnings. There were only warnings about TMFEditor and TMFParserContext not having their override versions of copy constructor and =operator overloading. I will implement them when I wake up, I don't see any need for them in our case, however.

Samir55:
@alexrj  @lordofhyphens  If there are any improvements I should do please tell me :)

lordofhyphens:
Not in MANIFEST: t/23_3mf.t
Not in MANIFEST: t/models/3mf/box.3mf
Not in MANIFEST: t/models/3mf/chess.3mf
Not in MANIFEST: t/models/3mf/FaceColors.3mf
Not in MANIFEST: t/models/3mf/gimblekeychain.3mf
Not in MANIFEST: t/models/3mf/SplitPyramid.3mf
Not in MANIFEST: t/models/amf/FaceColors.amf.xml
Not in MANIFEST: t/models/stl/gimblekeychainExtended.stl
Not in MANIFEST: t/models/stl/spikey_top.stl
MANIFEST appears to be out of sync with the distribution

Samir55:
I've added the test files to the MANIFEST But these files should be removed from the repo (they are no longer used)
```
t/models/3mf/FaceColors.3mf
t/models/3mf/SplitPyramid.3mf
t/models/stl/gimblekeychainExtended.stl
```
@lordofhyphens sorry I forgot to add them :)

lordofhyphens:
@Samir55 Then you should `git rm` everything that shouldn't be in the repo and remove from MANIFEST.

Samir55:
* t/models/3mf/FaceColors.3mf It was removed using unlink() function I've added in [this commit](https://github.com/alexrj/Slic3r/pull/4046/commits/92d7ba12d9484cde86063217151bd0d048962ed7) . It is generated only when the test is run.
* t/models/3mf/SplitPyramid.3mf this file was generated during a test I deleted later at [this commit ](https://github.com/alexrj/Slic3r/pull/4046/commits/cdef8943694860349a3525d855c100fee182d35a).
* t/models/stl/gimblekeychainExtended.stl was a file generated also by 3mf test (conversion from gimblekeychainExtended.3mf to STL) and I've removed that file [in that commit](https://github.com/alexrj/Slic3r/pull/4046/commits/dacb426c0b52583e6e0fcb328f68b5f77e3134cc) 
@lordofhyphens I mean they are only on your branch.* t/models/3mf/SplitPyramid.3mf this file was generated during a test I deleted later at [this commit ](https://github.com/alexrj/Slic3r/pull/4046/commits/cdef8943694860349a3525d855c100fee182d35a).
 
@lordofhyphens I mean they are only on your branch.

lordofhyphens:
Well, that would do it ;)

Issue #4045, opened 2017-06-24
 3D Honeycomb infill -- update to switch between horizontal/vertical at squares


gringer:
[I hope it's okay for me to cross-post this to both Cura and Slic3r at the same time]

I've been working on an optimisation of the 3D Honeycomb code after experimenting with creating a large version of the honeycomb structure on my 3D printer (which hopefully will print quicker with fewer bugs). If the print alternates between horizontal and vertical at each square, it is able to print a fully-supported honeycomb structure with half as much filament and only sweeping through one direction. Here's the R code that I've used to test/debug my code for this updated method. It takes as input two variables: `el` - the length of the edge of the squares (which is the same as the bridging distance), and `zi` - the current height of the layer. It currently only prints 4 rows/columns, but it's easy enough to modify the _xofs_ `for` loop to sweep through the width of a print, incrementing by `2*el` each time:

```
#!/usr/bin/Rscript
library(rgl);

el <- 5; ## edge length
ch <- 4*el / sqrt(2); ## cycle height
pl <- el*4; ## path length

frame <- 0;
plot3d(NA, xlim=c(-10,40), ylim=c(-10,40), zlim=c(-10,40),
       box=FALSE);
for(zi in 0:80/3){
    ##for(zi in c(16,32) / 2){
    ## sawtooth wave function
    dl <- (abs(((zi*4) %%  (4*ch)) - (2*ch)) - ch) / 2; ## diagonal length
    dc <- floor(((zi*2) %% (2*ch)) / ch); ## diagonal cycle
    dox <- dl / sqrt(2); ## diagonal x offset
    doy <- abs(dox); ## diagonal y offset
    ol <- (pl - 2*doy)/2; ## orthogonal length
    for(xofs in c(0,el*2,el*4,el*6)){
        if(dc == 0){
            lines3d(x=rep(c(dox/2,dox/2,-dox/2,-dox/2,dox/2),2)+xofs,
                    y=rep(c(0, ol/2, ol/2+doy, ol*1.5+doy, ol*1.5+doy*2),2) +
                        rep(0:1*pl,each=5),
                    z=rep(zi,10), col=rainbow(40)[zi+1]);
        } else {
            lines3d(y=rep(c(dox/2,dox/2,-dox/2,-dox/2,dox/2),2)+xofs+el,
                    x=rep(c(0, ol/2, ol/2+doy, ol*1.5+doy, ol*1.5+doy*2),2) +
[...]

PR #4049, opened 2017-06-26
Update 3D Honeycomb Infill to switch directions at squares rather than every line

gringer:
If the print alternates between horizontal and vertical at each square, it is able to print a fully-supported honeycomb structure with half as much filament and only sweeping through one direction. I was mostly there with the initial implementation of this code, but I needed to do some large octahedron printing myself to work out a better way to do it.

This modification reduces the need for bridging in the infill pattern (it only happens at the squares at the top and bottom of the print), which increases strength and reduces fill density.

see issue #4045 for more details.

FWIW, Slic3r has been much easier to hack than Cura2. Thanks for writing code that will compile somewhat easily on a Debian system.

### DIFF
--- a/xs/src/libslic3r/Fill/Fill3DHoneycomb.cpp
+++ b/xs/src/libslic3r/Fill/Fill3DHoneycomb.cpp
@@ -82,51 +82,45 @@ zip(const std::vector<coordf_t> &x, const std::vector<coordf_t> &y)
 }
 
 // Generate a set of curves (array of array of 2d points) that describe a
-// horizontal slice of a truncated regular octahedron with edge length `gridSize`.
+// horizontal slice of a truncated regular octahedron with edge length 1.
 // curveType specifies which lines to print, 1 for vertical lines
 // (columns), 2 for horizontal lines (rows), and 3 for both.
 static std::vector<Pointfs>
-makeNormalisedGrid(coordf_t z, coord_t gridSize, size_t gridWidth, size_t gridHeight){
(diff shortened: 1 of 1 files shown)


lordofhyphens:
I do my development work on Debian Jessie ;)

lordofhyphens:
https://ci.appveyor.com/api/buildjobs/nrry2oeviolc396q/artifacts/Slic3r-master.2017.06.26.1250.aec1c3c.64bit.zip

Windows build of this PR.

gringer:
This infill pattern seems to work better when I disable the bridge mode tweak and slightly reduce the infill speed. Otherwise, layers can pull the filament over from the layers underneath, creating a weaker (and less pretty) structure. Here's what it should look like:

https://twitter.com/gringene_bio/status/879932267954528256

gringer:
For people interested in trying out this on their 3D printers before downloading/compiling, I've got a couple of Gcode models including this fill pattern on my Truncated Octahedron thingiverse page:

https://www.thingiverse.com/thing:2404685/

One model is a 50x50x25mm box with no side or top walls, which has proven to be fairly resistant to crushing (particularly in the centre).

The other model is a 40mm truncated octahedron with walls, so that the internal structure can be seen without having to think too hard.

lordofhyphens:
On the windows build server there is an artifacts tab you can download
builds from

lordofhyphens:
Thank you!

lordofhyphens:
Also, sorry for taking forever to merge this.

PR #4083, opened 2017-08-11
Fix acceleration gcode for RepRapFirmware

cangelis:
This PR fixes acceleration gcode for `reprap` flavor as the documentation stated http://reprap.org/wiki/G-code#M204:_Set_default_acceleration & https://duet3d.com/wiki/G-code#M201:_Set_max_acceleration

### DIFF
--- a/xs/src/libslic3r/GCodeWriter.cpp
+++ b/xs/src/libslic3r/GCodeWriter.cpp
@@ -215,11 +215,15 @@ GCodeWriter::set_acceleration(unsigned int acceleration)
     this->_last_acceleration = acceleration;
     
     std::ostringstream gcode;
-    if (FLAVOR_IS(gcfRepetier)) {
+    if (FLAVOR_IS(gcfRepetier) || (FLAVOR_IS(gcfRepRap))) {
         gcode << "M201 X" << acceleration << " Y" << acceleration;
         if (this->config.gcode_comments) gcode << " ; adjust acceleration";
         gcode << "\n";
+    }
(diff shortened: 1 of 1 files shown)


lordofhyphens:
Ugh, when did this change, and for which firmwares did this change?

RepRap flavor covers a lot of ground.

cangelis:
Not sure about the date but this is how RepRapFirmware works and reprap documentation says the same. Please see the links I provided.

lordofhyphens:
Looks like Marlin supports the P/T configuration (and an addititional R for retract accel). 
It's *probably* safe to merge, assuming we don't mind breaking older Marlin firmware builds (according to docs, pre-2015). 

@alexrj Your thoughts on this?

lordofhyphens:
Example: Sprinter afaik only supports the S notation (which the older one), as does older Marlin versions. Is there a non-dev build of Marlin that has this in it? 

RepRap flavor in Slic3r meant the common firmwares associated with the Reprap project (originally this was Repetier, Sprinter, RepRapFirmware, and Marlin if I recall correctly), also theoretically teacup and Smoothie.

cangelis:
Is Sprinter still maintained? Correct me if I am wrong, last commit was 4 years old and there are issues and PRs from 2012. For Marlin, I think 2.5 years is enough time for BC break. If we still would like to support old version reprap firmwares we need to identify a new flavor (eg: `reprap-old` which is not necessary I think. Currently, slic3r does not work correctly with new versions.

PR #4086, opened 2017-08-14
New configuration for support material interface extrusion width

cangelis:
Currently, slic3r uses `support_material_extrusion_width` for interfaces. Assigning very low values to `support_material_extrusion_width` makes supports very easy to break which is very a good feature. But when `support_material_extrusion_width` is very low, interfaces don't stick to neither the lower support layers nor the adjacent extrusion lines.

So what I do is to define a new configuration parameter for support material interfaces, and I can still use very low values `support_material_extrusion_width` and higher values for support material interfaces for reliable extrusion. For example:

For `nozzle_diameter = 0.4`;
`support_material_extrusion_width = 0.4` makes the supports very thin and easy to break and
`support_material_interface_extrusion_width = 0.5` makes reliable and accurate extrusion lines for interfaces.

PR does not create any BC break, if the `support_material_interface_extrusion_width` is not defined, `support_material_extrusion_width` or `extrusion_width` will be used as the old behaviour.


lordofhyphens:
Would appreciate a test with this.

PR #4088, opened 2017-08-17
Use constant speed for too short displacements

cangelis:
I have been testing time estimation for about a week, it works pretty accurate with small models but the estimation for big and rounded models are very far (sometimes for days) from the expected.

Because generally the rounded models (means lots of gcode for a single line) are inaccurate, it looks like short displacement time calculation does not work well. So I considered constant speed for too short displacements and the results are really promising.

Time estimation is much more complex than it seems but this PR improves it a little bit.


lordofhyphens:
Your reasoning looks sound to me. And yes, time estimation is really complex.

PR #4100, opened 2017-08-28
Undo/Redo to rotate, miror, split, cut, etc

### LINKED ISSUE #3265: Feature Request: undo to rotate, flip, split, cut
Sometimes I hit the wrong button/menu entry or more often give the wrong angle on rotation and then have to figure out what the reverse action is to undo. Or delete and re-add the object. It would be ncie if there were an undo button for this.

Samir55:
Fixes #3265

Undo/Redo to:
- [x] Add Models.
- [x] Delete.
- [x] Delete All.
- [x] Increase.
- [x] Decrease.
- [x] Set Number of Copies.
- [x] Rotate.
- [x] Scale.
- [x] Split.
- [x] Cut.


Samir55:
@lordofhyphens  Is it useful to add undo/redo to open(load) models?

lordofhyphens:
I don't think so, if you are talking about undo/redo add and delete operations. The "redo" operation for Delete is Add. ;)

 Moreover, I'd see that as potentially dangerous (what happens if the file is on a network share and disappears between an undo/redo)?

Samir55:
@lordofhyphens Is undo/redo to x, y translations needed? I don't know to handle it, otherwise, the PR is ready for review.

lordofhyphens:
I don't think undo/redo to translations is particularly necessary myself. 

If someone complains about it (that is, opens an issue), it can be dealt with then. 

As for handling it, wouldn't it be sufficient to record a series of coordinates of the center of the bounding box and move there?

lordofhyphens:
An overview of the source indicates that it looks fine; how much have you tried to break it?

Samir55:
* I tried every single operation then undo/redo.
* Then tried a combination of destructive and non destructive operations. and I have fixed any issues that occurred.
I tried 7+ hours in the previous days. It would be very helpful if you tried a little cases.

@lordofhyphens There is something weird happens if I applied undo/redo very fast; Segmentation faults occur. Here is one of them 
```Console
Thread 78 terminated abnormally: Caught C++ exception of type or derived from 'std::exception': vector at /Users/ASamir/GSoC/Slic3rFork/Slic3r/lib/Slic3r/GUI/Plater.pm line 2950.
Segmentation fault: 11
```
and here is another one
```Console
libc++abi.dylib: terminating with uncaught exception of type std::length_error: vector
Abort trap: 6
```

**But It doesn't seem it's not from the undo/redo**. I've tried the **master branch** and added 3 extra rotate calls when rotate button is pressed and segmentation fault also occurs.
```Perl
# Added at line 2406 in object_menu sub routine in Plater.pm
wxTheApp->append_menu_item($rotateMenu, "Around X axis…", 'Rotate the selected object by an arbitrary angle around X axis', sub {
            #$self->rotate(undef, X);
            $self->rotate(0.0174533, X);
            $self->rotate(-0.0174533, X);
            $self->rotate(0.0174533, X);
            $self->rotate(-0.0174533, X);
  }
```

This is one of the models that  might cause segmentation fault 
[magnolia.obj.zip](https://github.com/alexrj/Slic3r/files/1294402/magnolia.obj.zip)

Do yo have insights about what's going on?

lordofhyphens:
Maybe there's some race condition going on? That'd be my first thought on it.

lordofhyphens:
Go ahead and create a github issue for the segfault.

lordofhyphens:
Merging. Nice work.

Issue #4106, opened 2017-09-02
Firmware Retract G10 S1 BEFORE TOOL CHANGE PLEASE


studiodyne:
🔵🔵🔵🔵🔵🔵🔵🔵🔵🔵🔵🔵🔵🔵🔵🔵
 Works now !  Read the end of topic
[Dual Y Single nozzle direct drive works on MARLIN with FW enabled  '0.6mm height](https://www.youtube.com/watch?v=n3wsoSrxNNM)
🔵🔵🔵🔵🔵🔵🔵🔵🔵🔵🔵🔵🔵🔵🔵🔵

**No G10 S1 before ToolChange**

**`Firmware Retractation not ready for Multi Extrusion printing`**

Firmware retraction when ' selected ' , forget to make ' tool change retractation in multi extruder printing , but works fine when only one extruder because no need of tool change long retract. 
Please implement  G10 S1 just before a toolchange , to make the long retractation , if a dual nozzle the print is still possible but dirty with missing perimeters because of not primed extruder , but if SINGLENOZZLE ' Y ' , with 50mm load/unload the print crashes !

**Firmware retraction activated** _Multi extruder print is impossible_
```
G1 X176.483 Y221.677 F7800.000
G1 X178.271 Y223.465 E924.73334 F1800.000
G10 ; retract    ************* NO SWAPPING RETRACTION AND NEED A G10 S1👎 
G92 E0
T1 ******* NOT RECOVERED OR LOADED OR PRIMED 👎 
G92 E0
G1 X200.222 Y222.931 F7800.000
G1 X200.154 Y223.286 E0.09169 F1800.000
G1 X199.990 Y223.777 E0.22301
G1 X199.819 Y224.121 E0.32070
```
**Using TOOLCHANGE TAB**
_ToolChange Tab  unusable because of REPEATING of the G10 than can confuse the machine_
```
G1 X214.271 Y160.965 E973.40355 F1800.000
G10 ; retract    ******************************* FIRST NORMAL RETRACT 👎 
G92 E0
G10 S1 ; Added by the SLIC3R tab *************** SECOND  RETRACT   BIG PROBLEM 👎 
T1
G92 E0
G10 ; retract ****************************** THIRD RETRACT  !!!!!!    it's too much  👎 
G92 E0
G1 X238.974 Y138.811 F7800.000
G11 ; unretract
G92 E0
G1 X238.783 Y139.181 E0.11126 F1800.000

```
[...]

PR #4108, opened 2017-09-02
Long retract support in marlin and repetier for FW retract.

lordofhyphens:
This is in the spec for both repetier and marlinfw.

Implements #4106 

Test checks for output from gcode->retract().


lordofhyphens:
Looks like "retract when tool is disabled" should itself be disabled on Reprap fw and firmware retract enabled, the two conflict with each other.

PR #4119, opened 2017-09-05
Move part meshes around in Object Settings.

### LINKED ISSUE #4118: Feature request: Move models around in Z in plater.
It would be great to be able to move the model in the z direction, especially for two extruder systems.
Unless I'm missing something.

lordofhyphens:
Re-added the ability to move part meshes relative to another part mesh in the same object.

Fixes #4118


lordofhyphens:
This puts the ability to move meshes relative to each other in the same object. The parent object is still stuck to Z=0 at its low point.

Issue #4054, opened 2017-07-02
Support not generated


freiser77:
### Version
slic3r-1.3.0-dev-13f121e-

### Operating system type + version
macOS Sierra

### Behavior
* Load the attached STL (scale: 16.1290322581, rotate X: -60 deg, rotate Z: 90 deg)
* Switch to preview tab
* Switch on support generation
* No support towers are generated below the feet which are not contacting the build plate
![screenshot](https://user-images.githubusercontent.com/14125584/27768420-89bdb138-5f13-11e7-887c-5d89bd7260a8.png)

_Is this a new feature request?_
No, bug report

#### STL/Config (.ZIP) where problem occurs
[Archiv.zip](https://github.com/alexrj/Slic3r/files/1117689/Archiv.zip)

PR #4129, opened 2017-09-20
Fix generating supports for very thin hovering slices

cangelis:
fix for #4054

Currently, slic3r does not generate support for very thin hovering surfaces. This PR fixes the issue. See the [link](https://www.dropbox.com/s/mecr68mqw035wrb/Grand%20Snaget-Jofo%20%281%29.stl?dl=0) for an example model.


lordofhyphens:
Looks fine by me. Starting the countdown for when someone else complains about it breaking something else.

PR #4140, opened 2017-09-29
Add lock/unlock using flock() for output gcode files.

### LINKED ISSUE #4134: Export does not lock the output file. (gcode)
### Version
1.2.9.41

_Use `About->About Slic3r` for release versions._

### Product Area
Export to file.

Windows 7 64bit

### Behavior
When exporting to file, Slic3r is not locking the file correctly. An OS copy action can be performed while the file is still being written too. It would appear that performing the copy and paste while slic3r is writing blocks slic3r from completing the write of the file. This results in a malformed file that will print, but will not complete the whole object.

### Reproduce
I believe this is a generic issue and therefore I shall not provide a config or files.
If it turns out not to be generic then I shall.
A) Observe the writing to file and the total size of the object,.
1. Import a large STL into slic3r (My object was about 120x40x40mm.
2. Open File explorer to the location where the file will be exported. (Make sure that "File Size" is visible.)
3. Using Slic3r, export you print to the location from 2.
4. Focus the explorer browser and repeatedly click refresh, The size of the file will slowly increase.
B) Malform the output file
1. repeat Steps 1-3 but export to a different name to keep your reference file.
2. Before the total file size has been achieved, perform a copy Paste within the folder.

You will now have two files, one of which is correctly formed, the other is significantly smaller. 
Printing with the small file will result in a print that will stop part way through the print.
I get the impression that homing at the end of print is a feature of the Gcode , but this will give the same effect as described by a "Faulty SD Card" symptom. 
It is in the domain of the user to make sure that the copy from HDD to SD completes correctly, however Slic3r put the output file into a state where it cannot be processed until the write to file is complete.
I have not tested other output format but the issue may manifest there too. 
In fact it may even be that this issue is only being hit after a duration of use as the size of build probably increases with confidence of using the 3D printer.

### Suggestion
Write to temp file and then rename at the end of the write operation.

No
Related guides for writing feature requests: http://meta.stackexchange.com/a/259196 http://nickohrn.com/2013/09/write-great-feature-request-bug-report/

_Upload a zipped copy of an STL and your config (`File -> Export Config`)_

lordofhyphens:
Fixes #4134


Careknight:
Tested against Slic3r-master.2017.09.29.1389.7bf8522.64bit on same Windows 7 64bit machine. 
Exports are now written to a temp file. Attempts to copy and paste do not interfere with the writing mechanism. When the operation completes the file is written in full to the relevant location with the correct ".gcode" extention. 

I am happy with this fix. Users will still need to make sure that they are not copying the temp files, but that is not unlike downloading files from a browser.

Regards.

lordofhyphens:
Not necessary as the main branch has equivalent functionality as reported by the requestor.

PR #4146, opened 2017-10-05
Added option to tell Slic3r CLI to not arrange at all.

### LINKED ISSUE #4145: Turn off auto-arrange / auto-centre in command line mode (CLI)
Huge thanks for the support on my last question, especially to @lordofhyphens . I stumbled upon another difficulty here probably considering my lack of knowledge in the code design structure. 

### Version 1.3.0-dev
revision : 5134 (I'm using TortoiseSVN)

### Operating system: Windows 10
A bit similar to issue #404

### Behavior
So I'm using a different GUI to pipe in information to the slic3r. (I'm guessing I'm using the slic3r-console.exe instead of the slic3r.exe)
Everytime I slice my STL, the resulting G-code is always shows that the STL is in the centre of the print bed. I even tried slicing it when the position is far off the bed, but the g-code still the same. 
My guess is that there is an auto-centering / auto-arranging stl defaultly turned on in the slic3r command line mode. I tried turning it off in the GUI, but as you can guess already, it does not affect the command line mode because they are independent. 

My question is, how can I turn off (or on) the auto centering in the command line mode?
And out of curiosity, when I slice with command line, am I really using the slic3r-console.exe? 

Thank you for your attention!

Best regards, 
Stephen

lordofhyphens:
Fixes #4145

### DIFF
--- a/lib/Slic3r/Print/Simple.pm
+++ b/lib/Slic3r/Print/Simple.pm
@@ -53,6 +53,11 @@ has 'dont_arrange' => (
     default => sub { 0 },
 );
 
+has 'model_coords' => (
+    is      => 'rw',
+    default => sub { 0 },
+);
+
 has 'output_file' => (
     is      => 'rw',
 );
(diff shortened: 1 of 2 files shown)


lordofhyphens:
@stepkurniawan please pull this branch and test it.

PR #4155, opened 2017-10-18
Fix comparison bw integer and pointer

cangelis:
This error is encountered while compiling on `gcc version 7.1.0 (x86_64-posix-seh, Built by strawberryperl.com project)`

`ISO C++ forbids comparison between pointer and integer [-fpermissive]`

This PR fixes the error.

Issue on original repository: https://github.com/artyom-beilis/nowide/issues/22

### DIFF
--- a/xs/src/boost/nowide/cenv.hpp
+++ b/xs/src/boost/nowide/cenv.hpp
@@ -101,7 +101,7 @@ namespace nowide {
     {
         char const *key = string;
         char const *key_end = string;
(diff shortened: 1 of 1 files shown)


lordofhyphens:
Can you add a null pointer check (using nullptr if c++11) ahead of the referenced code? I think the original code would segfault in this case (bad input). Returning error status would be sufficient I think.

lordofhyphens:
nvm, added it. Thanks for the PR.

PR #4156, opened 2017-10-18
Fix to autocenter option.

GilesBathgate:
Ensure that with autocenter turned off the part is not positioned
in the center of the bed.

### DIFF
--- a/lib/Slic3r/GUI/Plater.pm
+++ b/lib/Slic3r/GUI/Plater.pm
@@ -1264,6 +1264,7 @@ sub load_model_objects {
                 # if user turned autocentering off, automatic arranging would disappoint them
                 $need_arrange = 0;
 
+                $o->align_to_ground; # aligns object to Z = 0
                 $o->add_instance();
             }
         } else {
--- a/lib/Slic3r/GUI/Plater.pm
+++ b/lib/Slic3r/GUI/Plater.pm
@@ -1253,12 +1253,19 @@ sub load_model_objects {
         push @obj_idx, $#{ $self->{objects} };
     
         if ($model_object->instances_count == 0) {
-            # if object has no defined position(s) we need to rearrange everything after loading
-            $need_arrange = 1;
-        
-            # add a default instance and center object around origin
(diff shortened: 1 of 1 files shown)


lordofhyphens:
I'm okay with changing this behavior slightly, especially as autoarrange is on by default and triggering arrangement is a single click away. 

The only real concern I have is people with really idiotic STL coordinates such that they don't see it on the bed and then file bug reports against it.

GilesBathgate:
@lordofhyphens You make a good point about the coordinates being way outside the print bed area, and thus making the geometry invisible. I think it would be appropriate for slic3r to provide a warning message when the geometry is outside the print bed, (it already provides a similar warning when its too big for the build plater). I will see if I can create a new pull request that provides a warning message.

GilesBathgate:
@lordofhyphens I have prototyped the warning message change in https://github.com/alexrj/Slic3r/pull/4163

lordofhyphens:
Fun fact: This broke the slab mesh positioning.

GilesBathgate:
@lordofhyphens Damn, If you can describe the problem then I will see if I can provide a fix. Although if its the later, don't panic and have a warm relaxing cup of tea.

lordofhyphens:
@GilesBathgate  I tried to reproduce myself and couldn't. 

If you want to look yourself, grab a model, right-click -> Settings and create a slab modifier. In my tests it wasn't centering the slab properly in XY.  And then it was, hence my wondering about my hallucinating it in the first place.

GilesBathgate:
@lordofhyphens I think there is a change in behaviour, previously I think it would have centred the object, and the slab to the centre of the bed so they would be centre aligned. Because there is no repositioning of the part or the slab I think they can be misaligned. I think the slab should be positioned to the centre of the existing object and not to the centre of the bed or the 0,0 origin.

GilesBathgate:
@lordofhyphens Well I have managed to fix the unintended behaviour, but I am not confident that this is the best approach. I tried creating something more robust but I couldn't get it to work. The proposed fix certainly worked in my limited test cases.

https://github.com/alexrj/Slic3r/pull/4170

PR #4157, opened 2017-10-19
Math evaluation in gcode

### LINKED ISSUE #3390: Proposal: Conditional Gcode Syntax for Custom Gcode

lordofhyphens:
Starting to implement the following spec:

https://github.com/alexrj/Slic3r/wiki/Conditional-Gcode-Syntax-Spec

Fixes #3390 when it is finished.


lordofhyphens:
Not currently in the execution path, and currently assumes that the placeholders have been replaced before this runs.

ledvinap:
@lordofhyphens : Isn't it possible to use some existing library? Something based on grammar?

This approach will become complicated quickly. For example consider implementing unary `-`. Operator precedence is another complication.

bubnikv:
https://wiki.inspired-lua.org/math.eval

lordofhyphens:
Yeah, that was just for messing around. Going to find a lightweight in fix
parser for the math strings.

lordofhyphens:
... for certain values of "lightweight". 

Currently using exprtk by @ArashPartow, which is certainly much more powerful than strictly necessary, but it does support everything that is in the spec. 

Probably wise to disable the RTL I/O capabilities during compilation.

Current limitations:
math requires happening last in the gcode stack.

lordofhyphens:
Okay, it's passing the basic tests. If people find issues it should be straightforward to add the failure cases to the tests file.

If something fails to parse it emits a blank string.

xs/src/libslic3r/ConditionalGcode.cpp
```diff
@@ -0,0 +1,133 @@
...
+        #endif
+    }
+    std::string output = result.str();
+    trim(output);
+    return output;
+} 
+
+
+/// Parse an expression and return a string. We assume that PlaceholderParser has expanded all variables. 
+std::string expression(const std::string& input, const int depth) {
```
ledvinap: With `exprtk` all that is IMO necessary is to replace (multiple) `{expr}` with its evaluated form (ideally including string result) and handle `{if expr}` as conditional GCode. There is no need to handle subexpressions at this level (and it will interfere with exprtk)

With string support, there will appear problem with curly brackets embedded in strings. 
Minor problem may be with embedded placeholders, but it's easy to workaround (split text to avoid placeholder expansion)

lordofhyphens:
I disabled string support and conditional gcode happens after placeholder
parser.

lordofhyphens:
Show me a failing string and I will add it to the tests.

ledvinap:
@lordofhyphens : Without string support there is no problem.

With string support, curly brace may be embedded in string, so valid  expression may have unbalanced (raw) braces.

Also with strings, it may be desirable to put placeholder in generated output.

Both cases are mostly theoretical - edge cases that may cause problems in distant future.

lordofhyphens:
Sounds more like a possible future feature request, then ;)

platsch:
Yay, nice feature! Just started using it and it helps a lot.
Other than the wiki page https://github.com/alexrj/Slic3r/wiki/Conditional-Gcode-Syntax-Spec there is no documentation yet, right? The manual page http://manual.slic3r.org/advanced/conditional-gcode linked in the wiki throws 404.
I suppose inline documentation with tooltips is a bit pointless, since the evaluation is applied to every custom gcode and too complex for a tooltip explanation. Maybe a short hint in every custom gcode field tooltip?

lordofhyphens:
Unfortunately I need @alexrj to get the manual updated.

platsch:
Two more annotations:
- the conditional syntax parser currently doesn't apply to toolchange gcodes, fix submitted in #4229
- I think we should generally expose more dynamic variables to custom gcodes. Example: in my setup, I frequently want to wipe / prime extruders depending on there state for particular layers, i.e:
  - compensate retraction, 
  - wipe, 
  - apply current retraction again to get a correct start.

But the retraction can be different for different extruders and is also different for the first layer where toolchange retraction is not applied until an extruder is used the first time.
I added `[previous_retraction]` and `[next_retraction]` to the toolchange gcode and `[current_retraction]` to before / after layer change gcode in my slic3r installation. Can open a PR for this if someone is interested.

lordofhyphens:
Sounds pretty good to me.

platsch:
PR is updated to include the dynamic variables.

PR #4164, opened 2017-10-21
Fix not generating supports

cangelis:
Slic3r does not generate support correctly for some models. Download an example [here](https://www.dropbox.com/s/oi2u2m2efjlzhdz/slic3r-support-not-generating-example.stl?dl=0)

`clip_with_object()` method uses wrong `$support_z` indices `$i` which are support layer heights (`$support` hash indices).

This PR fixes the `$zmax` and `$zmin` at `clip_with_object()` and apply the method only on `$contact` rather than `$interface`s and `$base`s (which are generated using `$contact`). This also improves the performance drastically on complex models. 

This PR also fixes the first interface layers' xy spacing, which was not calculated before.

### DIFF
--- a/lib/Slic3r/Print/SupportMaterial.pm
+++ b/lib/Slic3r/Print/SupportMaterial.pm
@@ -49,6 +49,7 @@ sub generate {
         [ sort keys %$top ],
         max(map $_->height, @{$object->layers})
     );
+    $self->clip_with_object($contact, $support_z, $object);
     
     # If we wanted to apply some special logic to the first support layers lying on
     # object's top surfaces this is the place to detect them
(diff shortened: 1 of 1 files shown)


lordofhyphens:
Could you put a picture in of what looks wrong with the old logic with that model? I can't really see the difference myself with the provided model and my collection of settings.

A config export would also be helpful (you can upload .zip files to github).

cangelis:
These are the screenshots of the wrong interface layer and the contact layer. Please see the [link](https://www.dropbox.com/s/g92hkojkzzeme21/config.ini?dl=0) for the config.ini file.

screen shot 2017-10-23 at 09 49 30

screen shot 2017-10-23 at 09 49 22

cangelis:
Oh sorry this example works well on the current code. I need to find another one. I will provide as soon as I find.

The reason I encountered with the issue is that my codebase works different z-spacing for the support materials.

lordofhyphens:
@cangelis any updates?

cangelis:
Could not regenerate the issue. Better to close the PR until somebody encounters again!

lordofhyphens:
Thank you for trying to reproduce.

PR #4170, opened 2017-10-24
Ensure that modifiers are aligned to the center of the part.

GilesBathgate:
When autocentering is turned off the modifiers have to be aligned to the centre of the part.

### DIFF
--- a/lib/Slic3r/GUI/Plater/ObjectPartsPanel.pm
+++ b/lib/Slic3r/GUI/Plater/ObjectPartsPanel.pm
@@ -407,6 +407,14 @@ sub on_btn_lambda {
     } else {
         return;
     }
+
+    my $center = $self->{model_object}->bounding_box->center; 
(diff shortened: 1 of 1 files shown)


GilesBathgate:
@lordofhyphens Hang fire, I will see if I can come up with a better solution.

GilesBathgate:
@hyperair Can you have a look at this? It needs fixing but I am not sure which is the best approach. This works but I think its a bit messy. Thanks.

hyperair:
I'm not sure autocentering modifiers is a good thing. They're usually translated to the correct location in the STL before being imported into Slic3r as a modifier, so autocentering the modifier is going to throw things off.

GilesBathgate:
@hyperair This relates to the built-in modifiers such as Slab, Box, Cylinder and Spere (accessible via the `Create Modifier` button, not the `Load Modifier` button)

PR #4177, opened 2017-10-31
Add pause-at-layer functionality for 3092

goofdad:
Added 2 new configuration parameters: pause_gcode and pause_heights.

pause_gcode is under the Custom G-Code section of Printer Configuration.
pause_heights is under Layers and perimeters section of Print Configuration.

also added a test case.

### DIFF
--- a/lib/Slic3r/Print/GCode.pm
+++ b/lib/Slic3r/Print/GCode.pm
@@ -453,6 +453,16 @@ sub process_layer {
         $pp->set('layer_z'   => $layer->print_z);
         $gcode .= Slic3r::ConditionalGCode::apply_math($pp->process($self->print->config->layer_gcode) . "\n");
     }
+    if ($self->print->config->pause_gcode && $self->print->config->pause_heights) {
+        foreach my $height (split / /, $self->print->config->pause_heights) {
+            if ($height && $height eq $layer->print_z) {
+                my $pp = $self->_gcodegen->placeholder_parser->clone;
+                $pp->set('layer_num' => $self->_gcodegen->layer_index);
+                $pp->set('layer_z'   => $layer->print_z);
+                $gcode .= Slic3r::ConditionalGCode::apply_math($pp->process($self->print->config->pause_gcode) . "\n");
+            }
+        }
+    }
     
     # extrude skirt along raft layers and normal object layers
     # (not along interlaced support material layers)
--- a/xs/src/libslic3r/PrintConfig.hpp
+++ b/xs/src/libslic3r/PrintConfig.hpp
@@ -314,6 +314,8 @@ class PrintRegionConfig : public virtual StaticPrintConfig
 class GCodeConfig : public virtual StaticPrintConfig
 {
     public:
+    ConfigOptionString              pause_gcode;
+    ConfigOptionString              pause_heights;
     ConfigOptionString              before_layer_gcode;
     ConfigOptionString              between_objects_gcode;
     ConfigOptionString              end_gcode;
@@ -355,6 +357,8 @@ class GCodeConfig : public virtual StaticPrintConfig
     }
     
     virtual ConfigOption* optptr(const t_config_option_key &opt_key, bool create = false) {
+        OPT_PTR(pause_gcode);
+        OPT_PTR(pause_heights);
         OPT_PTR(before_layer_gcode);
         OPT_PTR(between_objects_gcode);
         OPT_PTR(end_gcode);
--- a/lib/Slic3r/GUI/PresetEditor.pm
+++ b/lib/Slic3r/GUI/PresetEditor.pm
@@ -441,7 +441,7 @@ sub options {
         perimeters spiral_vase
         top_solid_layers bottom_solid_layers
         extra_perimeters avoid_crossing_perimeters thin_walls overhangs
-        seam_position external_perimeters_first
+        seam_position external_perimeters_first pause_heights
         fill_density fill_pattern top_infill_pattern bottom_infill_pattern fill_gaps
         infill_every_layers infill_only_where_needed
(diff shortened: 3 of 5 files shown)


xs/src/libslic3r/PrintConfig.hpp
```diff
@@ -355,6 +357,8 @@ class GCodeConfig : public virtual StaticPrintConfig
     }
     
     virtual ConfigOption* optptr(const t_config_option_key &opt_key, bool create = false) {
+        OPT_PTR(pause_gcode);
+        OPT_PTR(pause_heights);
```
lordofhyphens: Please move these so that they're in alphabetical order in the list.

lordofhyphens: Please move these so that they're in alphabetical order in the list.

xs/src/libslic3r/PrintConfig.hpp
```diff
@@ -314,6 +314,8 @@ class PrintRegionConfig : public virtual StaticPrintConfig
 class GCodeConfig : public virtual StaticPrintConfig
 {
     public:
+    ConfigOptionString              pause_gcode;
+    ConfigOptionString              pause_heights;
```
lordofhyphens: Please move these so that they're in alphabetical order in the list.

lib/Slic3r/Print/GCode.pm
```diff
@@ -453,6 +453,16 @@ sub process_layer {
         $pp->set('layer_z'   => $layer->print_z);
         $gcode .= Slic3r::ConditionalGCode::apply_math($pp->process($self->print->config->layer_gcode) . "\n");
     }
+    if ($self->print->config->pause_gcode && $self->print->config->pause_heights) {
+        foreach my $height (split / /, $self->print->config->pause_heights) {
+            if ($height && $height eq $layer->print_z) {
```
lordofhyphens: Wouldn't it be appropriate to round the effective height (either up or down, down may be the safer option) if the current layer doesn't land on the prescribed height? 

It's something that can occur often with adaptive slicing turned on or if there's some oddball floating point math going on internally. 

Obviously you'd want to extend the tooltip appropriately.

lordofhyphens:
Implementing #4178 is another route to getting this PR with this functionality in ;)

lib/Slic3r/Print/GCode.pm
goofdad: I'll have to play with how this interacts with Adaptive Slicing, but my gut says no. I also work with ideamaker, which has this feature and does such rouding ... and the end result is that I have to manually open my gcode after export to make sure it paused where I wanted it to. They have multiple bugs open because of this. (now ... I have to open the gcode anyway because they don't allow me to customize the code inserted, but that's a different story).

goofdad:
I'll look at some changes in the morning after coffee (never code before coffee!).
I'll also take a look at #4178 because I agree ... the UI is getting cluttered!

lib/Slic3r/Print/GCode.pm
lordofhyphens: Since logically this is intended to be a before-layer gcode, using floor() would seem to be safe (as opposed to round(), which could go up or down and definitely could confuse people). 

That is, if layer 5 spans z=0.1 and z=0.3 and z=0.2 was requested, it would be inserted at 0.1. If z=0.299999999 was requested, it'd still be at z=0.1.

goofdad:
In response to your Philosophical Question:
For me, the workflow warrants being separate and not an if statement. Philosophically, the actual code being inserted is printer specific, and thus belongs in the "Printer" config, while the layers needed is part of this specific print, and thus belongs in the "Print" config. I wouldn't want to need to go mucking about with my Printer config for each and every print where I wanted a pause.

lordofhyphens:
@goofdad I did write #4178 with your use-case in mind ;)

I also just edited the main description to put some notes on some places that would need to be touched to implement this (not an exhaustive list).

ledvinap:
One possibility is to replace `pause_heights` with `mark_heights` and allow G-code scripting to access `mark` status (possibly with set of mark values).
Then `before_layer_gcode` would just use `{if [layer_mark]}` (or `{if [layer_mark] == 2}` / `{if [layer_mark] & 0x04}`)
Similar mechanism could be used for example in other g-code parts (toolchange gcode - pause before toolchange on specific layer)

goofdad:
I actually like the layer mark idea, but I'd like it to be more general. Now I'm contemplating syntax between what belongs on the "Printer" side and what belongs on the "Print" side.

lordofhyphens:
I am down with both (hence overrides in Print and Filament).

On Nov 1, 2017 8:49 AM, "goofdad" <notifications@github.com> wrote:

I actually like the layer mark idea, but I'd like it to be more general.
Now I'm contemplating syntax between what belongs on the "Printer" side and
what belongs on the "Print" side.

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/alexrj/Slic3r/pull/4177#issuecomment-341111084>, or mute
the thread

lordofhyphens:
@goofdad *prod*

goofdad:
Christmas...and a hypercube build.
I'll beat my head on this again soon...

lordofhyphens:
'S okay. I have 53+ hours logged at work for the week so far.

lordofhyphens:
@goofdad still need those changes made to accept.

lordofhyphens:
@goofdad :)

tcurdt:
@goofdad any plans on making those changes? or should someone else take over?

lordofhyphens:
@tcurdt Probably should take a swing and/or make PRs against his branch.

Issue #404, opened 2012-05-31
Feature Request: Turn off auto centering.


Intrinsically-Sublime:
First off thanks again for all your hard work slic3r is amazing and never stops getting better.

It would be really nice now that you have the GUI to be able to place an item(s) on the build surface without it auto centering every time. Maybe just a check box that says auto centering enabled? This is so when you wear out your build surface in the middle you can print in the corners or along the side and don't need to replace the tape as often. Also so you don't where out your smooth rods in the middle.

GilesBathgate:
@alexrj I am not sure the implementation does what is expected. (or my expectations are incorrect). The change does stop the parts from being auto-arranged, but if I load cube.stl (with its corner fixed to [0,0,0]) it is automatically placed in the center of the bed instead of the co-ordinates specified in the stl.
If I then add another cube2.stl (with its corner fixed to 50,0,0) it places that cube in the center on top of the original:

![image](https://user-images.githubusercontent.com/444143/31716015-148ed3d4-b3fe-11e7-8375-f4a6d17ea3f9.png)

What I was expecting was that the positions should not be altered at all and should just be what was specified in the stl file

![image](https://user-images.githubusercontent.com/444143/31716183-92de241a-b3fe-11e7-9406-b8bf15de3e94.png)

I have created a pull request https://github.com/alexrj/Slic3r/pull/4156 that changes the behaviour so that it matches what I expect (i.e the second screenshot)

expertmm:
see also (underlying issue): #4190

GilesBathgate:
@expertmm Well I think the underlying issue is fixed and merged in PR #4156 . Perhaps you could let me know after testing (a version more recent than Oct 20) it if there is still an issue.

expertmm:
"It would be really nice now that you have the GUI to be able to place an item(s) on the build surface without it auto centering every time." this is the specific part of the issue to which I was referring. For example, even with centering off, in recent git version as described in 4190, all of these end up at the same place. To reproduce:
* Add stl files (attached) to Slic3r (see attached Slic3r screenshots)
* compare result to importing them into other program such as Blender (see attached Blender screenshot)
![after importing all six into blender shows that blender had applied the position before exporting](https://user-images.githubusercontent.com/7557867/33630262-9b416d3a-d9fe-11e7-9509-5c3ccb245c57.jpg)
[...]

PR #4218, opened 2017-12-06
Auto align z preference

GilesBathgate:
This adds a new preference to turn off auto alignment on the z axis (z=0). This can be useful when the input file already contains the correct z-axis location, as discussed here: https://github.com/alexrj/Slic3r/issues/404#issuecomment-349438749

### DIFF
--- a/lib/Slic3r/GUI.pm
+++ b/lib/Slic3r/GUI.pm
@@ -84,6 +84,7 @@ our $Settings = {
     _ => {
         version_check => 1,
         autocenter => 1,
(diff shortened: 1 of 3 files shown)


lordofhyphens:
@GilesBathgate  could you draft a short PR to add this to Slic3r-Manual?

GilesBathgate:
@lordofhyphens I am happy to draft what this preference was 'supposed' to do, however, there seems to be some contention in #4273 as to whether the current behaviour is actually correct....

PR #4229, opened 2017-12-14
apply conditional gcode parser to tool change gcode

platsch:
In the current version, the conditional gcode parser doesn't apply to tool change G-codes, which is a major drawback (at least in my use-cases).
This PR executes the conditional parser for every toolchange.
I had to add an include guard to `ConditionalGcode.hpp` since it is already included via XS, otherwise it is a one-liner.


lordofhyphens:
Thanks for catching the include guard too.

lordofhyphens:
Could you make a short note in the relevant section of https://github.com/alexrj/Slic3r-Manual ?

PR #4297, opened 2018-02-12
[Feature Request] Add point3 support to UI

rockandsalt:
### Version
_1.2.9-1289-g67bd3ab5_

_1.3.0-dev_

_Ubuntu 64-bit_

### Behavior
* _Describe the problem_
I would to add a point3 as a config and be able to set it in the UI. But right now, slic3r only supports adding point. 

I tried to implement it. But when I set a coPoint3 type config property in printConfig.hpp .  It is set to point as default in perl. The point3 is therefore treated and displayed as a regular point in the UI(only x and y input box are shown). Also the slic3r::Pointf3 is being serialized raw in perl but I can't figure out where. 

### What works so far:

I created this pull request to show the work I have done so far. I am also working on adding support for a new flavor of gcode for a Nordson machine. So there's bits of that here and there. There's also a bunch of debug around that I will remove eventually.


lordofhyphens:
I won't be able to accept a PR that has 'this and that' in it. You may need to strip out the two components and submit two PRs later. Just FYI.

rockandsalt:
@lordofhyphens  yeah this is understandable. I was not expecting the PR to be approved. 

Right now  I was hoping to get some hints on where to look or why my the implementation doesn't work. As I mentioned, object slic3r::Pointf3 is being passed raw to the serialize function instead of being in "float , float ,float". I can't pinpoint where this happends.

The object Slic3r::GUI::OptionsGroup::Option in perl always has the type property set to "point" whenever I insert a coPoint3 in the config.I modified the Config.xsp file but it's doesn't seem to solve the problem. So I am not sure where coPoint3 are being mapped to type "point".

lib/Slic3r/GUI/OptionsGroup/Field.pm
```diff
@@ -510,6 +510,88 @@ sub disable {
...
+    my ($self) = @_;
+    
+    my $sizer = Wx::BoxSizer->new(wxHORIZONTAL);
+    $self->wxSizer($sizer);
+    
+    my $field_size = Wx::Size->new(40, -1);
+    
+    $self->x_textctrl(Wx::TextCtrl->new($self->parent, -1, $self->option->default->[X], wxDefaultPosition, $field_size));
+    $self->y_textctrl(Wx::TextCtrl->new($self->parent, -1, $self->option->default->[Y], wxDefaultPosition, $field_size));
+    $self->y_textctrl(Wx::TextCtrl->new($self->parent, -1, $self->option->default->[Z], wxDefaultPosition, $field_size));
```
lordofhyphens: So this might be your problem.

lordofhyphens:
Other than an unfortunate typo, it looks to me like you're doing it right.

rockandsalt:
@lordofhyphens  I believe this branch is ready for merge. I cleaned up my stuff and the point3 has been added to the UI. Let me know if you believe I should write some unit test.

lordofhyphens:
Please rebase the commits that are the intended contributions, add a unit test, and submit that as a PR instead of this.

Issue #3666, opened 2017-01-13
Create option for gyroid infill


mrjohnc:
Please create gyroid infill that is the shape as graphene. This shape has been shown to be very strong as a 3D printed object and if used as an infill would presumably create very strong 3D prints

http://itsybitsymag.com/2017/01/11/mit-research-ten-times-strong-lightweight-graphene/

I printed a gyroid shape where all surfaces are 0.4mm thick (the size of my nozzle), I think this is the same as what the infill would be like, the model is 40mm x 40mm. It is very very strong, I think I could maybe stand on it and it would be ok, it weighs very little.

I got the model from here https://grabcad.com/library/math-visualization-gyroid-1

![8c81dbf6-d980-11e6-8361-d3563cd03aad](https://cloud.githubusercontent.com/assets/7293392/21932158/66c2e6ac-d99f-11e6-9f73-673ba240cc9b.jpg)

Note: I have also created this request on the Prusa fork of Slic3r (I have a Prusa Mk2) https://github.com/prusa3d/Slic3r/issues/92

forReason:
seems nice but is it stronger than current infill like 3d honeycomb? seems rather comlex...

mrjohnc:
I created some test cubes to compare the strengths of infill, any ideas how to compare strengths? 

http://www.thingiverse.com/thing:2091011

lordofhyphens:
There is the crush test: use a press and slowly increase pressure until
something gives.

RafaelEstevamReis:
Poor man solution:
Put the object on a Scale and a cup on top of it and reset (zero) scale's
value.
Then put gravel, rocks, sand, water (caution, the cup can fall) until you
hear something crack.
Look at the scale.


2017-03-09 1:02 GMT-03:00 Joseph Lenox <notifications@github.com>:

lordofhyphens:
@mrjohnc any results from experimenting?

mrjohnc:
@lordofhyphens not really, I'm stuck on how to test fro strength. Any ideas that would be of suitable quality and repeatability?

lordofhyphens:
Crush test. If you are around a university with a mech engineering program
then they would have a good tool for measuring structural strength.

lordofhyphens:
Make sure you print with the same settings and the same roll of filament.

For your modeled boxes (assuming that the infill is modeled before it gets
to the slicer) perimeters very high and set infill to 0%.

Work out how much force it takes to deform and break the object.

[...]

PR #4317, opened 2018-02-22
gyroid infill

supermerill:
Gyroid infill.
Same version than in the PE.
#3666


lordofhyphens:
I've converted a few of your comments to the Doxygen triple slash instead, but otherwise am merging this.

foreachthing:
Where would I post an issue with the Gyroid pattern?

PR #4318, opened 2018-02-25
Reduce waste for MMU

kevinBecker:
I heard of people using the infill to chang ecolors instead of a separate purge tower, would this be possible in slic3r with the prusa mmu? For prints with little infill an extra tower might be needed, but it should save a lot of plastic overall, as the infill is not seen.  It would need two settings:

1.  minimum extrusion length for tool change setting (to ensure enough plastic is extruded and the color change is complete) (the user would input in mm the length of plastic needed to change)
2. An option that allows slic3r to increase the infill to make sure the color is completely changed in the infill, thus making the print heavier and eliminating the need for any purdge tower at all for most prints (and if this is not enough, have a small purge tower next to the print) (just needs a yes/no input to turn it on or turn it off if they want to reduce the weight of the part)

side note: For prints with more than two colors, it would have to switch from printing perimiters, to part of the infill as a color change, back to perimiters, then finish the infill for that layer.  For programing this in, should it have an extrusion length counter when writing the infill gcode to ensure it hits the min value set by the user (see point 1), and if not, increasing the infill on that layer or creating & going to a purge tower on the side (point 2).

 Obviously it would only work with multi-color prints and not multi material prints. Nonetheless, this would make the parts printed with the mmu less wasteful and the purge tower plastic would make the print stronger instead of go into the trash. Thank you for reading all of that! Any other suggestions/ideas about this would be nice to hear!


lordofhyphens:
https://github.com/alexrj/Slic3r/issues/new

Issue #4255, opened 2018-01-08
Adaptive Slicing changes overall thickness of parts


foreachthing:
### Version
Slic3r-master.2017.12.18.1561.d27b8a9.64bit

### Operating system type + version
Windows 7 x64

## Setting 1
* Got some parts, all 3 mm in thickness.
* Sliced it using the _adaptive slicing_ option - quality set to **80%**.

### Result
* Sliced parts are now **2.96 mm** thick.

## Setting 2
* Got a part, 3 mm in thickness.
* Sliced it using the _adaptive slicing_ option - quality set to **81%**.

### Result
* Sliced parts are now **3.04 mm** thick.

# Expected Behavior
* Same thickness for all quality settings!

Unless the quality setting is set to 0 or 100% (sorry, didn't try everything between 0 and 100.0), the thickness is _wrong_.

lordofhyphens:
@platsch

lordofhyphens:
@foreachthing please provide the usual: a set of exported configurations (ctrl-e for a single config, not a bundle), and a 3d model you see the issue on.

platsch:
Could be a spline-related problem...
Generally: for static slices, the thickness of an object can't be matched exactly unless it happens to be a multiple of the layer thickness. The adaptive algorithm should provide better, but in some cases still not exact results. Please provide the model for verification.

foreachthing:
@lordofhyphens Sorry for my laziness not uploading any model nor config file. This was with all models so I didn't bother.
So here is "any" model with the config file. Different model, same issue.
[cube10x10x3.zip](https://github.com/alexrj/Slic3r/files/1614215/cube10x10x3.zip)

As @platsch mentioned, this must be a math problem. 
Oh, and set the quality to zero (with this model) I only get 2.9 mm thickness this time.
Please let me know if you need more info.

platsch:
Thx, should do. Will check this the next days.

lordofhyphens:
Thank you. Even "all models do this" should include a simple model ;)

platsch:
This is a general problem related to the discretization into 2D-layers.
As I said, for static layers, the last layer generally doesn't match the top of an object, unless the user sets a layer height which does.

In your example, the adaptive algorithm tries to reduce the layer thickness towards the top to resemble the original model as precise as possible, but doesn't care about the top.

[...]

PR #4321, opened 2018-02-26
Match object top

platsch:
Potential fix for #4255 
Try to match object top by reducing / increasing thickness of last layer after adaptive layer generation.
Should not brake anything, but I'd like to get feedback whether or not this is useful before merging.


lordofhyphens:
@platsch did anyone give this a try? I think I'd rather merge at this point and see what breaks than try to get people to try pull requests.

Issue #3485, opened 2016-08-27
'Pillars' Infill columns may not be reliably printable


lordofhyphens:
### Version
_Version of Slic3r used goes here_
1.2.9, 1.3.0-dev
### Operating system type + version
All

### Behavior
* The thin towers produced by Slic3r may not be reliably printable with some printers.

![support](https://cloud.githubusercontent.com/assets/31754/18025658/8aed55f6-6bf6-11e6-8d91-e6775770f4c6.png)

_Is this a new feature request?_
Yes

### Suggested solution
The two primary options for pillar infill, spacing and size, are set as constants in SupportMaterial.pm.
While it is certainly feasible to make those options in the configuration file, need to solve some UI questions first to avoid cluttering the UI.

#### STL/Config (.ZIP) where problem occurs
[issue.zip](https://github.com/alexrj/Slic3r/files/440507/issue.zip)
Note: rotate the stl 180degrees around the X axis

lordofhyphens:
I've got an experimental (just looking at adding config options, not dealing with the UI) branch at: https://github.com/lordofhyphens/Slic3r/tree/config-pillar.

I've left the default options to what the constants were, so existing tests pass. I do need to come up with a few tests.

lordofhyphens:
@alexrj any comments with regards to your preferences as to UI, etc?

Lord2Vader:
Have you tried using a wider line width for the support? Works out for me most of the time. Although yes these pillars are very veeery fragile.

lordofhyphens:
I don't usually have problems, but I have heard them from others.

Since the parameters are already there, it is pretty straightforward to add
them to config (probably behind an optional menu that only is available
after setting that type.

DrLex0:
The biggest problem I see with the pillars is lack of footprint. A tall narrow pillar becomes very easy to knock over unless bed adhesion is insanely good. Even if it is not knocked over, when a narrow pillar becomes tall enough, it becomes too flexible and wobbles around while things are printed on top of it.

Pillars should be the wider, the taller they are. At least their footprint should be, the top end may be narrower.

![pillars](https://cloud.githubusercontent.com/assets/16401844/18226536/305a573a-720c-11e6-8520-9a167fe92d7c.jpg)
[...]

Issue #2017, opened 2014-05-06
Make pillars support material Z axis symmetrical to prevent them tipping over


magu:
First of all, the pillars support material is awesome. Much, much easier to remove than the others. I do however have encountered a big snafu with them tipping over most of the time.

![2014-05-06 20 43 26](https://cloud.githubusercontent.com/assets/79853/2894086/3b8032da-d54f-11e3-8378-091e5d6b4f05.jpg)

Attached is an image from a current print (ABS@245C, 0.2 mm layer height, 0.5 mm extrusion width, 5 mm brim, and supports (even for bridges); all other settings slic3r 1.1.2 default) with pillars supporting long bridges. Despite them only being 10 mm high, and having a 5 mm brim, most have still tipped over.

I think a solution to this would be to make the pillars hourglass shaped (with extendable middle), i.e. mirroring the top funnel shape on the bottom as well for increased stability. Would this be easy to implement?

lordofhyphens:
Actually, I think it would be fairly straightforward to implement.

PR #4327, opened 2018-03-01
Attempt to bridge support layer pillars

nylen:
This PR is intended to help address issues with the "pillars" support method producing very thin pillars that sometimes fall over during print (for example #3485 and #2017).

It simply disables retraction inside of all support layer moves, which helps the pillars adhere to each other without using much extra material.

It would be better to implement a more full fix like requiring a minimum area for support pillars (see #2949) or explicitly bridging the pillars, but that is beyond my current skill with this complicated codebase :)

If the approach taken in this PR is generally acceptable, there is still more that should be done:

- Preserve retraction behavior on the base layer of the support material
- Preserve retraction behavior when moving into or out of a support layer (maybe check if the starting and ending points of `travel` are within the support islands)
- Don't change behavior for other types of support material
- Tests

I would appreciate tips on how to achieve any of those things.

### DIFF
--- a/xs/src/libslic3r/GCode.cpp
+++ b/xs/src/libslic3r/GCode.cpp
@@ -681,8 +681,8 @@ GCode::needs_retraction(const Polyline &travel, ExtrusionRole role)
     
     if (role == erSupportMaterial) {
         const SupportLayer* support_layer = dynamic_cast<const SupportLayer*>(this->layer);
-        if (support_layer != NULL && support_layer->support_islands.contains(travel)) {
-            // skip retraction if this is a travel move inside a support material island
+        if (support_layer != NULL) {
+            // skip retraction if this is a travel move for a support layer
             return false;
         }
     }


nylen:
In add0104 I'm leaving retraction enabled if each point of `travel` is within the support islands.  This works (retraction occurs) for moves between a support structure and an object section.  It's not enough, though: retraction should still happen on the first layer, and when a move between two support islands crosses a perimeter of the object being printed.

lordofhyphens:
Wouldn't this, in general, add stresses to those small pillars due to plastic pulling? 

I'm very sorry for not seeing this until now, I appreciate the work.

lordofhyphens:
@nylen it'd probably be more fruitful to begin by porting the support approach taken by Prusa3d (or going from scratch with https://github.com/slic3r/Slic3r/wiki/Support:-Requirements). Either way it gets things out of Perl (which is a goal for the project IMHO).

Issue #3001, opened 2015-07-20
Raft pattern angles


wfcook:
I've noticed a couple of things about rafts that are less than ideal:

(1) The first two layers of the raft are printed at the same pattern angle, which results in the first layer (thicker extrusion) not properly supporting the second layer (thinner).

(2) The first interface layer is printed at the same pattern angle as the first raft angle, no matter what angle the prior raft layer was printed at. Again, this results in improperly supported interface material unless an odd number of non-interface raft layers is used.

(3) All interface layers are printed at the same angle. This is not necessarily a problem, but not what I expected and I would think it would be better if they alternated like the other layers. What it the reason for this?

Thanks,
Bill Cook _/)_

GregoryNoel:
I experienced a problem with interface layers in the same direction. The first layer was not perfectly sticking to the support material (5mm height) and then the second interface layer was imperfect.
The first layer above the support material is in fact very poor in quality because not sticking well on the irregular height of interface layer -> Very poor finish on this object layer.

GregoryNoel:
I'm not expert in perl, but I've tried changing angles. I add 2 lines in SupportMaterial.pm.
Interface layers are now each time rotated.

If someone want to try, to enhance and integrate it, feel free.

![interface_angle_0](https://cloud.githubusercontent.com/assets/13418536/9427669/b468e020-4989-11e5-8fb4-53c136495242.jpg)
![interface_angle_1](https://cloud.githubusercontent.com/assets/13418536/9427672/b49ed590-4989-11e5-8922-1ea75cdf8303.jpg)
![interface_angle_2](https://cloud.githubusercontent.com/assets/13418536/9427670/b492e6e0-4989-11e5-8436-2b3a26acfc0e.jpg)
![interface_angle_3](https://cloud.githubusercontent.com/assets/13418536/9427671/b498ab16-4989-11e5-83cd-eefee63940cc.jpg)

a4jp-com:
I can't understand what is happening here. My layers are working and each layer is rotated 90 degrees from the last so everything seems to work. How do you make the supports? Do you have a simple test file? Maybe a cube with supports on the side or something.

![screen shot 08-24-15 at 03 06 pm](https://cloud.githubusercontent.com/assets/9621044/9433755/c679e104-4a71-11e5-9914-ccf2642ca0f0.JPG)
[...]

PR #4334, opened 2018-03-05
Raft pattern angles (#3001)

dn63:
Added changes suggested by @GregoryNoel that alternates interface layers by 90 degrees (each layer) which improves layer support.

Changes can be seen when viewing interface layers in the layer preview for a given object (set interface layers >1).


lordofhyphens:
I'll accept this PR; if people complain about it we'll make the behavior configurable.

PR #4325, opened 2018-02-28
Post-Processing-Script Parameter(s)

foreachthing:
Added ability to use parameters with post-processing-scripts.

`c:\Slic3r_PostProcessing\Slic3rPostProcessing.exe ! --i ! --o "c:\temp\outputfilename.gcode"`
becomes:
`c:\Slic3r_PostProcessing\Slic3rPostProcessing.exe --i SLIC3RFILEANDPATHNAME.gcode --o "c:\temp\outputfilename.gcode"`

or 
`c:\Slic3r_PostProcessing\Slic3rPostProcessing.exe ! "c:\temp\outputfilename.gcode"`
becomes:
`c:\Slic3r_PostProcessing\Slic3rPostProcessing.exe SLIC3RFILEANDPATHNAME.gcode "c:\temp\outputfilename.gcode"`

or 
`c:\Slic3r_PostProcessing\Slic3rPostProcessing.exe`
becomes:
`c:\Slic3r_PostProcessing\Slic3rPostProcessing.exe SLIC3RFILEANDPATHNAME.gcode`

See here: https://github.com/alexrj/Slic3r/issues/4000

Issue #4000, opened 2017-06-02
Post-Processing Script Parameters?


foreachthing:
### Version
Any, so far.

### Operating system type + version
Windows 7 Pro

### Behavior
I made a post-processing script which accepts parameters (like: `Slic3rPostProcessing.exe --output "c:\GeeeeCode\[input_filename_base].gcode" --v 4 --i` ) but for some reason, whenever I add such a parameter to the script input box it will fail to execute my script.

```---------------------------
Error
---------------------------
The configured post-processing script is not executable: check permissions.
(c:\temp\Slic3rPostProcessing.exe --v 4 --i )
---------------------------
OK   
---------------------------
```

Is it not possible to pass parameters?

thuSkywalker:
I met the problem too. My solution is to write a `.bat` file, e.g. `postprocessing.bat` and Slic3r will run the script with filename as the first parameter: `postprocessing.bat <input_filename>`. Then you can pass the filename to `Slic3rPostProcessing.exe` in `postprocessing.bat`. See my configuration: https://github.com/thuSkywalker/wiiboox-c300-slic3r-config/blob/master/gpx/gpx.bat

foreachthing:
@thuSkywalker thank you! :-)

alranel:
Actually, that limitation was in place on purpose, for security reasons. I didn't want people to be able to distribute a malicious config file containing, say, `rm -rf ~/*` thus executing arbitrary commands on other people's machines.  Passing only the name of a single executable mitigates this risk. In order to supply arguments to a post-processing script, people could just create a wrapper shell script.

alranel:
I wonder if we should add an option to enable this, and leave it disabled by default.

foreachthing:
I understand your concern. But do you have a means to pass parameters the save way?
As long as one can execute something, there's a risk of misuse. I _could_ add that malicious command to my post-processor...

nwneisen:
It looks like #4363 added the ability to pass parameters to a post-processing script.  Is this issue still open because of the enable/disable feature?

foreachthing:
Nah, I forgot about it and never closed it ... we're done here!

PR #4363, opened 2018-04-03
Post-Processing Script Parameters

TheThirdOne:
This is a solution to #4000.

I didn't realize this before I finished, but #4325 also solves the issue (it did remind me that \ is a poor escaping character for windows). However, that solution prevents ! from being using in any of the arguments or executable name. This one instead splits on whitespace and allows escaping whitespace so it doesn't prevent any filename from being expressed.

Examples:
`/path/to/executable` becomes `/path/to/executable` with the arg `outputfilename.gcode`

`/path/to/executable -arg -arg2` becomes `/path/to/executable` with args `-arg`, `-arg2`, and `outputfilename.gcode`

`/path/to/executable! with! spaces -arg -arg2` becomes `/path/to/executable with spaces` with args `-arg`, `-arg2`, and `outputfilename.gcode`

`/path/to/executable!! -arg -arg2` becomes `/path/to/executable!` with args `-arg`, `-arg2`, and `outputfilename.gcode`

The escaping is solved robustly so it should work exactly as programmers would expect (but with ! instead of \ because of Windows). But non programmers may not find it as obvious and people with existing post processing scripts that have whitespace in them may be surprised by their script breaking. 

I haven't testing on Windows, but I don't see why it shouldn't work.

### DIFF
--- a/lib/Slic3r/Print.pm
+++ b/lib/Slic3r/Print.pm
@@ -71,6 +71,34 @@ sub process {
     }
 }
 
+sub escaped_split {
+    my ($line) = @_;
+
+    # Free up three characters for temporary replacement
+    $line =~ s/%/%%/g;
+    $line =~ s/#/##/g;
+    $line =~ s/\?/\?\?/g;
+
+    # replace escaped !'s
+    $line =~ s/\!\!/%#\?/g;
+    
+    # split on non-escaped whitespace
+    my @split = split /(?<=[^\!])\s+/, $line, -1;
+
+    for my $part (@split) {
+      # replace escaped whitespace with the whitespace
+      $part =~ s/\!(\s+)/$1/g;
+
+      # resub temp symbols
+      $part =~ s/%#\?/\!/g;
+      $part =~ s/%%/%/g;
+      $part =~ s/##/#/g;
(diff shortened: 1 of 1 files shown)


foreachthing:
Just as well! :-)
Q: May I have spaces in args? Like `-full path to output file`...

TheThirdOne:
Yeah that works just fine.

`/path/to/executable option! with! spaces` becomes `/path/to/executable` with args  `option with spaces` and  `outputfilename.gcode`

lordofhyphens:
I will take a look at it tonight and make any recommendations as necessary. Thanks for the PR.

foreachthing:
Downloaded the "artifacts" from here: https://ci.appveyor.com/project/lordofhyphens/slic3r/build/1.3.0-master-1639/ and tested with Windows 7 x64 => works.

@lordofhyphens: as far as I'm concerned, you can close https://github.com/slic3r/Slic3r/pull/4325 and merge _this_ PR.

lordofhyphens:
About the only thing I could want would be a regression test (to make sure that if someone manages to break it again the future we notice immediately).

lordofhyphens:
Yeah, looks pretty good. Thank you @TheThirdOne .

PR #4372, opened 2018-04-07
I dunno lol; to discuss

ntfshard:
The main question is xs/src/xsinit.h file. A Clone class doesn't look legit but it looks like doesn't needed in dtor(tried - got test fails) but operator= seems to be lost to prevent memory leak

Changes in xs/src/libslic3r/PrintConfig.cpp are obvious(removed dups and trailing spaces)

I wish I had build a gui version but it seems needed some special wxwidgets version

Could you please double check the Clone class

### DIFF
--- a/xs/src/xsinit.h
+++ b/xs/src/xsinit.h
@@ -98,6 +98,7 @@ SV* perl_to_SV_clone_ref(const T &t) {
     return sv;
 }
 
+// // poor man's  std::ref(c++11) ?
 template <class T> 
 class Ref {
     T* val;
@@ -108,7 +109,8 @@ public:
     operator T*() const { return val; }
     static const char* CLASS() { return ClassTraits<T>::name_ref; }
 };
-  
+
+// poor mam's std::shared_ptr(C++11) ?
 template <class T>
 class Clone {
     T* val;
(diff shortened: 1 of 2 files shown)


xs/src/xsinit.h
```diff
@@ -118,6 +120,7 @@ class Clone {
     Clone(const T& t) : val(new T(t)) {}
     operator T*() const { return val; }
     static const char* CLASS() { return ClassTraits<T>::name; }
+    Clone& operator=(const T& t) { delete val; val = new T(t); return *this; }
```
bubnikv: This new copy operator is not going to be doing anything. The two wrappers are there to assist Perl to manage the C++ instances by its internal reference counting, so the life of the wrapper is:

1) Created by the XS module
2) Referenced by the Perl, possibly passed back to the C++ code by the pointer of the embedded class.
3) Destructed through Perl's own reference counter.

If there is any cloning being done by the Perl on the XS/C++ instances, then it is when the worker threads are created, see https://perldoc.perl.org/perlmod.html
and the xs/lib/Slic3r/XS.pm and search for sub CLONE_SKIP and  sub clone,
and also Slic3r.pm, sub thread_cleanup where most of the XS/C++ instances are just abandoned by the worker threads.

The whole concept of the Perl worker threads working with the XS/C++ instances, the worker threads being killed hard, and abandoning the XS/C++ data will certainly lead to memory leaks, and we at Prusa Research are now working on rewriting the complete backend into C++, so there will be no Perl worker threads running in Slic3r PE 1.41.

lordofhyphens:
Please don't mix code cleanup/refactor with other changes. @bubnikv is also correct about the lifetime of the wrappers. 

Closing.

ntfshard:
I see, thank you.
Also I tested values in assigment operator(aka operator= ) and all of them are NULL, so it looks legit to use default

PR #4388, opened 2018-04-19
"Reload from disk" - UI function overhaul

Oekn5w:
### Description
The function now tries to reload all referenced files (original object and added parts and modifiers) and asks the user how to proceed.
In the issue some opinions were given, that the function should discard all manual configurations (as 3mf files already contain them), and some that it should keep them (working with multiple exported STLs).

Additionally, warnings are thrown, if the reload doesn't work (original file deleted, object cut/ split) or added parts/ modifiers loose their reference to the path (modifier file deleted, less volumes in file -> the mesh is copied in that case).

### Changes breakdown by file
**Plater.pm**
function load_file: pass the correct object index
                           set variables in the volumes to determine their origin
function reload_from_disk: see description

**ObjectPartsPanel.pm**
set variables in the volumes to determine their origin (parts/ modifiers)

**Model.xsp/.hpp/.cpp**
adding needed variables to ModelVolume

### Testing
Testing turned out to be tedious, as the names had to be changed repeatedly, but I think I covered all relevant cases (major mesh change, different number of volumes, application of configurations)
The attached files were used.
[Multidomain_test_files.zip](https://github.com/slic3r/Slic3r/files/1929979/Multidomain_test_files.zip)


lordofhyphens:
image
[test_reload.zip](https://github.com/slic3r/Slic3r/files/1930615/test_reload.zip)

Disable auto-center parts (x,y).

Load 20mmbox.stl, and then add 20mmbox_2z.stl. You'll see it goes off into the corner (may be a bug, idk). Move it back. 

Click OK and try to reload. The modifier mesh gets kicked off into the ether again. :(

lordofhyphens:
Also, a "do this all the time and don't bother me again" checkbox would be a good idea (with the ability to set/unset it from File->Preferences), I think.

Oekn5w:
> Load 20mmbox.stl, and then add 20mmbox_2z.stl. You'll see it goes off into the corner (may be a bug, idk). Move it back.

Not a bug, the mesh of 2Z is way of center
grafik

> The modifier mesh gets kicked off into the ether again. :(

Well, it reloads the source again, and scaling and translation in the parts panel are directly applied to the mesh without history, no way of fixing this for now. (Full transformation matrices and quaternion rotation in a future milestone?)
I didn't want to create a new window for that pop up only, but it seems to be the best option, especially since in a separat window a new option can be added to copy the modifier meshes without reloading them.

I also thought about a don't ask me again. The preference panel seems to be kept pretty clean on purpose, but that might be the best option, as the issue showed different workflows regarding reloading.

PR #4403, opened 2018-05-06
Cppgui

EitanSomething:
How can you see the GUI written in C++ instead of the GUI in Perl


lordofhyphens:
Please don't open pull requests to ask questions. Ever. That is what the Issues tab is for.

It's nowhere near compete (you can't generate gcode with it yet at all). I also can't say it will work on Windows or OSX without crashing, so don't report failures about that to me: I will close and lock the issue immediately.

If you want to help with the port, go to my fork lordofhyphens/Slic3r and check out the cppgui branch there. Open any pull requests to get code into cppgui against that. Also open an issue on my fork to say what you are working on. Eventually everything is getting merged back here.

To build and run this, you need to have wxwidgets (I am using 3.1) built and the development files available. You also need CMake and and a C++ compiler that supports C++11.

Cd into the src subdirectory and run `cmake .` 

Then `make`. You can speed things up with `make -j4`. 

The resulting slic3r binary must only be run from this directory (I have not implemented a more failure tolerant way of finding Slic3r's resources directory yet).

If something fails to build, then you get to go figure out what you are missing and install it.

I build on Linux Debian (Sid) with gcc 6.3 and the build server is using gcc7.

If you are on Linux then you can use the .travis.yml file as a guide for what to install.

Issue #4407, opened 2018-05-10
Invalid mid-air supports!


Coder-256:
### Version
Slic3r 1.3.0 (finally! 3ba8e0f)

### Operating system type + version
macOS High Sierra 10.13.4

### Behavior
* **Issue**: Supports are generated in mid-air in my model. This makes the exported GCODE completely unprintable. ☹️
* **Steps**:
  1. Load the included config and STL.
  2. Rotate 90 degrees around the X axis. (sorry)
  3. Look at the `Preview` and `Layers` tabs.
* **Expected Result**: The bottom of the supports would touch the model.
* **Actual Result**: The supports are floating. (also, why do all supports have different infill?)
![screen shot 2018-05-09 at 8 43 16 pm](https://user-images.githubusercontent.com/10134823/39846443-ca10e63c-53c9-11e8-9d48-2544daca7740.png)

#### STL/Config (.ZIP) where problem occurs
[Slic3rBugReport.zip](https://github.com/slic3r/Slic3r/files/1989947/Slic3rBugReport.zip) (mind my lunatic design)

lordofhyphens:
One thing you can do to avoid the problem is edit lib/Slic3r/Print/SupportMaterial.pm and tweak PILLAR_SIZE and PILLAR_SPACING (lines 26 and 27).

It looks like the defaults don't play nice on this model.

You don't need to recompile to see the changes, just close and reopen Slic3r.

Coder-256:
What exactly do they do, and what should I set them to? Why is there no config option for this?

lordofhyphens:
@Coder-256 It's really easy to get weird results. 

"PILLAR_SIZE" is the X/Y dimensions of the support pillars generated, PILLAR_SPACING is how close  those pillars are allowed to be to each other. 

With your config, 5/15 seems to avoid the problem (which is probably from the pillars because the sloping area makes them too thin).

Coder-256:
You're right, that works. Thanks! I guess one possible way to fix this bug is to add those options to the config?

lordofhyphens:
I'd probably accept a pull request to add them, but the UI would need to hide the options when it's not pillars type. 

I am a bit busy to go in and add it myself (working on an UI overhaul).

Coder-256:
This is mostly fixed by #4410, so I'll leave it up to @lordofhyphens whether this issue should be closed, or if further steps should be taken (maybe implementing some kind of detection/warning for invalid pillar supports).

PR #4410, opened 2018-05-12
Pillar config

Coder-256:
Workaround for #4407. Let me know if I missed anything. I was unable to build Slic3r (`cpanm Wx::Perl::Packager` wouldn't work for me), so I was unable to test the GUI.


lordofhyphens:
Tests look like they referenced the constants before; they'll need to be updated to use the config.

Build.PL
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/perl
+#!/usr/bin/env perl
```
lordofhyphens: This change needs to be reverted.

Coder-256: Alright, I fixed that.

lib/Slic3r/Print/SupportMaterial.pm
```diff
@@ -22,10 +22,6 @@ use constant DEBUG_CONTACT_ONLY => 0;
 # increment used to reach MARGIN in steps to avoid trespassing thin objects
 use constant MARGIN_STEP => MARGIN/3;
 
-# generate a tree-like structure to save material
-use constant PILLAR_SIZE    => 2.5;
```
lordofhyphens: https://travis-ci.org/slic3r/Slic3r/jobs/377987179

These are being referenced in a few tests. Please update the tests to pull the default values of the config instead. (in the tests the config is usually initialized as $config to defaults and you can reference it as $config->pillar_size in those tests or $config->pillar_spacing)

lordofhyphens: On the other hand, it built so that's good :)

If you drop a copy of local-lib from a build package into your Slic3r source directory, you can use the following to just build XS and usually avoid having to do a full "perl Build.pl"

On Windows, open up Powershell and cd to the build source.
`Invoke-Expression "$(perl -Mlocal::lib=./local-lib)"` will configure that window appropriately, so you can go into the xs directory and run `./Build install` and it'll just compile Slic3r without going to get any of the Perl dependencies.

Coder-256: I am using macOS, but I managed to run those commands. How do I actually launch Slic3r after that though? Either way, I think I fixed the tests in 1bad4af.

lordofhyphens: On OSX, the magic is `eval $(...)`

You run Slic3r as normal (`perl slic3r.pl`), and you can run the tests as Perl scripts or use the prove command . It lasts until you close that terminal.

lordofhyphens: On OSX I highly recommend using perl brew if you aren't.

Coder-256: I don't really understand your instructions. `local-lib` from where? What commands do I run to get the GUI working, starting with checking out my branch?

I think the issue is that I can't figure out how to install wxWidgets. I have tried many different ways but none of them have succeeded (except installing through Homebrew, but then `Alien::wxWidgets` wouldn't recognize it). Could you please test for me that the UI works properly? I fixed the tests.

lib/Slic3r/GUI/PresetEditor.pm
```diff
@@ -939,6 +942,7 @@ sub _update {
     
     my $have_support_material = $config->support_material || $config->raft_layers > 0;
     my $have_support_interface = $config->support_material_interface_layers > 0;
+    my $have_support_pillars = $have_support_material && $config->support_material_pattern == 'pillars';
```
lordofhyphens: Argument "pillars" isn't numeric in numeric eq (==) at /home/lenox/slic3r-release/lib/Slic3r/GUI/PresetEditor.pm line 945.
Argument "rectilinear-grid" isn't numeric in numeric eq (==) at /home/lenox/slic3r-release/lib/Slic3r/GUI/PresetEditor.pm line 945.
Argument "pillars" isn't numeric in numeric eq (==) at /home/lenox/slic3r-release/lib/Slic3r/GUI/PresetEditor.pm line 945.
Argument "pillars" isn't numeric in numeric eq (==) at /home/lenox/slic3r-release/lib/Slic3r/GUI/PresetEditor.pm line 945.

xs/src/libslic3r/PrintConfig.cpp
```diff
@@ -1502,6 +1502,24 @@ PrintConfigDef::PrintConfigDef()
...
+    def->label = "Pillar size";
+    def->category = "Support material";
+    def->tooltip = "Size of the pillars in the pillar support pattern";
+    def->sidetext = "mm";
+    def->cli = "support-material-pillar-size=f";
+    def->min = 0;
+    def->default_value = new ConfigOptionFloat(2.5);
+
+    def = this->add("support_material_pillar_spacing", coFloat);
+    def->label = "Pattern spacing";
```
lordofhyphens: Needs to be "Pillar spacing"

lib/Slic3r/GUI/PresetEditor.pm
lordofhyphens: Use `eq` instead of `==` and it works.

xs/src/libslic3r/PrintConfig.cpp
Coder-256: Fixed in c568db2d.

Coder-256:
Have the fixes worked for the `==`/`eq` and the GUI label?

lordofhyphens:
@Coder-256 Looks good, thanks for the work.

PR #4419, opened 2018-05-17
Extended GUI

foreachthing:
PLEASE let me know what you think!! 
Maybe other commands can be in the context or in the toolbar made more accessible... ?!
The first PR got shot down - so I try again with everything being optional ;-)
_I'm getting tired of merging my stuff into the latest build - so I hope this gets merged._

I use the same commands a lot. And since those are "buried" in submenus, I brought them up and to the toolbar (only a few) as well.

Everything is optional!

image

- Default color scheme (Toolbar and Context):
image

This is for `Solarized` _only_:

- Big icons
image

- Normal sized icons:
image

- Context:
image


lordofhyphens:
For bookkeeping, was there a related feature request for these changes?

dwillmore:
What is the purpose of this?  I can't tell you what I think of it unless I know what it's supposed to do.

My only comment at this point is, please don't use color coding for the axes (or don't use it exclusively).  I'm colorblind and I will never be able to tell the difference between the X and Y buttons if they are red and green.

VanessaE:
@dwillmore  makes a good point, but keep the colors - red/green/blue has been the standard for X/Y/Z for G*d knows how long.  Put letters on the icons (plus a "180" for those icons) to address the colorblindness compatibility.

foreachthing:
Ok, here goes nothing!
@lordofhyphens : no, no issue created :-) ! I just wanted to share my work. Since I'm a CAD-user and my (and other's) models are all over the place (and X is mostly Z in reality), I need to rotate, scale or mirror them quickly. So, I don't have to go through sub-menus anymore. Mostly it's 90° in X or 90°/180° in Y ...

@dwillmore; well, you might think of it as shortcuts to _my_ most used commands. Like I wrote above, I don't want to go through sub-menus to rotate a part around X by 90°. So I put it to the toolbar.
The "default" color theme, I didn't touch those icons. Well, ok, I added the mirror icon. With "Solarized", I've added a lable - I like that because I can see the icon right away.
I cannot imagine how you perceive my solarized icons. If you have advice for me how to improve those icons - I'm all ears!

@VanessaE: my english's a bit rusty; what do you mean by _'..should be exempt from "solarize" also.'_?
I'll keep the colors, no question! I understand not to use the same icon multiple times, but a dot tells me nothing. An arrow - ah: rotation! Labeling the icons might be helpful but those things are soooo tiny!!

Long story short: I hope to have made something others find useful. Sure, there is always room for improvements! :-) That's why I wanted to know what _you_ think.

On the other hand: Those changes/improvements are dear to me! If nobody thinks of them as a worthy addition, I shall withdraw my PR - no hard feelings!


So, beside labeling tiny icons, what else would you change to be worthy of merging?

VanessaE:
By "exempt" I mean, keep the colors pure :smiley:  in two of the screenshots, green looks like a sort of dark yellow. I assumed that was related to the "solarize only" comment.

VanessaE:
As for the uniqueness of the icons, you touched on in in the "solarize only" screenshots - rotate 180° has the mark going half-way around, while the others don't.  In the opening screenshot, that difference doesn't exist.  I might suggest the same style of arrows as in that image, but adjusted to show the three amounts of rotation that's obvious in the other images.

VanessaE:
The only other change I would suggest is to move "delete all" away from "arrange".

dwillmore:
@foreachthing Cool, 'scratching your own itch' is a good motivation for writing code.

This might make things harder, but most programs that have different sets of toolbars have a View->Toolbars menu with check/uncheck selectors for the different toolbars.

For colorblind people, it helps to use fully saturated colors.  Avoid red/green being next to each other--though that's unavoidable here.  The axis on the buttons should be in something that contrasts both the background and the color of the icon.  I'd suggest black.  X, Y, and Z are quite different looking from each other, so the text font can be very small and they'll still be recognizable.

That may be beyond the scope of what you're doing.  Maybe I could contribute a colorblind icon theme?

The solarized icon theme is very much not enjoyable by me. :)  Sorry.

FWIW, I would turn this tool bar on.  I do a lot of rotations like this and I would find it helpful.  Now, if we could bind key combinations to these menu selections, that would remove my perception of the need for this feature, but @lordofhyphens has pointed out that the widget set doesn't support that. :(

foreachthing:
>  Maybe I could contribute a colorblind icon theme?

Yes, please! :-) I shouldn't be too hard to add "color-blind"-theme. If you're at it, please suggest all other UI colors as well...

>The solarized icon theme is very much not enjoyable by me. :) Sorry.

The default theme was my motivation for the Solarized. I dind't like how Slic3r looked like a Win311 installer - no offense! So, I can understand that you don't like Solarized :-)

Although, I thought that the Solarized theme was also "working" for color blind people?! Maybe I misread or misunderstood http://ethanschoonover.com/solarized. Or, used the colors the wrong way - anything's possible.

sanchosk:
Not sure it counts, but I would love the buttons in top menu. I like the proposed design quite a lot.

foreachthing:
@sanchosk sure it counts - too me! :-) Thank you - _und hopp Schwiiz_!
I'm still working on it. So it will be a while before the next PR.
There's a lot of code changes behind this. I'm making nearly everything optional.
As you can see, even now, there are conflicts in plater.pm. Those changes cannot be merged automatically. So, with every new build, I need to redo my mods. And this gets quite boring...

Bluelinegecko:
Also, please adjust the spacing on the UI windows or make them re-sizable by the user. I have a lot of different profiles for print settings/filament settings, etc. and it is impossible to chose some of them because I can't read the entire profile in the drop down box. I don't know if this is the same in linux/Mac, but for Windows users its a pain. There is a TON of unused real estate on the data-entry side of the screen. i would think giving a few extra pixels to the left sidebar would be a relatively easy change.

screen

lordofhyphens:
There you go, @foreachthing. 

I merged in Rotate to Face.

lordofhyphens:
@Bluelinegecko in the future please don't add unrelated issues/comments to someone else's PR or issue; it just means that it is very likely that I sort out the main issue and then your comments get forgotten.

lordofhyphens:
@foreachthing had to fix one more thing: you neglected to use $Slic3r::var->() to wrap the path of the toolbar icons in the AddTool calls. I don't know why it worked for you. 

It's fixed and merged though, so enjoy.

Bluelinegecko:
> @Bluelinegecko in the future please don't add unrelated issues/comments to someone else's PR or issue; it just means that it is very likely that I sort out the main issue and then your comments get forgotten.

Sorry about that. I thought the thread was about asking for feedback and improvements to the UI. My mistake.

Issue #4406, opened 2018-05-09
Gyroid infill bug?


scott0122:
### Version
1.3.0-dev, Slic3r-master.2018.05.07.1676.8a93fbb.64bit

### Operating system type + version
Windows 10 64bit

### Behavior
My infill setting
![snag-0001](https://user-images.githubusercontent.com/39105828/39788795-b5976be2-535e-11e8-9327-f1a64cd22c93.png)
When I place model on this direction
![snag-0002](https://user-images.githubusercontent.com/39105828/39788806-c0b596fc-535e-11e8-8729-0f6669d3526c.png)
Infill on some layer seems broken
![snag-0003](https://user-images.githubusercontent.com/39105828/39788814-cbd2c424-535e-11e8-9cb9-95309d3943a2.png)
![snag-0004](https://user-images.githubusercontent.com/39105828/39788818-ce5b595e-535e-11e8-9719-ca96e2b80122.png)

When I place model on this direction, the problem is gone.
![snag-0005](https://user-images.githubusercontent.com/39105828/39788831-d9e7c0fa-535e-11e8-85a3-7898385495bd.png)
![snag-0006](https://user-images.githubusercontent.com/39105828/39788833-dc075f30-535e-11e8-8e82-5e79604ca482.png)
![snag-0007](https://user-images.githubusercontent.com/39105828/39788835-de246466-535e-11e8-9727-34d949cd3477.png)

#### STL/Config (.ZIP) where problem occurs
[Body1.zip](https://github.com/slic3r/Slic3r/files/1985977/Body1.zip)

foreachthing:
@scott0122 I think you need to go over here: https://github.com/prusa3d/Slic3r/issues/

Also, see here: https://github.com/prusa3d/Slic3r/issues/824
And maybe this one, too: https://github.com/prusa3d/Slic3r/issues/825

bubnikv:
Yes, it is  prusa3d#825 <https://github.com/prusa3d/Slic3r/issues/825> and
there is the fix included.

foreachthing:
@bubnikv would you be kind enough to PR your changes to Slic3r as well? Thanks!

supermerill:
I'm not at home this week. I can't test a slic3r build right now.
I should have PR the bubnikv changes, but as it's a bit time consuming, i waited the next release of slic3rPE. As it's now out, I promise to test it & PR as soon as possible. If someone want to Merge back the changes & fix before me, no problem.

[...]

PR #4425, opened 2018-05-22
merge FillGyroid to slic3rPe state

supermerill:
contains a bugfix for #4406
thanks to  bubnikv for the refactoring

note: for the "overlap", this pr does nothing, the bug is somewhere else in slic3r.

Iv'e tested with some stl, it slice and export gcode. But i didn't had the time to test a real print with the generate gcode. But it shouldn't be a problem.


lordofhyphens:
Refactor looks good to me. Appreciate the work with the PR @supermerill and @bubnikv

PR #4429, opened 2018-05-25
Request: option to disable the filament sensor from the filament profile

giudiceanto:
Hi, I want to premise this is my first request on github, so excuse me if I use a wrong form/terminology

Problem: For the first time I had an issue with a specific filament (Das Filament black pet-g) that caused the filament sensor to be triggered randomly.

Request: would it be possible to have a flag to check\uncheck in the filament profile to activate\deactivate the filament sensor, since the function is strictly related to it?

I understand the funcion will be effective only using the gcode (so no during the filament loading\unload)


bubnikv:
I think you mistakenly posted a pull request to a wrong slicer. It looks
like it belongs to the prusa3d fork.

lordofhyphens:
Do not open pull requests when you mean to make feature requests! 

As to your question,  Slic3r supports custom gcode per-filament for start/stop, so that will cover single-material prints.

For multimaterial prints, you'll currently need to work out something in the custom tool change gcode.

lordofhyphens:
@giudiceanto  https://github.com/slic3r/Slic3r/issues if you want to open a feature request against mainline Slic3r.

PR #4430, opened 2018-05-27
Allow to set shell thickness for top/bottom layers

### LINKED ISSUE #1405: Feature Request: Allow to set shell thickness

Samir55:
Fixes #1405 
I am still working on it (what's left is some refactoring, choosing a better name for this option and testing).
@lordofhyphens please review.


Samir55:
@lordofhyphens, I think it's ready, please review ;)

PR #4433, opened 2018-06-01
Size of the wipe tower

jpatenet:
With the last 1.39 version, it was possible to adjust te size of the wipe tower in length and width. Now it's not longer possible, if I reduce the length, th width is increased ... And the wipe tower is larger than the previous one.
I suggest to restore the previous wipe tower parameters


bubnikv:
You pushed the pull request into an incorrect repository.

This is the correct one:
https://github.com/prusa3d/slic3r

lordofhyphens:
Closing.

Issue #4464, opened 2018-06-28
"Move t bed center" somehow not properly working


Tinchus2009:
### Version
354b623

### Operating system type + version
win10 64b

### Behavior

Load any model, then select it, and chose from menu "Move to bed center", the object will be placed at the bed center, then slice (selecting the .preview tab). You will see that the sliced preview will not be in the bed center but in the original object location. The 2D view showd it in the bed center, the preview doesnt, and the layers view shows the layers in the object's original position (not in the bed center)

![imagen](https://user-images.githubusercontent.com/14299602/42009982-117e70b0-7a63-11e8-94c4-8acc01ac38eb.png)

Moved to bed center:

![imagen](https://user-images.githubusercontent.com/14299602/42010011-31db5562-7a63-11e8-911d-0f95b7acb59e.png)

Then the sliced preview:

![imagen](https://user-images.githubusercontent.com/14299602/42010024-3aadad3e-7a63-11e8-9645-3782491d5d5d.png)

lordofhyphens:
@Tinchus2009 "latest dev version" doesn't mean much, please include at least the short commit hash.

If you're using files off of dl.slic3r.org/dev/win, it's the number right before "32bit" or "64bit" in the filename. 

If you have a git checkout, `git rev-parse --short HEAD` will give the string.

Tinchus2009:
Sorry, it is this one:


354b623
-- | -- | -- | --

lordofhyphens:
Fixed via #4465

PR #4465, opened 2018-06-29
Actually update the print when moving to center (Fixing #4464)

Oekn5w:


Oekn5w:
Tagging #4464

lordofhyphens:
Thanks!

PR #4472, opened 2018-07-05
thinwall improvements

supermerill:
* no more strange turn at end => fusion of voronoi end lines when they are near enough
* anchoring inside the perimeters => extends the lines end inside the perimeter to join it. A gui parameter can be created to adjust it if needed (currently, it extends to meet the center).
* conservation of too small segments of thin walls when they are crossing each other => modification of the fusion algorithm to select the most strait path instead of a random one


lordofhyphens:
Merging, will keep an eye as to whether or not this causes other problems.

supermerill:
Just a note: the changes are in a function which is also used by gaps.
If gaps behaving weirdly, it may be useful to split the function for the 2 needs (thinwalls vs gaps).

PR #4483, opened 2018-07-11
Porting Slic3r Support

Samir55:
## Todo
- [x] Port the 4 remaining functions.
- [ ] Refactoring.
- [ ] Port the remaining tests.


src/test/libslic3r/test_support_material.cpp
```diff
@@ -0,0 +1,98 @@
...
+                    if (support_z[i] - support_z[i - 1] <= 0) check = false;
+                REQUIRE(check);
+            }
+
+            THEN("No layers thicker than nozzle diameter") {
+                bool check = true;
+                for (size_t i = 1; i < support_z.size(); ++i)
+                    if (support_z[i] - support_z[i - 1] > print.config.nozzle_diameter.get_at(0) + EPSILON)
+                        check = false;
+                REQUIRE(check);
```
lordofhyphens: It's a "small" issue, but please be slightly more verbose here:

`REQUIRE(check == true);` instead to make it easier to understand.

Samir55: OK :)  @lordofhyphens

Samir55:
@lordofhyphens , For the remaining tests (6), they are using export function written in GCode.pm.  What's the equivalent function in the c++ libslic3r?

src/CMakeLists.txt
```diff
@@ -192,7 +193,7 @@ set(UI_TEST_SOURCES
 set(SLIC3R_TEST_SOURCES
     ${TESTDIR}/test_harness.cpp
     ${TESTDIR}/libslic3r/test_trianglemesh.cpp
-)
+        test/libslic3r/test_support_material.cpp)
```
lordofhyphens: This should be `${TESTDIR}/libslic3r/test_support_material.cpp`.

src/test/libslic3r/test_support_material.cpp
```diff
@@ -0,0 +1,296 @@
+//#include <catch.hpp>
```
lordofhyphens: This should work now, btw.

xs/src/libslic3r/PrintObject.cpp
```diff
@@ -1095,4 +1095,55 @@ PrintObject::_infill()
     this->state.set_done(posInfill);
 }
 
+SupportMaterial *
+PrintObject::_support_material()
```
lordofhyphens: Any reason why you're using leading underscores with the names? If you don't want them to conflict with the things exposed in the perl half (and don't want to write XSP entries for them), wrap them in #ifndef SLIC3RXS ... #endif // SLIC3RXS

xs/src/libslic3r/SupportMaterial.cpp
```diff
@@ -0,0 +1,1225 @@
...
+    // Loop through object's top surfaces. TODO CHeck if the keys are sorted.
+    for (auto &top_el : top) {
+        // Keep a count of the interface layers we generated for this top surface.
+        int interface_layers = 0;
+
+        // Loop through support layers until we find the one(s) right above the top
+        // surface.
+        for (int layer_id = 0; layer_id < support_z.size(); layer_id++) {
+            auto z = support_z[layer_id];
+            if (!z > top_el.first) // next unless $z > $top_z;
```
lordofhyphens: Compiler warning: the `!` is only being applied to `z`, so the expression looks like `if ((!z) > top_el.first)` and not `if (!(z > top_el.first))`.

lordofhyphens:
@Samir55 I'm in the middle of writing it, as well as Slic3r::Test::init_print and Slic3r::Test::gcode. I hope to get it done by the end of my day today and submit a PR against mainline slic3r.

I've merged your branch as it is right now into cppgui (you implemented the extensions to PrintObject, which I needed).

Samir55:
Thanks, @lordofhyphens and, I will resolve the comments.

lordofhyphens:
BTW, the tendrils running out from the gcode export are even more numerous than I saw.

src/test/libslic3r/test_support_material.cpp
```diff
@@ -0,0 +1,295 @@
...
+    model.align_instances_to_origin();
+
+    // Create Print.
+    Print print = Print();
+    vector<coordf_t> contact_z = {1.9};
+    vector<coordf_t> top_z = {1.1};
+    print.default_object_config.support_material = 1;
+    print.default_object_config.set_deserialize("raft_layers", "3");
+    print.add_model_object(model.objects[0]);
+    print.objects.front()->_slice();
```
lordofhyphens: I suggest calling ->slice() instead of _slice().

Samir55: Ok, LoH I will replace it :),

Samir55:
closing for now and opening a PR against LoH/cppgui fork

PR #4484, opened 2018-07-11
Preliminary merge of cppgui

lordofhyphens:
Merge cppgui changes and tests to get the libslic3r side of changes in while maintaining existing build system.


lordofhyphens:
CppGui does not function with this PR. The primary use of this PR is to provide the test interface and extend libslic3r so that non-perl users do not need to reimplement parts of the interface that are currently coded in XS.

PR #4490, opened 2018-07-21
Variable support layer height

omercanvural:
Hi there,

I have been using adaptive layer height for some time now and I think I might have an idea to improve it. I tried to google if anyone came up with it but could not get much results or maybe it is already a thing and i don't know the term for it.

Anyway, long story short, lowering layer height increase both quality and print time dramatically.

My solution is to lower perimeter layer height but keep the infill layer height at max possible printer allows. Resulting with much less travel and print time per layer increasing print quality but not increasing the print time by that much.

Some sample numbers; 0.3 mm nozzle max layer height is 0.2 mm lowest possible layer height is 0.01 mm

so let's say adaptive slicing decides to use 0.04 mm as layer height, it should print 4 layers of perimeters only, at 5th layer it should print perimeters with extrusion calculated for 0.04 mm perimeters and start infill after perimeter with extrusion calculated for 0.2 mm layer thickness and repeat this for every 5 layers. This will save the Z axis movement layers will match and infill will be strong. Print time will only increase for perimeters.

I believe I can formulate the layer height / infill height calculation for this. But I am not a coder so i can not branch an open source slicer without help.

Would this worth looking at? Has it been done before?

I believe just increasing print time for perimeters and keeping infill time same would be way much more acceptable and would become new default for most printer users.

Looking forward to read your comments.

Thank you for your time already :)


bubnikv:
You have provided a pull request to a wrong repo. Please provide the pull
request to the Prusa3D fork. Thanks.

platsch:
Hi there, sorry for my late answer, I was somewhat off the grid for a while.

@omercanvural I'm not sure about the Prusa version, but the mainline Slic3r does exactly what you describe. Just set the `combine infill every n layers` variable to a high number (e.g. 5) and Slic3r automatically combines as many infill layers as possible, but not more than 5. This also works for automatically generated adaptive layers.
If it doesn't work as described, please open a new issue :-)

PR #4495, opened 2018-07-30
Fix int

lordofhyphens:
Use int64 on Windows; long is already 64-bit on Linux


lordofhyphens:
Builds on my dev system and on Linux, should be good to test.

PR #4504, opened 2018-08-07
First layer print order

AndWik:
If the bottom layer has small holes (for screws or similar) their perimeters easily came loose as they have nothing to attach to, except for the bed.
If the print order would be: Outer perimeter, Bottom layer and lastly Hole perimeters outside-in the possibility for the hole perimeters to attach have a far better chance.

My five cents
Mr Anders Wikström
Sweden


lordofhyphens:
Go to https://github.com/slic3r/Slic3r/issues/new and report the issue there. __READ__ the text present before writing your issue and make sure you have all of the required information.

lordofhyphens:
Closing because this isn't actually a pull request.

lordofhyphens:
@AndWik what is it about the GitHub interface that leads you to open a pull request instead of an issue? 

I'd really like to know because this happens with different people about twice a month and I'd like to at the very least see if there's a way I can redirect people to the correct page and/or make a bug report against GitHub itself.

AndWik:
OK!

It is not a real issue and maybe hard to do. There are always a way to go around adhesion issues and this is just an idea of making Slic3r the best slicing software ever. This is why I did write it here.

Thanks for your input!
Mr Anders Wikström

lordofhyphens:
@AndWik I appreciate the ideas and feature requests, but it really needs to go to https://github.com/slic3r/Slic3r/issues/new instead of here.

AndWik:
That's fine and now I know :-)

PR #4515, opened 2018-08-21
FIX: Forward declaration of std::ofstream caused CMake to abort

ElectroQuanta:
**Error**: The forward declaration of std::ofstream (used in export_gcode) caused the error below:

`/Slic3r/xs/src/libslic3r/Print.cpp: In member function ‘void Slic3r::Print::export_gcode(const string&, bool)’:
/Slic3r/xs/src/libslic3r/Print.cpp:723:36: error: variable ‘std::ofstream outstream’ has initializer but incomplete type
     std::ofstream outstream(outfile);`

**Solution**: Include `<fstream>`

### DIFF
--- a/xs/src/libslic3r/Print.cpp
+++ b/xs/src/libslic3r/Print.cpp
@@ -9,6 +9,7 @@
 #include <algorithm>
 #include <boost/filesystem.hpp>
 #include <boost/lexical_cast.hpp>
+#include <fstream>
 
 namespace Slic3r {
 


lordofhyphens:
Weird that it doesn't do this with the systems I've built with.

ElectroQuanta:
Hi, I can try to provide some light on that :)
I'm using Ubuntu Mate 16.04 on a VM (with pretty much everything outside of the box) with:

- CXX compiler: *c++* (version 5.4.0)
- CMake (version 3.12.1)

If you want, I can provide the *CMakeCache.txt* to assist on the enlightment.

lordofhyphens:
It was just weird. It didn't pop an error on clang (macOS) or on my home system (Debian sid with g++-8 or g++7) or on Travis (

Your patch makes sense, so I merged it :)

lordofhyphens:
Definitely please leave feedback for the C++ build toolchain in the future and pull requests :) 

It is considered experimental.

PR #4524, opened 2018-08-30
FIX: check bounds for fill->min_spacing causing fill_density<100 to abort

ElectroQuanta:
`Fill->min_spacing` was being initialized to 0, due to the `infill_extrusion_width`being set to 0 in the config file.
- As such, I defined a lower bound value to be positive (>0), in spite thinking that this should be at least the size of `nozzle_diameter`.
- Now, I don't get any exceptions as discussed [here](https://github.com/slic3r/Slic3r/issues/4509#issuecomment-417147752). 

Tests
-- 
For convenience, I had the result of some tests:
- `Infill_extrusion_width = 0.5`:
imagem
- `Infill_extrusion_width = 0.1`:
imagem
- `Infill_extrusion_width = 2.1`:
imagem
- `Infill_extrusion_width = 0.01`: Was taking too long.

This lasts two values highlight the importance of convenient bounding, but for now, I will leave as it is, as it suits my needs.

### DIFF
--- a/src/slic3r.cpp
+++ b/src/slic3r.cpp
@@ -154,12 +154,14 @@ main(int argc, char **argv)
             boost::nowide::cout << "File exported to " << outfile << std::endl;
         } else if (cli_config.export_svg) {
             std::string outfile = cli_config.output.value;
-            if (outfile.empty()) outfile = model.objects.front()->input_file + ".svg";
-            
-            SLAPrint print(&model);
-            print.config.apply(print_config, true);
-            print.slice();
-            print.write_svg(outfile);
+            if (outfile.empty()) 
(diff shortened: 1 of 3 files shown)


ElectroQuanta:
Yeah, I also thought that proper bounding was the way to go :)
I just didn't know the proper way to bound it :)

The way I see it (and I could be wrong), but if the `fill_spacing`is less than `nozzle_diameter`you'll get overlapping on the filling, and thus, the bounding suggested would not suffice :)

But, like you said, if the problem is upstream (`Flow`) and you already identified the need to refactor it, for me it's good, because like I said, it suits my needs for now :)

And thanks, for further investigating the issue :)

Issue #4511, opened 2018-08-14
Support loading deflated AMF files


Artefact2:
Using version 65bcaff (AppImage).

Trying to load a deflated AMF file fails (“Error while opening foo.amf: The supplied file couldn't be read because it's empty”). Inflating it by hand works:

```
Silmeria ~ % file foo.amf                                                               
foo.amf: Zip archive data, at least v2.0 to extract
Silmeria ~ % mv foo.amf foo.amf.zip
Silmeria ~ % unzip foo.amf.zip 
Archive:  foo.amf.zip
  inflating: foo.amf                 
Silmeria ~ % file foo.amf
foo.amf: XML 1.0 document, ASCII text
```

It would be nice if Slic3r could do it on its own, as some programs export compressed AMF files by default (looking at FreeCAD for instance). Would also be a great step towards usability, as the current behavior seems to indicate the problem comes from the file, not from an unsupported feature in Slic3r.

Regards

lordofhyphens:
The 3MF (TMF.cpp) loader does the zip archive work.

It shouldn't be too difficult to adapt this to AMF if someone wants to give it a go. 

As there isn't a hard requirement for the archived filename (other than it's one file), ZipArchive also needs to be extended to fetch the name(s) of anything in the archive (and reject if the file count > 1).

Here's a code snippet that retrieves the filenames with the underlying archive miniz. 
https://github.com/richgel999/miniz/issues/38

PR #4528, opened 2018-09-02
Deflate amf

lordofhyphens:
Implements #4511 with tests.


lordofhyphens:
@Artefact2 can you give this a spin and see how it works for you? If you find issues with loading, let me know and I can extend the tests to cover. 

The current known limitation is that you are only allowed 1 deflated file in the archive, but it can be in a subdirectory.

Lucki:
Sadly I wasn't able to import a quick .amf test export from FreeCAD with your changes. Slic3r hangs at _Processing input file…_ forever and ignores any further user input.
Bildschirmfoto-20190323010525-332x118

PR #4535, opened 2018-09-06
Auto-determine right pane button height.

oliof:
On HiDPI screens, the button height set to 30px will cause buttons of limited height with parts of the text cut off. Setting the height to `-1` seems to fix this so the buttons work fine.

### DIFF
--- a/lib/Slic3r/GUI/Plater.pm
+++ b/lib/Slic3r/GUI/Plater.pm
@@ -225,10 +225,10 @@ sub new {
     }
 
     # right pane buttons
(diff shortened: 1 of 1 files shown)


lordofhyphens:
From the wxSize docs:

----

size | Button size. If the default size is specified then the button is sized appropriately for the text.
-- | --

----

What this means is that the height can be inconsistent with this patch.

Related:
https://groups.google.com/forum/#!topic/wx-users/GE8snC1zwPc

lordofhyphens:
Looks like it shouldn't cause an issue on lower DPI screens.

lordofhyphens:
Thank you.

PR #4556, opened 2018-10-03
increasing fan in overhans and little section printing

MASPPrototipi:
I think that a function that it permise you tu increase fan where the extruder print in overhang.
For exemple make some limits where behind these the fan increase proportionally
0-50° %fan = %std fan
51-65 %fan= %std fan + 25% ( or select directly the % fan that you want)
66-75 %fan= %std fan + 40% ( or select directly the % fan that you want)
I think that this element can increase the printing quality.
You can also do the same thing when you print in small sector (like holes, angles or thin wall)
if there is a same selector for any kind of printing times you prevent little localized overheating.
Thank You! Great Work with the 1.41.1
I'm printing with a Zortrax M200


lordofhyphens:
Go to https://github.com/slic3r/Slic3r/issues hit new issue. 

Closing because again this is in the wrong place.

lordofhyphens:
ThIs is also not Prusa3d fork, but it is a reasonable request for us

Issue #4569, opened 2018-10-28
Placeholder not working?


hoh61:
**Describe the bug**
For a filament specific command I need to know to which extruder this filament is assigned. Unfortunately I failed up to now to find the suitable placeholder. 

From the list in the cartesio wiki I assumed that the "filament_settings_id" could be a suitable placeholder, but this placeholder is delivering nothing.

Is there any placeholder I can use for this?

; Filament gcode
M214 T[filament_settings_id'] ...

**Desktop (please complete the following information):**
 - Linux
 - Version 4.4.0-66-generic
 - generated by Slic3r 1.3.1-dev (Build )

lordofhyphens:
Can you upload a CTRL-E export so I can verify this with your config? Thank you.

lordofhyphens:
Which custom gcode are you trying to use this with? There's a few that I suspect aren't valid in Start or End Gcode, but would be valid in before/after layer gcode or the toolchange gcode.

lordofhyphens:
What I have been able to determine is that it's a "legit" Slic3r setting, but it's being set to "". That's also what it is in the config. 
````
Placeholder: DEBUG: Replacing filament_settings_id with ().
Placeholder: DEBUG: Replacing filament_settings_id with ().
Placeholder: DEBUG: Replacing filament_settings_id with ().
Placeholder: DEBUG: Replacing filament_settings_id with ().
Placeholder: DEBUG: Replacing filament_settings_id with ().
Placeholder: DEBUG: Replacing filament_settings_id with ().
Placeholder: DEBUG: Replacing filament_settings_id with ().
Placeholder: DEBUG: Replacing filament_settings_id with ().
Placeholder: DEBUG: Replacing filament_settings_id with ().
Placeholder: DEBUG: Replacing filament_settings_id with ().
Placeholder: DEBUG: Replacing filament_settings_id with ().
Placeholder: DEBUG: Replacing filament_settings_id with ().
Placeholder: DEBUG: Replacing filament_settings_id with ().
Placeholder: DEBUG: Replacing filament_settings_id with ().
Placeholder: DEBUG: Replacing filament_settings_id with ().
Placeholder: DEBUG: Replacing filament_settings_id with ().
Placeholder: DEBUG: Replacing filament_settings_id with ().
Placeholder: DEBUG: Replacing filament_settings_id with ().
Placeholder: DEBUG: Replacing filament_settings_id with ().
[...]

PR #4577, opened 2018-11-06
Add [filament_extruder_id] and revert work done for [filament_settings_id]

lordofhyphens:
Add a placeholder [filament_extruder_id], which is valid in Filament custom gcode and refers to whatever extruder it is loaded into. 

Also revert filament_settings_id.

#4569


lordofhyphens:
Merging as it meets the needs of the requestor @hoh61

PR #4580, opened 2018-11-06
Build on OSX with TravisCI

alranel:


package/osx/patch-wxwidgets.diff
```diff
@@ -0,0 +1,46 @@
+diff --git src/osx/carbon/dataobj.cpp src/osx/carbon/dataobj.cpp
```
lordofhyphens: Why is patching wxWidgets necessary?

package/osx/travis-build-main.sh
```diff
@@ -0,0 +1,24 @@
...
+#perl ./Build.PL
+
+# Only recompile Wx if it's not already there
+if !(perl -MWx -e1 2> /dev/null); then
+    # our patch-wxwidgets.diff assumes Alien::wxWidgets installs wxWidgets 3.0.2
+    PERL_USE_UNSAFE_INC=1 \
+        CXX="clang++ -mmacosx-version-min=10.12" \
+        CPPFLAGS="-mmacosx-version-min=10.12 -D__ASSERT_MACROS_DEFINE_VERSIONS_WITHOUT_UNDERSCORES=1" \
+        LDFLAGS="-mmacosx-version-min=10.12" \
+        cpanm --local-lib local-lib Alien::wxWidgets --reinstall \
```
lordofhyphens: Given the time it can take to build Alien::wxWidgets, it may be more prudent to precache the build on a system that is "close enough" and just upload it (similar to how I did it for both Linux and Windows) and the other --gui prerequisites.

lordofhyphens: There is a 10 minute "build with no log output" limit by default with travis; and I fear that trying to build/configure the entire stack will continually run afoul of these limits.

Additionally, for the primary deployment purposes, it should not be necessary to build all the entire stack on Travis.

package/osx/patch-wxwidgets.diff
alranel: Because wxWidgets 3.0.2 doesn't compile on recent macOS. It includes QuickTime.h which is not present anymore. Another option would be to make Alien::wxWidgets download 3.0.4 but I think that's more work.

package/osx/travis-build-main.sh
alranel: I'm trying to address the Travis limits. Basically, the long build time is only necessary for the first build, because Travis is caching local-lib for next builds.

alranel: Travis docs say that the default branch cache is used for all branches. I think the only problem is that cache is invalidated after 28 days of inactivity. I think the solution is to add our own caching to a remote server: similar to what you did, but created/updated by the Travis script itself without the need to prepare it manually in advance.

alranel:
Done.

This Pull Request provides:
* updated code signing on OS X
* Travis builds/deploy for OS X (builds are tested to work also on 10.11 and 10.12)
* support for using perl, wxwidgets and boost from Homebrew
* refactoring of the Travis config file (scripts are split into platform-specific directories)

Issue #4294, opened 2018-02-09
Feature request: manually set small radius (perimeter) for other parameter


ArtemKuchin:
Multiple parameter use "small radius" to take effect. For example, there is separate setting for small perimeter speed. But small perimeter radius is hard coded. Please, add a manual setting for small radius perimeters. The idea is that small radius is different for different mechanics. For better quality or lower end machines small radius is probably good to set a lot higher.

gennartan:
Hello, I got this!!
I just need to know where I should add this option in the slicer. Do you think under "vertical shell" in the page "Layers and perimeters" or under the option "small perimeter speed" in the page "speed"
The small perimeter length option is only used for the computation of the speed for the small perimeter, that is why I would place this option in the speed section.

ArtemKuchin:
I think logical place for it is the same tab where the speed for small radius is set, so everything willl be in one place.  So, i think it should be on the "speed" tab.

gennartan:
Thank you for this confirmation.
I created a new pull request at https://github.com/slic3r/Slic3r/pull/4581

PR #4581, opened 2018-11-09
add the small perimeter length parameter

gennartan:
Implements #4294


lordofhyphens:
A lot of the "changes" are whitespace changes. Please revert those changes and update the PR. 

If you want to commit a separate PR for code beautification, go ahead, but please don't mix functional changes and completely unrelated aesthetic changes.

lordofhyphens:
@gennartan Please provide a general description of a test for this feature; as it stands, there's no way to tell if this code breaks in the future (go ahead and add it just to src/test/libslic3r).

src/test/inputs/test_config/config_bundle_test.ini
```diff
@@ -74,6 +74,7 @@ skirt_distance = 6
 skirt_height = 1
 skirts = 1
 small_perimeter_speed = 15
+small_perimeter_length = 20
```
VanessaE: Maybe make this a bit larger by default?  20 mm would translate to a hole bit under 6.4 mm diameter.  I would suggest a default of 32, for a hole about 1 cm diameter.

xs/src/libslic3r/GCode.cpp
```diff
@@ -391,23 +391,23 @@ GCode::extrude(ExtrusionLoop loop, std::string description, double speed)
...
     // get paths
     ExtrusionPaths paths;
     loop.clip_end(clip_length, &paths);
     if (paths.empty()) return "";
-    
+
     // apply the small perimeter speed
     if (paths.front().is_perimeter()
         && !loop.has(erOverhangPerimeter)
-        && loop.length() <= SMALL_PERIMETER_LENGTH
```
VanessaE: I think you forgot to remove this constant's declaration?

https://github.com/slic3r/Slic3r/blob/fb2fefd4daa5600b7b46b428ba4278f3dc5a6e11/xs/src/libslic3r/libslic3r.h#L86

lordofhyphens:
@gennartan probably the "easiest" way to revert these changes is through the use of gitk and git-gui.

1. Start gitk and find the commit before you started to make your changes.
Right-click on the commit and select "reset branch to here"
image
1. Select "mixed" and click "OK"
image
1. With git-gui, select just the functional changes, and revert the rest.
image
1. Rewrite your commit message. 
1. Commit.
1. `git push --force` from the command line. 

This gets the diff of the changes down to around 15 lines of code (which is about what you'd expect from this change) and makes it much easier to tell what is going on.

In the future, using a different branch than master on your fork will cause fewer problems as well on your end (more flexibility), as more commits to master on your system will cause the PR to get updated.

[changes.txt](https://github.com/slic3r/Slic3r/files/2568964/changes.txt)

xs/src/libslic3r/PrintConfig.cpp
```diff
@@ -1281,6 +1281,15 @@ PrintConfigDef::PrintConfigDef()
...
     def->default_value = new ConfigOptionFloatOrPercent(15, false);
 
+    def = this->add("small_perimeter_length", coFloat);
+    def->label = __TRANS("Small perimeter length");
+    def->category = __TRANS("Layers and Perimeters");
+    def->tooltip = __TRANS("This setting controls the length of the small perimeter parameter.");
+    def->sidetext = "mm";
+    def->cli = "small-perimeter-length=f";
+    def->min = 0;
+    def->default_value = new ConfigOptionFloat(20);
```
lordofhyphens: set the default to `6.5 * 2 * PI`, so that we aren't breaking existing configurations.

src/test/inputs/test_config/config_bundle_test.ini
```diff
@@ -74,7 +74,7 @@ skirt_distance = 6
 skirt_height = 1
 skirts = 1
 small_perimeter_speed = 15
-small_perimeter_length = 20
+small_perimeter_length = 40
```
lordofhyphens: Why is this changed here? It is an input for a config test.

PR #4589, opened 2018-11-10
Simplify the Preferences UI for the default reload behavior and clarify labels

alranel:
As of now, the "Don't remind me again" checkbox upon selection of the reload behavior for models with added parts/modifiers is tied to a dedicated checkbox in the Preferences.
This can be simplified by adding "Prompt user" as one more possible value to the behavior setting itself. A "prompt" item is a common UI pattern.

I also tried to clarify the labels of the options because there were not clear at all to me and I had to look at the code to understand it better. :)

@Oekn5w @gege2b

### DIFF
--- a/lib/Slic3r/GUI/Plater.pm
+++ b/lib/Slic3r/GUI/Plater.pm
@@ -2493,10 +2493,10 @@ sub reload_from_disk {
         }
     }
 
-    my $reload_behavior = $Slic3r::GUI::Settings->{_}{reload_behavior};
+    my $reload_behavior = $Slic3r::GUI::Settings->{_}{reload_behavior} || 'prompt';
 
     # ask the user how to proceed, if option is selected in preferences
-    if ($org_obj_has_modifiers && !$Slic3r::GUI::Settings->{_}{reload_hide_dialog}) {
+    if ($org_obj_has_modifiers && $reload_behavior eq 'prompt') {
         my $dlg = Slic3r::GUI::ReloadDialog->new(undef,$reload_behavior);
         my $res = $dlg->ShowModal;
         if ($res==wxID_CANCEL) {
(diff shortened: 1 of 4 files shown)


lordofhyphens:
@storagejarett

lib/Slic3r/GUI/ReloadDialog.pm
```diff
@@ -12,16 +12,16 @@ use base 'Wx::Dialog';
...
 
     # selector
     $self->{choice} = my $choice = Wx::Choice->new($self, -1, wxDefaultPosition, wxDefaultSize, []);
-    $choice->Append("Reload all linked files");
-    $choice->Append("Reload main file, copy added parts & modifiers");
-    $choice->Append("Reload main file, discard added parts & modifiers");
+    $choice->Append("Reload parts/modifiers");
+    $choice->Append("Keep parts/modifiers (don't reload)");
+    $choice->Append("Discard parts/modifiers");
     $choice->SetSelection($default_selection);
```
lordofhyphens: With the addition of 'prompt', `$default_selection` may not be in the choice dropdown, which is causing a warning.

lib/Slic3r/GUI/Plater.pm
```diff
@@ -2510,12 +2510,15 @@ sub reload_from_disk {
...
         }
-        if ($dlg->GetHideOnNext) {
-            $Slic3r::GUI::Settings->{_}{reload_hide_dialog} = 1;
+        if (!$dlg->GetHideOnNext) {
+            $Slic3r::GUI::Settings->{_}{reload_behavior} = 'prompt';
             $save = 1;
         }
         Slic3r::GUI->save_settings if $save;
         $dlg->Destroy;
+    } else {
```
lordofhyphens: I added this to avoid one of the warnings (the rest of the code assumes that the value of `$reload_behavior` is numeric).

lordofhyphens:
@alranel there are still some unresolved issues with the changes around assumptions about the effective type of `$reload_behavior`

PR #4590, opened 2018-11-10
Rename "Extended GUI" to "Rotation controls" and simplify it

alranel:
I think the idea proposed in #4419 is great, as power users need advanced and flexible controls for manipulating models. Slic3r definitely needs to be very customizable in the long term so that every user can adapt it to their preferred workflow. As of now we don't have customizable palettes or other things like this, so the idea proposed by @foreachthing is a great start.

I think we can achieve the same goal by simplifying it a bit. I understand @foreachthing did it fully customizable in order to make sure his Pull Request was merged :-) and now I think we can improve over that.

Here I'm proposing the following changes:
* **always** have the *Rotate by 180°* commands in the contextual menu item (no need to make this configurable)
* No need to show the *Rotate Z 90°* buttons in the toolbar as we already have the *Rotate Z 45°* (two clicks are quicker than looking for the right button in a long row)
* Rename *Extended GUI* to *Rotation controls* as it's more descriptive

In this proposal I kept the big buttons proposed; however I'd like to understand better: @foreachthing, do you think that *all* toolbar buttons should be offered in the big size? Or just some? Only the rotation ones seem arbitrary, so we should probably think about a more consistent pattern.

schermata 2018-11-10 alle 23 38 39

schermata 2018-11-10 alle 23 41 52

### DIFF
--- a/lib/Slic3r/GUI.pm
+++ b/lib/Slic3r/GUI.pm
@@ -94,7 +94,7 @@ our $Settings = {
         tabbed_preset_editors => 1,
         show_host => 1,
         nudge_val => 1,
-        extended_gui => 0,
+        rotation_controls => 'z',
         reload_hide_dialog => 0,
         reload_behavior => 0
     },
--- a/lib/Slic3r/GUI/ColorScheme.pm
+++ b/lib/Slic3r/GUI/ColorScheme.pm
@@ -102,7 +102,7 @@ sub getSolarized { # add this name to Preferences.pm
     $SOLID_BACKGROUNDCOLOR = 1;               # Switch between SOLID or FADED background color
     my $largeicons         = 0;               # Default: 0. 1 for large icons-set.
     
-    if ($Slic3r::GUI::Settings->{_}{extended_gui} >= 4){
+    if ($Slic3r::GUI::Settings->{_}{rotation_controls} eq 'xyz-big'){
         $largeicons = 1;
     }
     
--- a/lib/Slic3r/GUI/Plater.pm
+++ b/lib/Slic3r/GUI/Plater.pm
@@ -199,14 +199,12 @@ sub new {
         $self->{htoolbar}->AddTool(TB_FEWER, "Fewer", Wx::Bitmap->new($Slic3r::var->("delete.png"), wxBITMAP_TYPE_PNG), '');
         $self->{htoolbar}->AddSeparator;
        
-        if ($Slic3r::GUI::Settings->{_}{extended_gui} >= 2) {  # if Toolbar enabled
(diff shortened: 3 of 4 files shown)


lordofhyphens:
@storagejarett

VanessaE:
Seems okay, but I wonder if it wouldn't be a better idea to use the three rotate-around-N-axis items, without the preceding 180° items, but pre-fill their dialogs with "180" instead of "0" (perhaps with auto-select-all when clicking on the number field, so that one need not manually delete the number to change it).

foreachthing:
Hmmm ... well, I (almost) never use _rotate XYZ **180°**_. It's mostly 90° (or 45°). So, I'd be back at square one. I just kept 180° in there, because it was in there. And some people don't like to change their habits.

> do you think that all toolbar buttons should be offered in the big size?

At work, I have a 4k monitor (170 DPI) and those buttons are tiny! If you can come up with a way of reading svgs (or different sized pngs, because of quality) as icons and scale them accordingly - I'd be very fond of that! :-) _(TeXstudio and Blender have such an option to scale the entire GUI at once)_

@VanessaE: Again: I don't like going through menus and change some default value manually (#1 reason for my work).
Too bad, it's such hard work to change the buttons ... thinking about SolidWorks, where you can drag and drop every function (button) at any place and then scale those as well. Just like TotalCommander - you can customize everything within the program.
I'd love to see button which you could assign macros. I.e. choose the function _rotate_, append a value, the add that button to the GUI - DONE.

@alranel: It does not matter to me how you call this option. As long as I can have my shortcuts.

OR: I like how you can rotate/scale a part in Cura! Sorry ...! But I am no programmer and don't know how one can do those GUI things.

This is just my opinion. And after all, _you_ should do what the majority wants.

alranel:
Thank you for your feedback @foreachthing. So you're proposing to make the size of all buttons configurable, and not just the rotate ones. This makes sense.

Visual rotation would be very nice indeed.

So according to your feedback I think we can remove the 180° rotation commands from the contextual menu, as long as we have the 90° buttons in the toolbar.

dwillmore:
The most common rotation I use is 45 degrees around Z so that diagonal infills print aligned to a rectilinear model.  For all other rotations I do, it's normally a multiple of 45.  So, for that, I have to agree with @foreachthing that a Cura like "pick the part and drag to rotate w/ optional snap-to-multiples of some angle".  For me that base angle would be 45 degrees.  Then I could pick a part, right click 'snap-rotate' (or whatever) and then drag it around to rotate it in 45 degree increments.  You could set that value to 1 degree to get a 'free rotate'.

A huge bonus would be to highlight all facets which are on the bed while rotating.  Maybe that could be a completely separate GUI function to toggle on and off and it works all the time.  Maybe I should open a separate issue for that.

lordofhyphens:
@dwillmore you should do that or promote an existing issue that asks for a very similar thing

PR #4592, opened 2018-11-11
C++ GUI Partial Merge - Presets

lordofhyphens:
Update PR from cppgui with current work for purposes of comment and review. 

Theory of Operation
-----
Slic3r::GUI::Preset is a store to manage loading and modifying a configuration referred to as a preset.

`load_config()` returns a shared_ptr to the dirty config; the clean config is not accessible to the out side world.

I've been thinking now and again to return a [weak_ptr](https://en.cppreference.com/w/cpp/memory/weak_ptr) to the dirty config to other systems instead (or having it still return a shared_ptr but it is a converted weak_ptr) .


src/GUI/Preset.hpp
```diff
@@ -21,23 +32,28 @@ class Preset;
 
 using Presets = std::vector<Preset>;
 
+class PresetEditor;
+
 class Preset {
 public:
+    friend class PresetEditor;
```
lordofhyphens: Rather than giving PresetEditor friend access, we can instead make a method config() that returns a config_ptr that is a transformed weak_ptr.

lordofhyphens: Better answer: new config_ref which is a std::weak_ptr<Slic3r::Config>. 

Usage in client functions that were asking for a dirty config is:
````
if (config_ptr x = preset.config()) { 
    ...
}
````

lordofhyphens:
I think this is ready to merge.

PR #4598, opened 2018-11-17
Save Purging Volume settings (MMU2)

JerryEnt:
It would be nice if we can save the Purging Volume setting (MMU2).
Now i have to set it every time I wanted to slice something.


lordofhyphens:
This is not Prusa3d.

lordofhyphens:
Nor is it actually a PR.

PR #4604, opened 2018-11-23
Expanding Slic3r::Log

lordofhyphens:
I've added basic filtering capabilities based on log level and topic name to Slic3r::Log. 

This review is primarily for usability questions; getting added next is basic file redirection and making output of ExtrusionEntities via SVG easy to command. I'm open for ideas for more things to log. 

This includes tests for the std::string system (std::wstring is converted based on locale).

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -198,6 +198,7 @@ add_library(libslic3r STATIC
     ${LIBDIR}/libslic3r/LayerRegionFill.cpp
     ${LIBDIR}/libslic3r/LayerHeightSpline.cpp
     ${LIBDIR}/libslic3r/Line.cpp
+    ${LIBDIR}/libslic3r/Log.cpp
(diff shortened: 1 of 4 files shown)


bubnikv:
Why don't you use the boost::log as we do for a long time in the Slic3r PE?

lordofhyphens:
I did not want to lock myself into littering Boost::log everywhere in the UI code. The default sink isn't that useful compared to std::clog and doing more with it puts a lot of complexity in the using code. I would much rather have a defined interface with testable, expected behavior for people to use.

Refactoring to make use of boost::log internally is certainly an option.

alranel:
Reverted because CI wasn't building. Sorry for merging too early.

codecvt is not available on gcc -4.9 which we are using on Travis/Linux. Looks like we need 5.1 for that.

lordofhyphens:
Ugh, I will probably do an #ifdef for now or find another conversion option because it is painful to change the gcc version used on the perl builds.

PR #4608, opened 2018-11-24
Slic3r::Log 

lordofhyphens:
As #4604 was merged by mistake, re-creating PR.

Changes:
Use boost::locale instead of corecvt.  

Note: "Revert Revert" is required because the previous reversion deleted files which confuses git when trying to merge later.

I suggest using squash + merge when merging PRs that are only reversions; it should provide a clean baseline.


lordofhyphens:
As @alranel had merged the earlier version of this and the blocking issue has been fixed, merging.

PR #4610, opened 2018-11-24
Use GCC7 instead of 4.9 with perl build.

lordofhyphens:
Branch to exercise building with gcc7 instead of gcc-4.9


lordofhyphens:
Yeah, this is a major pain to change slic3r-perl.

PR #4618, opened 2018-11-27
Remove support test compilation

lordofhyphens:
The c++ support tests were never fully completed and should be disabled until that code is looked at again and updated in a PR.


lordofhyphens:
Merged manually.

PR #4621, opened 2018-11-28
Refactoring: make Config a thin wrapper around DynamicPrintConfig

alranel:
I refactored Config.*pp by making it just a thin wrapper around DynamicPrintConfig, which was intended to be used directly. I expanded the ConfigBase interface with the missing accessors.

As of now, the only added value of Config are the static factory methods, which could be either removed or moved to DynamicPrintConfig.

@lordofhyphens what do you think about this?


alranel:
This refactoring makes the accessors less liberal in terms of type conversions. I think that stricter type checking is better; however these conversions can be restored by defining the needed accessors in the individual ConfigOption* classes.

lordofhyphens:
I can agree with it in theory; will look at it today in depth.

xs/src/libslic3r/ConfigBase.hpp
```diff
@@ -38,14 +63,22 @@ class ConfigOption {
...
-    virtual void setInt(int val) {};
-    virtual void setFloat(double val) {};
-    virtual void setString(std::string val) {};
-    virtual std::string getString() const { return ""; };
-    virtual std::vector<std::string> getStrings() const { return std::vector<std::string>(); };
+    
+    virtual bool getBool() const { throw BadOptionTypeException(); };
+    virtual void setBool(bool val) { throw BadOptionTypeException(); };
+    
+    virtual int getInt() const { throw BadOptionTypeException(); };
```
lordofhyphens: This is a good upgrade to the default behavior.

xs/src/libslic3r/ConfigBase.hpp
```diff
@@ -317,9 +351,19 @@ class ConfigOptionFloatOrPercent : public ConfigOptionPercent
     ConfigOptionFloatOrPercent() : ConfigOptionPercent(0), percent(false) {};
     ConfigOptionFloatOrPercent(double _value, bool _percent)
         : ConfigOptionPercent(_value), percent(_percent) {};
-    ConfigOptionFloatOrPercent* clone() const { return new ConfigOptionFloatOrPercent(this->value, this->percent); };
+    ConfigOptionFloatOrPercent* clone() const override { return new ConfigOptionFloatOrPercent(this->value, this->percent); };
     
-    void set(const ConfigOption &option) {
+    double getFloat() const override { throw BadOptionTypeException(); };
```
lordofhyphens: Wouldn't this be appropriate to only throw `BadOptionTypeException` if it's not a percent?

xs/src/libslic3r/ConfigBase.hpp
```diff
@@ -317,9 +351,19 @@ class ConfigOptionFloatOrPercent : public ConfigOptionPercent
     ConfigOptionFloatOrPercent() : ConfigOptionPercent(0), percent(false) {};
```
lordofhyphens: Since `ConfigOptionPercent` inherits off of `ConfigOptionFloat`, was it intended to have `getFloat()` return a bad option type or use the default behavior (return the floating point value)?
Between leaving it from `ConfigOptionFloat`'s behavior and then overriding it in `ConfigOptionFloatOrPercent`, I am unsure about the intended behavior here.

xs/src/libslic3r/ConfigBase.hpp
```diff
@@ -706,18 +750,29 @@ class ConfigBase
     virtual ~ConfigBase() {};
```
lordofhyphens: We should consider overriding Slic3r::ConfigBase&(const Slic3r::ConfigBase&) and operator=(const Slic3r::ConfigBase&) or deleting them as we have pointer data members, just to make sure the usage is obvious (and that the pointer is copied (as ConfigBase doesn't own its def). 
(-Weffc++)

xs/src/libslic3r/ConfigBase.hpp
```diff
@@ -727,14 +782,31 @@ class ConfigBase
...
+    void set_deserialize_throw(t_config_option_key opt_key, std::string str, bool append = false);
     double get_abs_value(const t_config_option_key &opt_key) const;
     double get_abs_value(const t_config_option_key &opt_key, double ratio_over) const;
-    bool getBool(const t_config_option_key &opt_key, bool default_value = false) const;
-    double getFloat(const t_config_option_key &opt_key, double default_value = 0.0) const;
-    int getInt(const t_config_option_key &opt_key, double default_value = 0) const;
-    std::string getString(const t_config_option_key &opt_key, std::string default_value = "") const;
-    std::vector<std::string> getStrings(const t_config_option_key &opt_key, std::vector<std::string> default_value = std::vector<std::string>()) const;
+    
+    bool getBool(const t_config_option_key &opt_key) const;
```
lordofhyphens: Good idea; this obviates the need for a lot of convoluted access.

xs/src/libslic3r/Print.cpp
```diff
@@ -230,7 +230,7 @@ Print::make_skirt()
         skirts = 1;
     
     const std::set<size_t> extruders{ this->extruders() };
-    auto extruder_it { extruders.begin() };
+    auto extruder_it = extruders.begin();
```
lordofhyphens: Can this be cbegin() instead of begin()?

xs/src/libslic3r/Config.hpp
```diff
@@ -5,30 +5,13 @@
 #include <initializer_list>
 #include <memory>
 #include <regex>
```
lordofhyphens: Is this still necessary?

src/test/libslic3r/test_config.cpp
```diff
@@ -48,25 +54,23 @@ SCENARIO("Generic config validation performs as expected.") {
 SCENARIO("Config accessor functions perform as expected.") {
     GIVEN("A config generated from default options") {
         auto config {Slic3r::Config::new_from_defaults()};
-        WHEN("A boolean option is set through the bool interface") {
-            config->set("gcode_comments", true);
-            THEN("The underlying value is set correctly.") {
-                REQUIRE(config->get<ConfigOptionBool>("gcode_comments").getBool() == true);
+        WHEN("A boolean option is set to a boolean value") {
```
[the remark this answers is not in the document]
alranel: This is a mistake :)

src/test/libslic3r/test_config.cpp
```diff
@@ -48,25 +54,23 @@ SCENARIO("Generic config validation performs as expected.") {
...
             }
         }
-        WHEN("A boolean option is set through the string interface") {
+        WHEN("A boolean option is set to a string value") {
             config->set("gcode_comments", "1");
             THEN("The underlying value is set correctly.") {
                 REQUIRE(config->get<ConfigOptionBool>("gcode_comments").getBool() == true);
             }
         }
-        WHEN("A boolean option is set through the int interface") {
```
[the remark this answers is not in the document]
alranel: I'm a bit afraid of all type conversions, so I'd avoid them unless needed. Where do we need to set a boolean option with a string value?

src/test/libslic3r/test_config.cpp
```diff
@@ -156,36 +159,34 @@ SCENARIO("Config accessor functions perform as expected.") {
...
+                REQUIRE_THROWS_AS(config->get_ptr<ConfigOptionFloat>("deadbeef_invalid_option", false), UnknownOptionException);
+                REQUIRE_THROWS_AS(config->get_ptr<ConfigOptionInt>("deadbeef_invalid_option", false), UnknownOptionException);
+                REQUIRE_THROWS_AS(config->get_ptr<ConfigOptionBool>("deadbeef_invalid_option", false), UnknownOptionException);
             }
         }
 
         WHEN("getX called on an unset option.") {
             THEN("The default is returned.") {
                 REQUIRE(config->getFloat("layer_height") == 0.3);
-                REQUIRE(config->getString("layer_height") == "0.3");
```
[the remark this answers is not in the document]
alranel: Not entirely sure. Where do we need this as a string? For the GUI fields maybe? In that case I'd consider doing the conversion/rendering in the Slic3r::GUI::OptionsGroup::Field::TextCtrl class.

xs/src/libslic3r/Config.hpp
alranel: No, right

xs/src/libslic3r/ConfigBase.hpp
```diff
@@ -26,6 +26,31 @@ extern std::string escape_strings_cstyle(const std::vector<std::string> &strs);
 extern bool unescape_string_cstyle(const std::string &str, std::string &out);
 extern bool unescape_strings_cstyle(const std::string &str, std::vector<std::string> &out);
 
+/// Specialization of std::exception to indicate that an unknown config option has been encountered.
```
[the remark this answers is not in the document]
alranel: If they're only used by a single .hpp/.cpp pair, I'd leave them there.

xs/src/libslic3r/ConfigBase.hpp
```diff
@@ -26,6 +26,31 @@ extern std::string escape_strings_cstyle(const std::vector<std::string> &strs);
...
+class ConfigOptionException : public std::exception {
```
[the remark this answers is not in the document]
alranel: True, but `logic_error` has a different interface as it accepts a generic string message which we don't need for the base `ConfigOptionException`. Other than this, no strong opinion.

xs/src/libslic3r/ConfigBase.hpp
```diff
@@ -26,6 +26,31 @@ extern std::string escape_strings_cstyle(const std::vector<std::string> &strs);
...
+    public:
+    virtual const char* what() const noexcept {
+        std::string s("Invalid value for option: ");
+        s += this->opt_key;
+        return s.c_str();
+    }
+};
+
+/// Specialization of std::exception to indicate that an unsupported accessor was called on a config option.
+class BadOptionTypeException : public std::exception {};
```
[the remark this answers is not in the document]
alranel: Okay

xs/src/libslic3r/ConfigBase.hpp
alranel: It's wanted, as `getFloat()` cannot return the `percent` flag, so the semantics of the returned float are totally ambiguous...

xs/src/libslic3r/ConfigBase.hpp
alranel: But you don't know whether it's a percent or not before calling this :)
(You usually call `ConfigBase::getFloat()` which in turn calls this. If you dereference the option via `config->option(opt_key)` you can then call `->value` and `->percent` directly)

xs/src/libslic3r/ConfigBase.hpp
alranel: True.

xs/src/libslic3r/Print.cpp
alranel: Sure. AFAIK, `begin()` called on a `const` variable behaves as `cbegin()` anyway.

xs/src/libslic3r/ConfigBase.hpp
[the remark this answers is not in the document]
lordofhyphens: You're right about the default. We can leave it as it is.

xs/src/libslic3r/ConfigBase.hpp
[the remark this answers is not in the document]
lordofhyphens: I think we can leave this here (after thinking about it more), as only Config should throw these exceptions.

src/test/libslic3r/test_config.cpp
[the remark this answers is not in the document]
lordofhyphens: I was thinking about GUI first and foremost. You are probably right about conversion, I will have to try using it more to see how bad it is.

lordofhyphens:
I cherry picked the compiler fix, FYI

xs/src/libslic3r/Print.cpp
lordofhyphens: Works for me.

xs/src/libslic3r/ConfigBase.hpp
lordofhyphens: So ConfigOptionPercent supports getFloat, then?

xs/src/libslic3r/ConfigBase.hpp
lordofhyphens: The option knows the state of its own percent flag, though.

src/test/libslic3r/test_config.cpp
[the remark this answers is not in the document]
lordofhyphens: I am okay with type conversions where the input is strongly defined by convention.

Booleans happen to be defined in a few common ways.

YAGNI is as good as reason as any to pull out.

xs/src/libslic3r/ConfigBase.hpp
lordofhyphens: Deleting is no good; I've defined the relevant methods and set them to = default .

PR #4632, opened 2018-12-03
Printer Estimate Duration on filename

bionno:
Can you add , like an option on the filename, the printer estimate duration on the filename ?

F.I. SimpleSquare-0.2mm-Mk2.5-1h-30m.gcode (so 1h-30m) is the printing estimated time .

Thanks,
Andrea


supermerill:
This is slic3r, not slic3rPE, you may want to open your request in their github: https://github.com/prusa3d/Slic3r/issues

PR #4634, opened 2018-12-03
remove warnings (int -> size_t & cast)

supermerill:
remove some wrning from int<-> size_t comaprisons
remove unsafe "auto" from SupportMaterial.cpp
remove unneeded copies (inside for-loop) from SupportMaterial.cpp


ilovelinux:
Maybe because size_t can be at most 2^64-1 and int can be at most 2^31-1. I think that the segfault could be because of an overflow

PR #4669, opened 2019-01-02
Fix skirt/Brim

lordofhyphens:
I've finished porting the tests from `t/skirt_brim.t` and found/fixed some bugs along the way.


alranel:
It's not clear to me what issue required to expose `skirt_height_z`

lordofhyphens:
Tests were failing and I tracked it down to that logic. skirt_height_z was never actually used for anything outside of that method. Moreover, skirt_height is in layers and the index for the map was in scaled mm so the logic around

PR #4671, opened 2019-01-03
Adaptive slicing tests

lordofhyphens:
Ported tests from t/adaptive_slicing.t to CATCH. 

Compare my written tests to the original Perl.

Comments/approval from @platsch would be very helpful.


platsch:
Also: I think with this commit we can remove
```
t/adaptive_slicing.t
xs/xsp/SlicingAdaptive.xsp
```
to get rid of the remaining XS/Perl code

platsch:
@lordofhyphens One more thing: it's not directly related to this PR, but I've ported the `Point` test to catch some time ago when we started playing with cmake and catch:
https://github.com/platsch/Slic3r/blob/catch-testing/xs/t/point_t.cpp

It includes a custom matcher and `std::ostream` operator to get useful evaluations if a test fails.
Would you appreciate an updated PR?

lordofhyphens:
@platsch yes I would appreciate a PR for the point test.

PR #4672, opened 2019-01-03
Fix division by zero when min=max layer heights

### LINKED ISSUE #4670: v1.3.1-dev-45353b7: Division by zero if max_layer_height == min_layer_height
### Version
Slic3r-1.3.1-dev-45353b7-x86_64.AppImage

Linux be10 3.16.0-7-amd64 #1 SMP Debian 3.16.59-1 (2018-10-03) x86_64 GNU/Linux

### Behavior
Set extruder limits to 0.1 mm / 0.1 mm
Load any STL
Right click → settings

Resulting crash:
**Illegal division by zero at /home/7eggert/.lib/slic3r-1.3.1-dev-d17659b-linux-x64/local-lib/lib/perl5/Slic3r/GUI/Plater/SplineControl.pm line 270.**

I don't believe the config to be necessary, but here it is:
[config.ini.gz](https://github.com/slic3r/Slic3r/files/2721345/config.ini.gz)

PS and OT: On Printer Settings → Extruder $n → Limits, the label should say "Layer Height Limits" instead of "Limits"

lordofhyphens:
Fixes #4670.

### DIFF
--- a/lib/Slic3r/GUI/Plater/SplineControl.pm
+++ b/lib/Slic3r/GUI/Plater/SplineControl.pm
@@ -267,10 +267,10 @@ sub _update_canvas_size {
     my @size = ($canvas_w - 2*$padding, $canvas_h - 2*$padding);
     $self->{canvas_size} = [@size];
     
-    if ($self->{max_layer_height} == $self->{min_layer_height}) {
-        $self->{scaling_factor_x} = 1.0;
-    } else {
+    if ($self->{max_layer_height} > $self->{min_layer_height}) {
         $self->{scaling_factor_x} = $size[0]/($self->{max_layer_height} - $self->{min_layer_height});
+    } else {
+        $self->{scaling_factor_x} = 1.0;
     }
     $self->{scaling_factor_y} = $size[1]/$self->{object_height};
 }
--- a/lib/Slic3r/GUI/Plater/SplineControl.pm
+++ b/lib/Slic3r/GUI/Plater/SplineControl.pm
@@ -267,7 +267,11 @@ sub _update_canvas_size {
     my @size = ($canvas_w - 2*$padding, $canvas_h - 2*$padding);
     $self->{canvas_size} = [@size];
     
-    $self->{scaling_factor_x} = $size[0]/($self->{max_layer_height} - $self->{min_layer_height});
+    if ($self->{max_layer_height} == $self->{min_layer_height}) {
+        $self->{scaling_factor_x} = 1.0;
+    } else {
+        $self->{scaling_factor_x} = $size[0]/($self->{max_layer_height} - $self->{min_layer_height});
+    }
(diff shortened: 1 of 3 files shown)


lib/Slic3r/GUI/Plater/SplineControl.pm
```diff
@@ -267,7 +267,11 @@ sub _update_canvas_size {
     my @size = ($canvas_w - 2*$padding, $canvas_h - 2*$padding);
     $self->{canvas_size} = [@size];
     
-    $self->{scaling_factor_x} = $size[0]/($self->{max_layer_height} - $self->{min_layer_height});
+    if ($self->{max_layer_height} == $self->{min_layer_height}) {
```
hyperair: These are floats, aren't they? I think you shouldn't use an equality comparison here.

```suggestion
    if (abs($self->{max_layer_height} - $self->min_layer_height) < 0.000001) {
```

platsch: Yes. But a max_layer_height small than min_layer_height doesn't make any sense anyway. So I suggest:
```suggestion
    if ($self->{max_layer_height} > $self->{min_layer_height}) {
        $self->{scaling_factor_x} = $size[0]/($self->{max_layer_height} - $self->{min_layer_height});
    } else {
        $self->{scaling_factor_x} = 1.0;
    }
```
Additionally, we should catch this case when sanitizing the config and request the user to enter useful values.

supermerill: You can use EPSILON instead of "0.000001"

lordofhyphens: Done.

lordofhyphens: Did it with a comparator as @platsch suggested (which makes more sense here than my original test).

lib/Slic3r/GUI/Plater/SplineControl.pm
```diff
@@ -267,7 +267,11 @@ sub _update_canvas_size {
     my @size = ($canvas_w - 2*$padding, $canvas_h - 2*$padding);
     $self->{canvas_size} = [@size];
     
-    $self->{scaling_factor_x} = $size[0]/($self->{max_layer_height} - $self->{min_layer_height});
+    if ($self->{max_layer_height} > $self->{min_layer_height}) {
```
hyperair: I think you should still check for a minimum difference between max and min layer height so that we don't end up with silly numbers for scaling_factor_x when they're almost the same but still satisfy this condition.

lordofhyphens: I think if the user is doing particularly silly things that are technically correct (here, it is technically correct from the operation of the software to have min layer 0.1 and max layer 0.100000000001),  they deserve to get technically correct yet silly outputs.

We check for max <= min via the config validation as well.

PR #4679, opened 2019-01-06
Patch for #4620 - ensure posDetectSurfaces is invalidated on slice change

### LINKED ISSUE #4620: Top and bottom solid layers vanish when changed
**Describe the bug**
Whenever I change the number of top or bottom solid layers, either through retyping or up/down arrows, all existing top and bottom solid layers vanish.  Independent of tab/dialog setting.  Independent of background processing.

**To Reproduce**
Load any model.  A test cube will do.  Use a config that has top and bottom layers. 
Change the number of top layers.
See the top/bottom layers vanish

[missing-tb-layers.zip](https://github.com/slic3r/Slic3r/files/2622904/missing-tb-layers.zip)

**Expected behavior**
Number of top and bottom layers should match config

**Screenshots**
missing-tb-layers

**Desktop (please complete the following information):**
 - OS: Fedora 26 x86-64
 - Version: git 33b3faacfd3b50cc643d6eecd6f48aec8ddcc7bc

lordofhyphens:
Invalidate posDetectSurfaces during make_perimeters() when posSlice would also get invalidated.
Can't use invalidate_step() as that invalidates the perimeters as well.

Fixes #4620

### DIFF
--- a/src/test/libslic3r/test_print.cpp
+++ b/src/test/libslic3r/test_print.cpp
@@ -91,7 +91,7 @@ SCENARIO("Print: Changing number of solid surfaces does not cause all surfaces t
             print->regions[0]->config.top_solid_layers = 3;
             print->objects[0]->invalidate_step(posPrepareInfill);
             print->process();
-            THEN("Print object does not have 0 bottom surfaces.") {
+            THEN("Print object does not have 0 solid bottom layers.") {
                 test_is_solid_infill(print, 0, 0);
             }
             AND_THEN("Print object has 3 top solid layers") {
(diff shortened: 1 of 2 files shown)


lordofhyphens:
Also have a regression test added to the CATCH test suite.

VanessaE:
I can't comment on the code itself, but the behavior is spot-on.  lgtm.

alranel:
I'd turn `this->state.invalidate(posSlice);` into `this->invalidate_step(posSlice);` like we were doing before the bisected commit. In order to do this safely, I'd move set_started(posPerimeters) after the "if (this->typed_slices) {" block.
I think that state.invalidate() should be never called directly, except from invalidate_step() which ensures dependencies are invalidated too

lordofhyphens:
Done.

PR #4686, opened 2019-01-09
README.md:grammar improvement suggestions [skip ci]

andy5995:
I won't explain why I'm making these suggestions unless I'm asked. I think they are self-explanatory ;)


lordofhyphens:
Thanks!

PR #4691, opened 2019-01-10
brim ears

supermerill:
https://github.com/slic3r/Slic3r/issues/4687 

- add brim_ears setting: bool: set to true to activate this new funtino
 - add brim_ears_max_angle : float : max angle for a corner to be assigned a brim ear.

### DIFF
--- a/xs/src/libslic3r/Print.cpp
+++ b/xs/src/libslic3r/Print.cpp
@@ -349,6 +349,8 @@ Print::invalidate_state_by_config(const PrintConfigBase &config)
             osteps.insert(posSupportMaterial);
         } else if (opt_key == "brim_width"
             || opt_key == "interior_brim_width"
+            || opt_key == "brim_ears"
+            || opt_key == "brim_ears_max_angle"
             || opt_key == "brim_connections_width") {
             steps.insert(psBrim);
             steps.insert(psSkirt);
@@ -1084,6 +1086,7 @@ Print::_make_brim()
     
     const coord_t grow_distance = flow.scaled_width()/2;
     Polygons islands;
+    Points pt_ears;
     
     for (PrintObject* object : this->objects) {
         const Layer* layer0 = object->get_layer(0);
@@ -1103,6 +1106,10 @@ Print::_make_brim()
             for (Polygon p : object_islands) {
                 p.translate(copy);
                 islands.push_back(p);
+                if(this->config.brim_ears)
+                    for (const Point &p_corner : p.convex_points(this->config.brim_ears_max_angle.value * PI / 180.0)) {
+                        pt_ears.push_back(p_corner);
+                    }
             }
         }
     }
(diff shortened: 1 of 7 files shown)


lordofhyphens:
Any thoughts on how to test this?

VanessaE:
Seems to me that the best way is to actually print something that's normally guaranteed to curl off of the bed, in ABS on untreated glass, playing with the brim settings to see the effect

lordofhyphens:
@VanessaE I mean for the code generation.

Maybe inspecting the data structures?

VanessaE:
You got me.......

supermerill:
I forgot: you should check the tooltips & text. I'm not a good english speaker.

xs/src/libslic3r/Print.cpp
```diff
@@ -1123,6 +1130,124 @@ Print::_make_brim()
         ));
     }
     
+    if(this->config.brim_ears){
+        
+        //create ear pattern
+        coord_t size_ear = (scale_(this->config.brim_width.value) - flow.scaled_spacing());
+        Polygon point_round;
+        point_round.points.push_back(Point(size_ear*1, 0*size_ear));
```
ledvinap: This can be replaced with array of x,y coefficients ...
Also maybe angle + sin/cos may be used to generate the ear?

xs/src/libslic3r/Print.cpp
```diff
@@ -1123,6 +1130,124 @@ Print::_make_brim()
...
+                        best_idx = i;
+                    }
+                }
+                if (best != NULL) {
+                    //copy new line inside the sorted array.
+                    lines_sorted.push_back(lines[best_idx]);
+                    lines.erase(lines.begin() + best_idx);
+                    
+                    //connect if near enough
+                    if (lines_sorted.size() > 1) {
```
ledvinap: this is always true

xs/src/libslic3r/Print.cpp
```diff
@@ -1123,6 +1130,124 @@ Print::_make_brim()
...
+                        size_t idx = lines_sorted.size() - 2;
+                        bool connect = false;
```
ledvinap: it is possible to check `best_dist` first and skip testing distance again.

xs/src/libslic3r/Print.cpp
```diff
@@ -1123,6 +1130,124 @@ Print::_make_brim()
...
+        Polylines lines = intersection_pl(union_pt_chained(loops), mouse_ears);
+        
+        //reorder them
+        Polylines lines_sorted;
+        Polyline* previous = NULL;
+        Polyline* best = NULL;
+        double best_dist = -1;
+        size_t best_idx = 0;
+        while (lines.size() > 0) {
+            if (previous == NULL) {
```
ledvinap: (this can be moved before `while`, reducing nesting)

xs/src/libslic3r/Print.cpp
supermerill: yes. I used that because it's quick and easy to write & test. Also we don't need a big precision and this number of point is enough. Too much and the intersection takes ages.

What code do you propose?

xs/src/libslic3r/Print.cpp
supermerill: ?

xs/src/libslic3r/Print.cpp
supermerill: Also, with some more tough, the if (previous == NULL)  block should be put before the while.

xs/src/libslic3r/Print.cpp
ledvinap: something like (untested):
```
for (int i = 0; i < POLY_SIDES; i++) {
  float angle = 2 * PI / POLY_SIDES * i;
   point_round.points.push_back(Point(size_ear * cos(angle), size_ear * sin(angle));
}

xs/src/libslic3r/Print.cpp
ledvinap: Closest distance between lines is already available in `best_dist`

supermerill: but i need to know where to do reverse(). So i need to store more information than just best_dist...

ledvinap: Yes, the tests will be necessary. But testing best_dist can skip the tests altogether.
It's just minor optimization ...

supermerill: oh, i understand

PR #4701, opened 2019-01-14
Add raft_offset to the options dialog.

mat813:
I have a large object with big holes and the raft does not work as well as it could with big holes, and I discovered there was already an option that I could use to fix that, but it was not available in the GUI, for some reason.


mat813:
Ah, I spoke too soon, it does not seem to be doing anything :(

mat813:
My problem is the raft here has holes and it makes it harder to stick.

capture d ecran 2019-01-14 a 13 07 12

PR #4717, opened 2019-01-25
extra_perimeters : now add perimeters for overhangs.

supermerill:
extra_perimeters: now also add extra perimeters if infill-over-overha…ngs are detected.
extra_perimeters: Use CW instead of CCW every odd layers for overhangs.

note: currently building & debugging, please wait a bit before review.


lordofhyphens:
Extra perimeters triggers on some slopes, where otherwise you would see infill

supermerill:
could you tell me the file and the lines? I tried to search each call to the setting but maybe i forgot some files in my search.

PR #4728, opened 2019-02-06
Simplify collinear points

### LINKED ISSUE #4726: "simplication" of contour polygons: remove collinear points from tesselation artifacts
Consider a simple object like this:

object

The cylinder is composed of a number of rectangular surfaces, which is a very common pattern for many objects. The tesselated representation of this model looks like this:

object_tesselated_label

All rectangular surfaces are now split into triangles. When slicing this model, the result looks like this (Layer 5):

layer_5

Every additional triangle line (red marker in the tesselated image) in each rectangle causes slic3r to add a point which is perfectly collinear with its two neighbors, but entirely useless.
Most (but not all!) of the redundant points are removed during the gcode-export by the Douglas-Peucker algorithm (invoked via polygon.simplify [here](https://github.com/slic3r/Slic3r/blob/master/xs/src/libslic3r/GCode.cpp#L323), but all polygon operations during the entire slicing process have to respect the useless points.

Is there any specific reason to keep collinear polygon points? If not, I propose to remove them right after or during the `slice()` step in TriangleMeshSlicer.
I did an implementation of this, the result for the same object is as expected:

layer_5_no_collinear

Some quick tests with different objects indicated a speedup of the full slicing process between 10% and 30%, but I did not extensively test this.  Worst case would be a model purely composed of triangles which do not assemble a larger, planar surface. In this case the `remove_collinear_points` would add some overhead.

I also noted a `remove_vertical_collinear_points` method in Polygon.cpp which is apparently used for rectilinear infill patterns, but I'm not sure whether a general implementation of remove_collinear_points could replace this.

If there is no obvious reason to not apply this simplification, I can quickly file a PR with my implementation.

platsch:
Fixes #4726.

collinear lines are removed from the `slices` polygons of each layer (parallelized).

This could potentially also be implemented in the [make_loops](https://github.com/slic3r/Slic3r/blob/master/xs/src/libslic3r/TriangleMesh.cpp#L1135) method where the polygons from slicing are assembled, not sure if that would gain any performance.

Includes a simple CATCH testcase, but only for the remove_collinear_points part of Polygon.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -306,6 +306,7 @@ set(SLIC3R_TEST_SOURCES
     ${TESTDIR}/libslic3r/test_geometry.cpp
     ${TESTDIR}/libslic3r/test_log.cpp
     ${TESTDIR}/libslic3r/test_model.cpp
+    ${TESTDIR}/libslic3r/test_polygon.cpp
     ${TESTDIR}/libslic3r/test_print.cpp
     ${TESTDIR}/libslic3r/test_printgcode.cpp
     ${TESTDIR}/libslic3r/test_skirt_brim.cpp
--- a/src/test/libslic3r/test_polygon.cpp
+++ b/src/test/libslic3r/test_polygon.cpp
@@ -1,0 +1,48 @@
+
+#include <catch.hpp>
+
+#include "Point.hpp"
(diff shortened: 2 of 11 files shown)


lordofhyphens:
Thanks @platsch

Issue #3139, opened 2015-11-25
Another bridging issue


BurntNoodles:
Hello,

(using Slicer 1.2.9, windows 7 x64)
I have what should be a pretty straightforward part to bridge, but it looks like the bridge detection is having some issues. Here is the layer before the bridge:

![prebridgelayer](https://cloud.githubusercontent.com/assets/16022535/11408047/d4ee1f16-9373-11e5-8e22-6883f6518f9c.PNG)

Here is the first bridge layer:

![bridgelayer](https://cloud.githubusercontent.com/assets/16022535/11408055/e4efc2c0-9373-11e5-89e9-33da5a5da3c5.PNG)

The specific issues are that on one bridge (the bottom in the image), it chooses the longest direction to bridge; the bridge on the other side (which should be symmetric?) is at some peculiar angle.
From hunting around on here, it looks like this has cropped up before, and sometimes makes a re-appearance in the main slic3r branch- maybe this happened again?

A maybe related issue, the next layer after the bridge doesn't merge the solid infill into one region:

![bridge2ndlayer](https://cloud.githubusercontent.com/assets/16022535/11408128/71086fd2-9374-11e5-9a71-b3f39d3112f1.PNG)

Thanks for any help!

Here is the .stl file and config:
https://www.dropbox.com/s/2g3vaxrwnons29k/BridgingIssues%2011-25-2015.zip?dl=0

lordofhyphens:
Unfortunately Dropbox has killed your test case. Can you please reupload them here?

lordofhyphens:
Alternatively, could you test with a recent build from http://dl.slic3r.org/dev?

VanessaE:
This may have worked properly at some point since this issue was filed, but I can confirm a substantially similar issue with commit 748dd3c23a31e34a155c325f281d21b7cd516c72 (2019-01-05) also.

Example:

![screenshot_2019-01-12_16-46-04](https://user-images.githubusercontent.com/1733537/51078824-9ead5100-1689-11e9-9f3e-4618aa059c09.png)

I would have expected the bridging infill to be drawn parallel to the perimeters, since this is a rectangular gap.

Model:  [y_tensioner_with_internal_nut_retainer_-__M3_bearing_screw.stl.zip](https://github.com/slic3r/Slic3r/files/2752536/y_tensioner_with_internal_nut_retainer_-__M3_bearing_screw.stl.zip)
[...]

PR #4743, opened 2019-02-20
Fixed that bridge angle is lost on shell layers

nstbayless:
Previously, if a bridge was on a layer which had solid infill due to a horizontal shell, the bridge angle would be reset to -1. This could occur e.g. if a model has two bridges on on neighbouring layers (even if the bridges aren't near each other).

This may be a fix for issues https://github.com/slic3r/Slic3r/issues/3175 and https://github.com/slic3r/Slic3r/issues/3139.

### DIFF
--- a/xs/src/libslic3r/PrintObject.cpp
+++ b/xs/src/libslic3r/PrintObject.cpp
@@ -1664,7 +1664,7 @@ PrintObject::_discover_neighbor_horizontal_shells(LayerRegion* layerm, const siz
             append_to(tmp, to_polygons(internal));
             
             const auto solid_surfaces = diff_ex(to_polygons(s), tmp, true);
-            neighbor_layerm->fill_surfaces.append(solid_surfaces, s.front()->surface_type);
+            neighbor_layerm->fill_surfaces.append(solid_surfaces, *s.front());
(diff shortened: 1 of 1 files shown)


lordofhyphens:
@nstbayless I'd appreciate a regression test case and an example as situations where the existing behavior is doing something wrong before merging.

VanessaE:
This does fix #3139, for the most part (Slic3r still chooses some odd/non-ideal bridging angles, but I think that's a different issue).

lordofhyphens:
@nstbayless do you have any interest in completing this PR?

nstbayless:
@lordofhyphens Sure thing. I'll have free cycles to implement the required changes in about 2 weeks.

VanessaE:
*bump*  @nstbayless  I know time is short but let's get this done and merged.

VanessaE:
I just tested this against 4f5b935ecf1335755d607ec38395bd94106daa7f.  It still seems to work as it did before:  it's not perfect, but it compiles fine, all unittests pass, and it fixes bridging angle well enough to to be worth merging, in my opinion.  Whatever details are left to fix should just be sorted out later.

PR #4809, opened 2019-05-17
Allow for flattening with and without sort.

lordofhyphens:
Responding to comments in 6710b24 by @bubnikv and incorporating suggested changes made by @supermerill from #4553 as a single function with a flag instead of two methods.

### DIFF
--- a/xs/src/libslic3r/ExtrusionEntityCollection.cpp
+++ b/xs/src/libslic3r/ExtrusionEntityCollection.cpp
@@ -220,25 +220,41 @@ ExtrusionEntityCollection::items_count() const
     return count;
 }
 
-/* Returns a single vector of pointers to all non-collection items contained in this one */
+/* Returns a single vector of chained (new) pointers to all non-collection items contained in this one */
 void
-ExtrusionEntityCollection::flatten(ExtrusionEntityCollection* retval) const
+ExtrusionEntityCollection::flatten(ExtrusionEntityCollection* retval, bool preserve_ordering) const
 {
-    for (ExtrusionEntitiesPtr::const_iterator it = this->entities.begin(); it != this->entities.end(); ++it) {
-        if ((*it)->is_collection()) {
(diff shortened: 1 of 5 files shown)


xs/src/libslic3r/ExtrusionEntityCollection.cpp
```diff
@@ -220,25 +220,41 @@ ExtrusionEntityCollection::items_count() const
...
-            ExtrusionEntityCollection* collection = dynamic_cast<ExtrusionEntityCollection*>(*it);
-            retval->append(collection->flatten().entities);
-        } else {
-            retval->append(**it);
+    if (this->no_sort and preserve_ordering) {
+        /// if we want to preserve ordering and we can't sort, break out the unsorted ones first.
+        ExtrusionEntityCollection *unsortable = new ExtrusionEntityCollection(*this);
+        retval->append(*unsortable);
+        unsortable->entities.clear();
+        for (ExtrusionEntitiesPtr::const_iterator it = this->entities.begin(); it != this->entities.end(); ++it) {
```
lordofhyphens: This probably could be reduced more than it is currently.

supermerill: We can replace 
```
        ExtrusionEntityCollection *unsortable = new ExtrusionEntityCollection(*this);
        retval->append(*unsortable);
        unsortable->entities.clear();
```
by

```
        ExtrusionEntityCollection *unsortable = new ExtrusionEntityCollection();
        unsortable->no_sort = true;
        unsortable->orig_indices = this->orig_indices;
        retval->append(*unsortable);
```

Maybe we should also say that flatten() will add at least somthing inside retval and it can't add an empty ExtrusionEntityCollection if there are no empty ExtrusionEntityCollection inside the tree.

Other note: I added a visitor to The extrusion entity object hierarchy, to remove all dynamic_cast. Are you interested by a pr about that?

lordofhyphens:
I am usually interested in things that reduce the amount of dynamic cast calls.

Issue #3317, opened 2016-05-03
Slic3r crash with STL file


sebastien:
### Version

1.2.9

### Operating system type + version

Ubuntu Linux 16.04LTS 64b

### Behavior

Loading the attached STL file (`crash.stl`) makes slic3r crash with the following backtrace:

```
*** Error in `slic3r': realloc(): invalid old size: 0x0000000004cbe400 ***
======= Backtrace: =========
/lib/x86_64-linux-gnu/libc.so.6(+0x77725)[0x7f480753c725]
/lib/x86_64-linux-gnu/libc.so.6(+0x82bfa)[0x7f4807547bfa]
/lib/x86_64-linux-gnu/libc.so.6(realloc+0x179)[0x7f4807548db9]
/home/sebastien/Local/share/Slic3r/lib/vrt/bbbcb3e4bbe8939428f72f745f596360/XS.so(stl_add_facet+0x16b)[0x7f4805b2046b]
/home/sebastien/Local/share/Slic3r/lib/vrt/bbbcb3e4bbe8939428f72f745f596360/XS.so(stl_fill_holes+0x3f9)[0x7f4805b208a9]
/home/sebastien/Local/share/Slic3r/lib/vrt/bbbcb3e4bbe8939428f72f745f596360/XS.so(_ZN6Slic3r12TriangleMesh6repairEv+0xe9)[0x7f4805ab6e49]
/home/sebastien/Local/share/Slic3r/lib/vrt/bbbcb3e4bbe8939428f72f745f596360/XS.so(+0xbbbab)[0x7f4805955bab]
/home/sebastien/Local/share/Slic3r/bin/../dll/libperl.so(Perl_pp_entersub+0x5e3)[0x7f48086b9db3]
/home/sebastien/Local/share/Slic3r/bin/../dll/libperl.so(Perl_runops_standard+0xe)[0x7f48086b83fe]
/home/sebastien/Local/share/Slic3r/bin/../dll/libperl.so(Perl_call_sv+0x48b)[0x7f480865228b]
/home/sebastien/Local/share/Slic3r/lib/vrt/8b57d08e0d55cfe6498642b59ecf5ad1/Wx.so(_ZN18wxPliEventCallback7HandlerER7wxEvent+0x31c)[0x7f480413f58c]
[...]

PR #4811, opened 2019-05-18
Use admesh's stl_repair() instead of simulating it ourselves.

### LINKED ISSUE #4805: Slic3r crashs loading STL
**Describe the bug**
I need to print the gears of this package: https://www.thingiverse.com/thing:974906
When I load the STL file, Slic3r crashes. This is the output:
```
$ slic3r
realloc(): invalid next size
Aborted (core dumped)
```
I also tried to fix the .stl file with `slic3r --repair` but it doesn't work too.

**To Reproduce**
1. Go to 'https://www.thingiverse.com/thing:974906'
2. Download and unzip files
3. Open _gears.stl_

**3D Model and Slic3r Configuration Export**
[issue.zip](https://github.com/slic3r/Slic3r/files/3167614/issue.zip)

**Expected behavior**
Open STL file

**Desktop (please complete the following information):**
 - OS: Archlinux
 - Version 1.3.0 (installed by [AUR](https://aur.archlinux.org/packages/slic3r/))

lordofhyphens:
fixes #4805

### DIFF
--- a/xs/src/libslic3r/TriangleMesh.cpp
+++ b/xs/src/libslic3r/TriangleMesh.cpp
@@ -174,7 +174,21 @@ TriangleMesh::repair() {
     this->check_topology();
 
     /// Call the stl_repair from admesh rather than reimplementing it ourselves.
-    stl_repair(&(this->stl), 1, 1, 0, 0, 0, 0, 0, 100, 1, 1, 1, 1, 1, 0);
+    stl_repair(&(this->stl),  // operate on this STL
+               true,  // flag: try to fix everything
+               true,  // flag: check for perfectly aligned edges
+               false, // flag: don't use tolerance
+               0.0,    // null tolerance value
+               false, // flag: don't increment tolerance
+               0.0,   // amount to increment tolerance on each iteration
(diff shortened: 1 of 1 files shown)


lordofhyphens:
Also probably avoids/patches #3317

xs/src/libslic3r/TriangleMesh.cpp
```diff
@@ -172,23 +172,9 @@ TriangleMesh::repair() {
...
-    }
-    
-    // normal_directions
-    stl_fix_normal_directions(&stl);
-    
-    // normal_values
-    stl_fix_normal_values(&stl);
+
+    /// Call the stl_repair from admesh rather than reimplementing it ourselves.
+    stl_repair(&(this->stl), 1, 1, 0, 0, 0, 0, 0, 100, 1, 1, 1, 1, 1, 0);
```
hroncok: The long list of arguments is very hard to read (yes, the API is not good), maybe add comment that explains the nonzero values?

lordofhyphens: Good idea. Fixed.

Issue #641, opened 2012-08-26
Feature request: adjustable height brim


ellenhp:
For those of us who print with small layer heights, one layer of brim often isn't enough to hold tall, top-heavy pieces to the bed. Adjusting the first layer height would also adjust brim height, I suspect, but it would be easier to get a nice thick brim on prints if we could tell Slic3r to make the brim multiple layers tall.

foulowl:
I would absolutely love to be able to do multi-layer brims as well, for adhesion. I get curling of corners when I use a single layer brim but less than with no brim. If I design the part with multi-layer "mouse ears" then things print much better, with less curling. Having the ability to print with a multi-layer brim would save me the trouble of having to add these mouse ears manually. Thank you!!

mswillia:
I was just looking for this feature as well. As mentioned before skirt height is adjustable.
You can set your skirt distance to 0, increase the loops, and set the height. The function would be the same as a multilayered brim.

lordofhyphens:
What is the point of a multilayer brim? Mostly curious, cleaning up 1 layer
brims is bad enough for me.

nopoe:
As of 2013 when I last used Slic3r, the brim is always one layer, regardless of layer height. Considering how wildly different layer height can be for different prints, I thought it might be worth having a multi-layer brim. This would possibly allow prints with small layer heights to have a brim that is just as strong as brims on prints with large layer heights. After some reflection during the time I owned my printer, I ultimately decided that it wouldn't matter for me, because I hit the layer height limits of my printer before I hit the limits of a single-layer brim. Perhaps it could help those with better printers.

foulowl:
I would use a multilayer brim for adhesion. Right now I add in a multilayer "brim" manually on my models, as a single layer brim isn't enough to prevent ABS parts from curling off the plate.

However, adding "mouse ears" manually definitely works.

gege2b:
Correct me if I'm wrong, but I don't think a multilayer brim would really help to get a better adhesion

More layers means more curling force from the ABS, it's a known effect

If an object begin to curl, then why a multilayer brim (witch is finally the same thing) wouldn't curl at all ?

Maybe a larger brim would help, but I didn't think a multilayer brim would.

ambrop72:
[...]

PR #4836, opened 2019-07-07
PR for multiple brims to exercise build agents

lordofhyphens:
DO NOT MERGE. The UI is not well-tested and this has not been proven to actually solve the underlying problem the feature was requested for.

### DIFF
--- a/xs/xsp/Print.xsp
+++ b/xs/xsp/Print.xsp
@@ -178,8 +178,13 @@ _constant()
     // TODO: status_cb
     Ref<ExtrusionEntityCollection> skirt()
         %code%{ RETVAL = &THIS->skirt; %};
-    Ref<ExtrusionEntityCollection> brim()
+    Ref<std::vector<ExtrusionEntityCollection> > brim()
         %code%{ RETVAL = &THIS->brim; %};
+
+    Ref<ExtrusionEntityCollection> brim_layer(int id)
+        %code%{ RETVAL = &(THIS->brim.at(id)); %};
+    size_t brim_count()
+        %code%{ RETVAL = THIS->brim.size(); %};
 
     PrintObjectPtrs* objects()
         %code%{ RETVAL = &THIS->objects; %};
@@ -288,7 +293,7 @@ _constant()
     Clone<Flow> brim_flow();
     Clone<Flow> skirt_flow();
     void make_skirt();
-    void _make_brim();
+    void _make_brim(size_t i);
 %{
 
 double
--- a/lib/Slic3r/GUI/3DScene.pm
+++ b/lib/Slic3r/GUI/3DScene.pm
@@ -1444,26 +1444,27 @@ sub load_print_toolpaths {
     }
     
     my $color_by_extruder = $self->color_toolpaths_by eq 'extruder';
-    
-    if (!$print->brim->empty) {
-        my $color = $color_by_extruder
-            ? $self->colors->[ ($print->brim_extruder-1) % @{$self->colors} ]
-            : $self->colors->[2];
-        
-        push @{$self->volumes}, my $volume = Slic3r::GUI::3DScene::Volume->new(
-            bounding_box    => $bb,
-            color           => $color,
-            qverts          => Slic3r::GUI::_3DScene::GLVertexArray->new,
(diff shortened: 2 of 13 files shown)


lordofhyphens:
Experimental implementation for #641

lordofhyphens:
@zapta @pun-e @ellenpoe @foulowl Experiment to your heart's content and figure out if it actually solves the underlying problem.

xs/xsp/Print.xsp
```diff
@@ -178,9 +178,14 @@ _constant()
     // TODO: status_cb
     Ref<ExtrusionEntityCollection> skirt()
         %code%{ RETVAL = &THIS->skirt; %};
-    Ref<ExtrusionEntityCollection> brim()
+    Ref<std::vector<ExtrusionEntityCollection> > brim()
```
lordofhyphens: this is probably useless now, but I am very bad at using XS to get a useful kind of reference.

lordofhyphens:
Windows https://ci.appveyor.com/api/buildjobs/f55y9i7574adq9bt/artifacts/Slic3r-master.2019.07.07.2347.fc28ba8f.64bit-PR.zip

ellenpoe:
I've long since sold my printer and am firmly on team close-that-original-issue. Thicker layers and support should solve most build issues and if appearance is an issue, there are other ways to solve that. Just my 2c. Unfollowing since I don't have much else to offer here.

Pun-e:
So I've just received some new parts for my printer and I'm planning on printing them all in PC. I downloaded the provided test branch, can see the new brim height option, which is working as expected, but it seems to be very "fragile":
image
The errors don't persist into a fresh launch, but there's definitely something going on that's causing issues with the skirt / brim generator. When I want to slice a file I have to relaunch it every time or I'll get all kinds of strange behaviour.
First load:
image
But if I change any parameter:
image
Not sure if it's just a difference in expected behaviour between the Prusa and main branches. I use background processing.

Pun-e:
[MC_Shroud_Gcode.zip](https://github.com/slic3r/Slic3r/files/3384584/MC_Shroud_Gcode.zip)

This is a copy of a good gcode run, and a dodgy one. Seems like any change in the layout, or in skirt /brim options causes it to happen. This makes it incredibly difficult to use this release as after every change I need to close, re-add pieces, re-arrange, slice. If anything is NQR, repeat the process.

The brims themselves are functionally everything I was hoping for.

lordofhyphens:
I did specifically say it was a hack.

It is probably having to do with the fact that the slicing pathway wasn't written with multiple layer brims in mind so I didn't get all of the places that assumed it.

lordofhyphens:
@Pun-e export the plate with modifiers as 3mf or amf. It will be less of a headache to experiment with.

lordofhyphens:
And to reiterate, the units for this will *always* be in layers. If someone needs it in mm they can refactor the thing to get what they want, because getting the units to be in mm is way more work than I am willing to put into it.

Pun-e:
@lordofhyphens 
So, I've spent a few hours and come across a few major errors:
- Sending to my networked printer will fail if the print is to big (unsure if this is standard slic3r bug)
- Moving / changing the parts after initial slice will fail
- Changing any brim / skirt option after initial slice will fail

The only one I really need to be functional is the "cannot slice more than once due to moving parts". I have all my settings dialled in, start, load parts, slice. But if the brim, or skirt is too far, or some parts look iffy all I can do is try again from scratch. If I could just get more than one slice per load this would be more than enough for the limited times that I need this feature.

I've also noticed that it prints all the brims before printing the part, which was surprising behaviour. Is that intended or is it a bug?

lordofhyphens:
Normal behavior re: brims printing first. Brims are generated first. It is not appropriate to request a change to that behavior here (out of scope).

Pun-e:
> 
> 
> Normal behavior re: brims printing first. Brims are generated first. It is not appropriate to request a change to that behavior here (out of scope).

I'm sorry, I don't understand how this relates?

eliliam:
Any updates on this? This would be an amazing feature to add, especially for the case mentioned above where the contact point between the brim and model is very weak, and adding additional layers may help it adhere better.

PR #4843, opened 2019-07-21
Allow fan to be specified from 0-100 (decimal) instead of 0-255

### LINKED ISSUE #4818: Provide option to generate fan control (M106) as a percentage duty cycle instead of 0-255
**Is your feature request related to a problem? Please describe.**
I am using Smoothieware as my 3d printer platform. In my case I am specifically using a pinn which generates hardware PWM, where the `S` value denotes the duty cycle in percent. In the Slic3r cooling control I can set the max value for the fan control, but it always assumes the `S` value for `M106` to be in a 0-255 scale. 

**Describe the solution you'd like**
Having an option to define (May be at printer level as the flavor of M106 is more board specific i.e. whether board is using software or hardware PWM than the filament specific) will be great for his case.

**Describe alternatives you've considered**
I can manually calculate actual maximum duty cycle allowed (example for a specific filament I want to run the cooling only up to 89%) using max%/255*100 (89/255*100=~35%) and entering that in the filament config section as max percentage. 

**Additional context**
My current solution works, but that makes the filament configuration not portable. I can't share my filament config with anyone who is not using same setup of hardware PWM like mine. Before sharing I have to remember to update the config to 89% in my example.

lordofhyphens:
Fixes #4818

### DIFF
--- a/lib/Slic3r/GUI/PresetEditor.pm
+++ b/lib/Slic3r/GUI/PresetEditor.pm
@@ -1251,6 +1251,7 @@ sub title { 'Printer Settings' }
 sub options {
     return qw(
         bed_shape z_offset z_steps_per_mm has_heatbed
+        fan_percentage
         gcode_flavor use_relative_e_distances
         serial_port serial_speed
         host_type print_host octoprint_apikey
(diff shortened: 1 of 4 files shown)


supermerill:
A bit late, but why not transforming the ConfigOptionFloat of the fan to ConfigOptionFloatOrPercent
 (and using it with get_abs_value(1)) ?

Issue #4826, opened 2019-06-16
crash after reload model


nesergen:
**Describe the bug**
GUI crash after reload rotated model from disk.

**To Reproduce**
1. load model
2.  right click on model, rotate model for 90 deg around Y axis
3. right click on model, select "Reload from disck"
4. model updated, then GUI crash. It needs to restart program.

**3D Model and Slic3r Configuration Export**
[BathHolder.stl.zip](https://github.com/slic3r/Slic3r/files/3294495/BathHolder.stl.zip)

**Expected behavior**
Reload model withouy crash

**Desktop (please complete the following information):**
 - OS: Linux Mint 18
 - Version 1.3.0
  

**Additional context**
Another wish. When you update the model, all changes to the model position are reset. That is, the model returns to its original state. It would be desirable that the reloaded model kept the added changes .

Oekn5w:
Regarding your additional: I'm currently implementing transformation matrices, which will carry the transformation over to the reloaded object. As I am the last person who touched the reload code I will look into this when I have time.

Oekn5w:
I was able to recreate the issue both in downloaded compiled version 1.3.0 and built from the 1.3.0 tag from git (Windows). The crash occurs without anything printed to the console.
But: The crash only occured with the file included in the description. Other STLs (both ascii and binary) I had lying around reload fine in 1.3.0.

I specify the version explicitly because 1.3.1dev (built from commit 48d5b1130130373e990dd40cc811b252556edcb4 ; didn't check auto builds) does not crash with the given file.

@nesergen how did you create / where did you get that file from?

The rotation is irrelevant to the crash.

nesergen:
I use the Onshape.com online CAD service.
--
Отправлено из Mail.ru для Android среда, 26 июня 2019г., 11:05 +03:00 от Michael Kirsch  notifications@github.com :

Oekn5w:
TBH I don't want to create an account to check what might be the root cause for this issue, especially since the development version works.

Oekn5w:
and for cross referencing about what I said earlier regarding the rotation: #4401

nesergen:
Anyway, thank you for your time.
--
Отправлено из Mail.ru для Android среда, 26 июня 2019г., 13:10 +03:00 от Michael Kirsch  notifications@github.com :

Oekn5w:
@lordofhyphens imo this can be closed

lordofhyphens:
Can't reproduce?

Issue #4401, opened 2018-05-05
[Feature request] Transformation matrices to keep history of mesh changes


Oekn5w:
### Version
1.3.0 dev

### Problem
Reload from disk dismisses non-Z rotations and translations applied in the settings menu, as these are applied to the mesh itself without any history apart from undo stack (not viable to read from in reload)

### Suggested adaptation
* Create class to contain a transformation matrix and the required math operations
* Assign one matrix to each mesh, volume and object, maybe ModelInstance
* Changes in various parts of GUI: Instancing, PartObjectDialog
* Changes in backend: Undo/ Redo, ModelObject, ModelVolume, TriangleMesh

Oekn5w:
@lordofhypens I'm almost done with the implementation but I wonder whether I should also adapt the C++ GUI especially since I don't really know how to build with that part.

lordofhyphens:
You can probably leave it be as I don't have the part editor implemented yet.

Oekn5w:
Alright, I'll wrap it up and open a PR soon(TM)

Issue #4269, opened 2018-01-22
"Reload from Disk" does not work with network shares


dartrax:
### Version
19ea7a5

### Operating system type + version
Win 7 64bit

### Behavior
"reload from Disk" does simply nothing when the current stl model is located on a network share. Maybe "disk" is taken too literally... ;-)

### Steps to reproduce:
1. Save a .stl to network, something like: \\\\SATCAD01\Users\stl_3D-Druck\35052608.stl
2. Add to Slicer
3. Update .stl on network
4. Right click > Reload from Disk (nothing happens, not even status error message)

foreachthing:
Works on my Windows 7 64bit and Slic3r-master.2017.12.18.1561.d27b8a9.64bit
Added STL from `\\server\dir\dir\file.stl`, changed it, reloaded it in slic3r, then got new file.

dartrax:
Thank you @foreachthing for testing. I investigated further and the reason was that I splitted the model. I've updated the reproduce steps and the topic title, because the old one is misleading now.

foreachthing:
Jup! Now it does nothing :-)

lordofhyphens:
So we're expecting that after you make changes to the model in-memory (split is a change), hitting reload would discard everything and do a fresh load?

dartrax:
Yes.
When I have an updated model, every way to get this model into slicer leads to a point where I have to reiterate through all changes that I've made to this model, like positioning, rotation, cutting or splitting.
In fact, everything is already lost at the point where I decide that I have to update the model.

The perfect solution would probably to record all modifications that I've made to a model and to apply them automatically again after updating the model.

Positioning and rotations are being discarded as well right now and as end user I'd never make a difference between rotating the model and splitting it. I'm used to the fact that I lose positioning and rotating when reloading, so I expect to lose splitting and cutting, too. What I would not expect is that reloading doesn't work at all and does not give me a hint, why.

PR #4845, opened 2019-07-21
Transformation matrix class - preserve transformations during reload

Oekn5w:
Corresponding feature request: #4401 (sidenoted in #4269, #4826 )

As it turned out, the necessary modifications span many classes, predominantly Model and its subclasses , the IO classes and the admesh utility.

The UI doesn't change apart from the option added to the reload preferences, where you can choose whether you want to try and preserve the transformations or not.

All the transformations are now normalized to result in a general purpose matrix transformation. The volumes' meshes are changed as before, but the volumes' TransformationMatrix property keep track of the transformations applied to their corresponding meshes.

On reload, the meshes are reloaded and the matrices applied when the corresponding options are selected.

As all geometric manipulations now result in matrices, those are used in the undo / redo stack as opposed to (multiple) rotations / scales / translations.

This PR is rounded off by a test file for the new class.


Oekn5w:
I don't know why the 64-bit inverse test passes, but fails in 32-bit.

Oekn5w:
@lordofhyphens thoughts? Is this PR too big? I didn't really think it would make sense to just add the class without adding the base functionality.

lordofhyphens:
Sorry, got distracted and will look at it in a bit

xs/src/admesh/util.c
```diff
@@ -185,24 +185,74 @@ void calculate_normals(stl_file *stl) {
...
     return;
   for (i_face = 0; i_face < stl->stats.number_of_facets; ++ i_face) {
     stl_vertex *vertices = stl->facet_start[i_face].vertex;
     for (i_vertex = 0; i_vertex < 3; ++ i_vertex) {
       stl_vertex* v_dst = &vertices[i_vertex];
-      stl_vertex  v_src = *v_dst;
-      v_dst->x = trafo3x4[0] * v_src.x + trafo3x4[1] * v_src.y + trafo3x4[2]  * v_src.z + trafo3x4[3];
-      v_dst->y = trafo3x4[4] * v_src.x + trafo3x4[5] * v_src.y + trafo3x4[6]  * v_src.z + trafo3x4[7];
-      v_dst->z = trafo3x4[8] * v_src.x + trafo3x4[9] * v_src.y + trafo3x4[10] * v_src.z + trafo3x4[11];
+      double v_src_x = (double)(v_dst->x);
```
lordofhyphens: Was there a reason you needed to modify admesh? This is technically code external to the Slic3r project and proposed changes should also be pushed upstream.

lordofhyphens:
Thank you very much for the work; I've done some UI testing of it and it looks fine to me. Sorry it took me this long to get to it.

xs/t/25_transformationmatrix.t
```diff
@@ -0,0 +1,156 @@
+#!/usr/bin/perl
```
lordofhyphens: Can you port this to the C++ tests after merging?

xs/src/admesh/util.c
Oekn5w: as this is now the remaining function that is called in admesh, I wanted it it to be efficient wrt the checks done on each transform (size, volume, faces) and to have and option to directly assign a transformed mesh. I'll look into purposing the corresponding upstream change.

xs/t/25_transformationmatrix.t
Oekn5w: sure, I'll open another PR when I'm done

PR #4866, opened 2019-09-22
xs/Build.pl: fix typo BOOST_INCLUDEPATH -> BOOST_INCLUDEDIR

jpnurmi:


lordofhyphens:
Why is `BOOST_INCLUDEDIR` more correct?

jpnurmi:
`BOOST_INCLUDEDIR` is used in the build script:
https://github.com/slic3r/Slic3r/blob/master/xs/Build.PL#L89

lordofhyphens:
Yes, but why is your change correct?

jpnurmi:
Would you prefer to keep the error message intact and change the build script to use ```BOOST_INCLUDEPATH``` instead?

PR #4876, opened 2019-10-22
some more tests

supermerill:
Hi

I have written some tests that could make their way here.
I didn't had the time to test them since this merge, but i will do when possible.

### DIFF
--- a/src/test/libslic3r/test_extrusion_entity.cpp
+++ b/src/test/libslic3r/test_extrusion_entity.cpp
@@ -1,6 +1,9 @@
 #include <catch.hpp>
+#include "test_data.hpp" // get access to init_print, etc
+#include "Config.hpp"
 #include "ExtrusionEntityCollection.hpp"
 #include "ExtrusionEntity.hpp"
+#include "GCodeReader.hpp"
 #include "Point.hpp"
 #include <cstdlib>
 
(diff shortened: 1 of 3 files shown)


lordofhyphens:
Thanks!

lordofhyphens:
@supermerill I'm seeing some failures in the build output, was that expected?

src/test/libslic3r/test_extrusion_entity.cpp
```diff
@@ -27,12 +30,13 @@ ExtrusionPaths random_paths(size_t count=10, size_t length=20, float LO=-50, flo
...
 
 using namespace Slic3r;
+using namespace Slic3r::Test;
 
 SCENARIO("ExtrusionEntityCollection: Polygon flattening") {
     srand(0xDEADBEEF); // consistent seed for test reproducibility.
 
     // Generate one specific random path set and save it for later comparison
-    ExtrusionPaths nosort_path_set {random_paths()};
+    ExtrusionPaths nosort_path_set{ random_paths() };
```
lordofhyphens: Was this the result of an automatic code formatter?

src/test/libslic3r/test_print.cpp
```diff
@@ -138,8 +138,18 @@ SCENARIO("Print: Brim generation") {
             config->set("first_layer_extrusion_width", 0.5);
             auto print {Slic3r::Test::init_print({m}, model, config)};
             print->make_brim();
-            THEN("Brim Extrusion collection has 12 loops in it") {
-                REQUIRE(print->brim.items_count() == 12);
+            double nbLoops = 6.0 / print->brim_flow().spacing();
```
lordofhyphens: Why the change here for the extrusion collection?

src/test/libslic3r/test_extrusion_entity.cpp
supermerill: probably

src/test/libslic3r/test_print.cpp
supermerill: I made it that way because I wanted it to check if it makes the correct number of loop and not checking if the spacing is exactly what we tough about it (the layer heigh isn't set, btw)
I should had changes other tests also, i think that only this one bother me at the time because it was so large, a little difference in spacing can change the line count.

PR #4906, opened 2019-12-22
cpp porting: TransformationMatrix class tests

Oekn5w:
Requested in #4845


lordofhyphens:
@Oekn5w can you rebase this onto master? I've made some upstream changes that should sort out spurious build failures.

Oekn5w:
@lordofhyphens done and succeeded

src/test/libslic3r/test_transformationmatrix.cpp
```diff
@@ -0,0 +1,219 @@
...
+    equal &= std::abs(matrix.m20 - m20) < EPSILON;
+    equal &= std::abs(matrix.m21 - m21) < EPSILON;
+    equal &= std::abs(matrix.m22 - m22) < EPSILON;
+    equal &= std::abs(matrix.m23 - m23) < EPSILON;
+    return equal;
+}
+
+bool check_point(const Pointf3 & point, coordf_t x, coordf_t y, coordf_t z)
+{
+    const coordf_t EPSILON = 1.0e-3;
```
lordofhyphens, line 213: Does this fail using libslic3r's epsilon constant?

src/test/libslic3r/test_transformationmatrix.cpp
```diff
@@ -0,0 +1,219 @@
+#include <catch.hpp>
+
+#include <vector>
+
+#include "libslic3r.h"
+#include "TransformationMatrix.hpp"
+
+#include "Log.hpp"
```
lordofhyphens, line 8: Unnecessary include

src/test/libslic3r/test_transformationmatrix.cpp
Oekn5w: Probably not, I think I relaxed that threshold as it compares to the manual values given, not to the numerical precision. Bad naming from me, rename to EQUALITY_THRESHOLD (defined in the header) or adapt to the EPSILON (whose comments suggest only being used for floating-point precision stuff)?

Oekn5w: I pushed the threshold definition version

src/test/libslic3r/test_transformationmatrix.cpp
Oekn5w: removed, good catch

PR #4919, opened 2020-01-17
Add G-code export for M3D Micro

natanoj:
Add a new printer profile for M3D Micro which needs minor differences to all the existing printer profiles to print the generated G-code. Builds with all existing tests passing on my Debian Buster. Have tested a few printouts which works on the M3D Micro.

### DIFF
--- a/xs/src/libslic3r/GCodeWriter.cpp
+++ b/xs/src/libslic3r/GCodeWriter.cpp
@@ -67,7 +67,9 @@ GCodeWriter::preamble()
     std::ostringstream gcode;
 
     if (FLAVOR_IS_NOT(gcfMakerWare)) {
-        gcode << "G21 ; set units to millimeters\n";
+        if (FLAVOR_IS_NOT(gcfM3dMicro)) {
+            gcode << "G21 ; set units to millimeters\n";
+        }
         gcode << "G90 ; use absolute coordinates\n";
     }
     if (FLAVOR_IS(gcfRepRap) || FLAVOR_IS(gcfTeacup) || FLAVOR_IS(gcfRepetier) || FLAVOR_IS(gcfSmoothie)) {
--- a/xs/src/libslic3r/PrintConfig.cpp
+++ b/xs/src/libslic3r/PrintConfig.cpp
@@ -666,6 +666,7 @@ PrintConfigDef::PrintConfigDef()
     def->enum_values.push_back("mach3");
     def->enum_values.push_back("machinekit");
     def->enum_values.push_back("smoothie");
+    def->enum_values.push_back("m3dmicro");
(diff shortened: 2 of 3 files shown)


xs/src/libslic3r/GCodeWriter.cpp
```diff
@@ -67,7 +67,9 @@ GCodeWriter::preamble()
     std::ostringstream gcode;
 
     if (FLAVOR_IS_NOT(gcfMakerWare)) {
-        gcode << "G21 ; set units to millimeters\n";
+        if (FLAVOR_IS_NOT(gcfM3dMicro)) {
+            gcode << "G21 ; set units to millimeters\n";
```
lordofhyphens: So that printer firmware does something weird for g21?

But you need G90 in the defaults?

natanoj: Yes, that printer does not accept G21. More specifically, it doesn't accept other combinations of specific printer profiles in the export code either. I went through them one by one, and each profile adds a g-code or syntax that this simple printer does not accept unfortunately. 

A better, more generic, way would perhaps be to have a generic default, which would be similar to this new profile, and have a specific option for the G21.

While I was looking through the export code, I noticed that it bocomes more cluttered the more printers are added. For that reason I don't like my own change, but it gets the job done for me.

natanoj: I went through the Spooler logs from the M3D supplied spooler, and in there I can see that the printer identifies as a Repetier protocol 2 firmware. However, when looking at the generated G-code from the supplied Cura engine, it does not fully match what is generated with the Repetier setting in Slic3r. The printer returns an error code for G21, and other G-codes for mainly extrusion does not match the M3D generated G-code. 

I'll make a test with the Repetier setting, and manually remove the G21 to see if that works. The G21 is still problematic though.

natanoj: Tested G-code generated by the Repetier export. With the G21 left in, the printer just stops at that command. Without G21 it stops after warming up. It is interesting though that when I run the exact same sequence of generated G-codes manually without G21 (i.e. with a lot of time in between each command) they work in the generated sequence. I suspect it is the M82 directly after G90 that causes the hickup. So the most reliable setting for this printer is with this small addition. 

Of course, still up to you if this is something that should be included or not. At least now I have a setup that works for me for this printer.

lordofhyphens: I suspect that the better long term solution is to have a "generic" flavor that lets the user provide a file that says "use X to do task Y".
Leaving an entry blank would be enough to say "my printer can't do this, so skip"

Issue #4174, opened 2017-10-28
Low brim to perimeter bonding


TheBoojah:
### Version
1.3.0 dev oct 23

### Operating system type + version
Win 10 64bit latest patch

### Behavior
The line spacing between the bottom outer perimiter and the innermost brim line is too wide. This is especially noticeable if a print is done with z-height adjustment so that the printed bottom layer is exactly or slightly taller than the set first layer height (for printing with high-stick build plate material for example). What happens is that the lines of the brim and the part adhere well to each other, but the brim is not attached at all or weakly to the part - rendering the brim useless.

1.3.0 dev:
![image](https://user-images.githubusercontent.com/33135012/32133560-b9ca83dc-bbda-11e7-9216-a1ef1e0cb009.png)
Note the gap between the green brim lines and the part.

Correct results are obtained on 1.2.9:
![image](https://user-images.githubusercontent.com/33135012/32133567-e0c41354-bbda-11e7-8484-218638fae4e8.png)

#### STL/Config (.ZIP) where problem occurs
[Low brim to perimeter bonding.zip](https://github.com/alexrj/Slic3r/files/1424143/Low.brim.to.perimeter.bonding.zip)

lordofhyphens:
I'm thinking it's fp math somewhere. 

If you want to play around with the brim distance, see line 863 in Print.cpp. 

It shifts the loops in by half a scaled_width, so vary the value between 0 and -1. Positive values would put it further away.

lordofhyphens:
![image](https://user-images.githubusercontent.com/31754/32140407-a7c6b1b2-bc2d-11e7-9d21-b845ac48e542.png)

That's with a -0.55 modifier, which would provide a more firm connection. It'd probably do as a workaround for your immediate needs.

lordofhyphens:
I've pushed a workaround.

paulfertser:
#2230 seems to be related here. Apparently the connection was nice and strong before af46db320f820cc68e692cccf2f83890f4ee1596 and in this ticket the workaround is not really a revert of it. Was this considered? I would expect strong brim connection to be the default behaviour and those who need to separate it more (like the #2230 reporter) should get a configuration option.
(just had one ABS part lifted off bed together with internal brim, with outer brim still there on the glass)

lordofhyphens:
I suspect we can't get around a config option now for brims distance (interior and exterior). I was really hoping to avoid it.

lordofhyphens:
[...]

PR #4957, opened 2020-04-18
Add brim_overlap option to allow control of brim bonding strength. Issue #4174.

ianmackinnon:
This is my first pull request. Please let me know if there is anything in terms of content or protocol that I should do differently.

The value I have specified as the default (-3.13%) maintains the current brim distance although it's not a very round number.

slic3r-branch-brim-overlap -3 13 default

I have added a “default” option in `PrintConfig.cpp` but the default value still shows as `-3.13%` in the Print Settings panel. I'm not sure what I would have to change to make this show `default` instead.

### DIFF
--- a/src/test/libslic3r/test_skirt_brim.cpp
+++ b/src/test/libslic3r/test_skirt_brim.cpp
@@ -199,6 +199,23 @@ SCENARIO("Original Slic3r Skirt/Brim tests", "[!mayfail]") {
             }
         }
 
+        WHEN("brim overlap") {
+            config->set("skirts", 0);
+            config->set("first_layer_extrusion_width", 0.5);
+            config->set("brim_width", 1);
+            config->set("brim_overlap", 15);
+
+            Slic3r::Model model;
+            auto print {Slic3r::Test::init_print({TestMesh::cube_20x20x20}, model, config)};
+            print->process();
+
+            THEN("Gcode generates") {
+                Slic3r::Test::gcode(gcode, print);
+                auto exported {gcode.str()};
+                REQUIRE(exported.size() > 0);
+            }
+        }
+
         WHEN("Object is plated with overhang support and a brim") {
             config->set("layer_height", 0.4);
             config->set("first_layer_height", 0.4);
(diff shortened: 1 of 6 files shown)


lordofhyphens:
@ianmackinnon please rebase. 

Additionally, once you've saved a config the defaults aren't going to show for you.

src/test/libslic3r/test_skirt_brim.cpp
```diff
@@ -199,6 +199,23 @@ SCENARIO("Original Slic3r Skirt/Brim tests", "[!mayfail]") {
...
+            Slic3r::Model model;
+            auto print {Slic3r::Test::init_print({TestMesh::cube_20x20x20}, model, config)};
+            print->process();
+
+            THEN("Gcode generates") {
+                Slic3r::Test::gcode(gcode, print);
+                auto exported {gcode.str()};
+                REQUIRE(exported.size() > 0);
+            }
+        }
```
lordofhyphens, line 217: This is a really basic test -- would really prefer testing that the spacing is what is expected. 

Perhaps some parsing of the GCode to look at the brim lines?

lordofhyphens:
@ianmackinnon see the review comment, please.

lordofhyphens:
@ianmackinnon https://github.com/slic3r/Slic3r/pull/4997 has some code around testing the brim generation itself (it also fixes a minor off-by-one error I found when writing the test).

It looks like you're pushing the entire brim out by that little amount; I think that referring to the gap in some small absolute unit like mm or maybe um would be more understandable by users than percentage, as you're basically changing the inflate/deflate scale numbers. 

@VanessaE any thoughts about this?

ianmackinnon:
@lordofhyphens Thanks for the feedback and linking that issue with the relevant test. When I get a moment I'll rebase and add a proper test.

I imagine that the ability to control the gap will be important only to a small subset of users, so in my opinion the priority should be to not alter current behaviour. -3.13% is not a neat number but it is the value that matches current behaviour and functionally I think this default in the existing code works great for the vast majority of prints. I also think it makes sense for the size of the gap to scale with the print width rather than be an absolute value.

So my preference would be to stick with a percentage and keep that number as the default but if others feel strongly it should have a different unit or default value I'm happy to try it out.

VanessaE:
I favor an exact measure for this e.g. mm.  µm might be okay, though I guess it would be the only to use that.

PR #4974, opened 2020-06-02
Do not undefine __STRICT_ANSI__

jwakely:
The `__STRICT_ANSI__` macro is defined by the compiler and you should not undefine or redefine it.

Using `-U__STRICT_ANSI__ -std=c++11` is just silly. If you don't want strict mode, don't ask for strict mode. Certainly don't ask for strict mode and then undefined the macro that is defined by strict mode.

The correct solution is `-std=gnu++11` which doesn't define the macro in the first place.


lordofhyphens:
Merging as the failing jobs aren't relevant to this RP.

PR #4976, opened 2020-06-02
Fedora patches for Boost 1.73 compatibility

hroncok:
Fixes https://github.com/slic3r/Slic3r/issues/4967

Those are 2 of 3 Fedora patches created to be able to build with Boost 1.73.

The third is obsoleted by already existing ce966c2450183dfc1c6415161463ea8d59dcccd5.
￼
Authored by @jwakely. Only tested on 1.3.0 (and I am afraid I won't be able to test this on master any time now, I've simply rebased it).

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -114,7 +114,7 @@ endif(NOT GIT_VERSION STREQUAL "")
 
 find_package(Threads REQUIRED)
 
-find_package(Boost REQUIRED COMPONENTS system thread filesystem)
+find_package(Boost REQUIRED COMPONENTS system thread filesystem nowide)
 
 set(LIBDIR ${CMAKE_CURRENT_SOURCE_DIR}/../xs/src/)
 set(GUI_LIBDIR ${CMAKE_CURRENT_SOURCE_DIR}/GUI/)
--- a/xs/Build.PL
+++ b/xs/Build.PL
@@ -152,7 +152,7 @@ if (defined $ENV{BOOST_LIBRARYPATH}) {
 }
 # In order to generate the -l switches we need to know how Boost libraries are named
 my $have_boost = 0;
-my @boost_libraries = qw(system thread filesystem);  # we need these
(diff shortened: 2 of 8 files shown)


lordofhyphens:
@hroncok does this patch force the minimum boost version to 1.73 then?

hroncok:
I ma unsure yet. Might add some ifdefs around it.

lordofhyphens:
@hroncok while I'm not going to object to inching up the minimum Boost requirement, it does involve reconfiguring and updating all of my agents. 

I also know that even Debian sid still only has 1.71 available, so if it's something that changed explicitly in 1.73 then there's a low chance that I can accept it w/o workarounds to support the older version.

hroncok:
Completely understood.

jwakely:
The changes related to bind placeholders do not require a newer Boost. Those changes should work fine with old Boost versions as well as new.

Adding `nowide` to `CMakeLists.txt` and `xs/Build.PL` does require Boost 1.73.0 because that's the first release to include nowide. So you probably want a different fix which will work with either the standalone `nowide` library that you've been using until now, or the official one that's part of Boost 1.73.0.

lordofhyphens:
For cmake we just make a separate call to find_packages after looking for boost. Not a big deal. 

I don't know how to do that in Build.PL.

jwakely:
> The changes related to bind placeholders do not require a newer Boost. Those changes should work fine with old Boost versions as well as new.

Actually it turns out the `boost::placeholders` namespace has only existed since Bosot 1.60.0 (released December 2015). So not that old. You could do this though:

```
#include <boost/version.hpp>
#if BOOST_VERSION >= 106000
using namespace boost::placeholders;
#endif
```

lordofhyphens:
I think the oldest we build with is 1.60.0

lordofhyphens:
Travis CI builds use 1.63.0; I'm pretty sure both of our OSX builds are on 1.67.0 or newer.

lordofhyphens:
Closing in favor of #4996

hroncok:
Thank You!

PR #4996, opened 2020-07-23
Backwards-compatible Boost 1.73

### LINKED ISSUE #4967: Cannot compile when using Boost 1.73
### Version
git head, version 54a31eed204863c49dfaa3b6fe4809f6d10a94ba

Gentoo Linux unstable (rolling release date: May 9)

### Behavior
Cannot compile when using Boost 1.73.  Boost now requires bind.hpp to be included explicitly when using placeholder arguments.  Typical error resembles the following: 
src/libslic3r/PrintObject.cpp: In member function ‘void Slic3r::PrintObject::infill()’:
src/libslic3r/PrintObject.cpp:1214:49: error: ‘_1’ was not declared in this scope
 1214 |         boost::bind(&Slic3r::Layer::make_fills, _1),

The source files that need to be updated are:
src/libslic3r/SLAPrint.cpp
src/libslic3r/PrintObject.cpp
src/libslic3r/SupportMaterial.cpp
src/libslic3r/TriangleMesh.cpp

No STL or config is relevant, but I attached a patch file that fixes the problem.

[boost-headers.patch.gz](https://github.com/slic3r/Slic3r/files/4607849/boost-headers.patch.gz)

lordofhyphens:
Amend work submitted by @hroncok to allow for building on Boost 1.73 while maintaining compatibility for older versions.

Fixes #4967

### DIFF
--- a/xs/src/libslic3r/ConfigBase.cpp
+++ b/xs/src/libslic3r/ConfigBase.cpp
@@ -212,8 +212,7 @@ ConfigOptionDef::ConfigOptionDef(const ConfigOptionDef &other)
 
 ConfigOptionDef::~ConfigOptionDef()
 {
-    if (this->default_value != nullptr)
-        delete this->default_value;
+    delete this->default_value;
 }
 
 std::vector<std::string>
@@ -716,7 +715,7 @@ DynamicConfig::swap(DynamicConfig &other)
 DynamicConfig::~DynamicConfig () {
     for (t_options_map::iterator it = this->options.begin(); it != this->options.end(); ++it) {
         ConfigOption* opt = it->second;
-        if (opt != NULL) delete opt;
+        delete opt;
     }
 }
 
--- a/xs/src/libslic3r/ConfigBase.cpp
+++ b/xs/src/libslic3r/ConfigBase.cpp
@@ -7,6 +7,7 @@
 #include <sstream>
 #include <exception> // std::runtime_error
 #include <set>
+#include <boost/version.hpp>
 #include <boost/algorithm/string.hpp>
 #include <boost/algorithm/string/classification.hpp>
 #include <boost/algorithm/string/erase.hpp>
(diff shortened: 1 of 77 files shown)


lordofhyphens:
@jwakely @hroncok could you check that this still builds fine on Fedora? I should have gated things appropriately.

hroncok:
Attempt 1: https://koji.fedoraproject.org/koji/taskinfo?taskID=47680738

PR rebased to attached patch: [4996.patch](https://github.com/slic3r/Slic3r/files/4965121/4996.patch.txt)

xs/src/libslic3r/ConfigBase.cpp
```diff
@@ -16,7 +17,11 @@
 #include <boost/config.hpp>
 #include <boost/foreach.hpp>
 #include <boost/lexical_cast.hpp>
+#ifdef BOOST_NOWIDE_FOUND
+#include <boost/nowide/cstdlib.hpp>
+#else
 #include <boost/nowide/cenv.hpp>
+#endif
```
hroncok, line 24: I might have screwed up the rebase, but here it fails with:

```
src/libslic3r/Config.cpp:20:10: fatal error: boost/nowide/cenv.hpp: No such file or directory
   20 | #include <boost/nowide/cenv.hpp>
      |          ^~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

jwakely: The attached 4996.patch above modifies that file, but the spec file uses https://github.com/slic3r/Slic3r/pull/4996.patch and that one doesn't modify `src/libslic3r/Config.hpp`

hroncok: The url in spec is fake, I was not doing it properly.

krop:
I tested the changes on openSUSE with master & boost 1.74, an error remains in xs/src/libslic3r/SupportMaterial.cpp:56 ('_1' was not declared in this scope)

PR #5010, opened 2020-08-24
Switch to XCode 11.6 so that we can notarize (required for Catalina).

lordofhyphens:


supermerill (approved):
I don't have a mac, so do what needs to be done.

Issue #5042, opened 2021-01-17
Remove unnecessary null pointer checks


elfring:
[An extra null pointer check is not needed in functions](https://isocpp.org/wiki/faq/freestore-mgmt#delete-handles-null "Do I need to check for null before delete p?") like the following.
- [AvoidCrossingPerimeters](https://github.com/slic3r/Slic3r/blob/92abbc42dfdd5385c1f9c3a450e2f3da835f8b8d/xs/src/libslic3r/GCode.cpp#L17 "Destructor for AvoidCrossingPerimeters")
- [ConfigOptionDef](https://github.com/slic3r/Slic3r/blob/92abbc42dfdd5385c1f9c3a450e2f3da835f8b8d/xs/src/libslic3r/ConfigBase.cpp#L208 "Destructor for ConfigOptionDef")
- [Option](https://github.com/slic3r/Slic3r/blob/92abbc42dfdd5385c1f9c3a450e2f3da835f8b8d/src/GUI/OptionsGroup.hpp#L34 "Destructor for Option")

roy-stewart:
I second this. It won't make a functional difference but it will definitely clean up the code base. I'm working on a pull request to remove these null pointer checks now.

lordofhyphens:
@roy-stewart I will review it.

elfring:
How do you think about to [use a development tool like “clang-tidy” for corresponding source code adjustments](https://clang.llvm.org/extra/clang-tidy/checks/readability-delete-null-pointer.html "Deletion of unnecessary null pointer checks")?

lordofhyphens:
Merged, thank you.

I currently don't want to deal with the diff from clang-tidy or agonize over the ruleset.

PR #5049, opened 2021-01-28
Removed null checks before calls to delete and free

roy-stewart:
Implemented changes mentioned in #5042. Removed null checks before calls to delete or free since both are null safe.


lordofhyphens (approved):

elfring:
Thanks for your small source code improvement.

PR #5063, opened 2021-03-15
Fix CVE-2020-28591

### LINKED ISSUE #5061: AMF File AMFParserContext::endElement() out-of-bounds read vulnerability (TALOS-2020-1215, CVE-2020-28591)
Hi

https://talosintelligence.com/vulnerability_reports/TALOS-2020-1215 (CVE-2020-28591) refers to an issue in Slic3r but looking at the recents I did not spot a fix.

Is this issue open yet or can you point to the fixing commit?

lordofhyphens:
Don't add any facets to the read-in list if it calls out a vertex that is more than we have read in.

(assumes that we have the vertex list before the volumes, which should be guaranteed in a normal file).
Fixes #5061

### DIFF
--- a/src/test/libslic3r/test_amf.cpp
+++ b/src/test/libslic3r/test_amf.cpp
@@ -1,0 +1,28 @@
+#include <catch.hpp>
+#include <test_options.hpp>
+#include "Model.hpp"
+#include "IO.hpp"
+
+
+using namespace Slic3r;
+
+SCENARIO("Reading AMF file") {
+    GIVEN("badly formed AMF file (missing vertices)") {
+        auto model {new Slic3r::Model()};
+        WHEN("AMF model is read") {
+            auto ret = Slic3r::IO::AMF::read(testfile("test_amf/5061-malicious.xml"),model);
+            THEN("read should return True") {
+                REQUIRE(ret);
+            }
+        }
+    }
+    GIVEN("Ok formed AMF file") {
+        auto model {new Slic3r::Model()};
+        WHEN("AMF model is read") {
+            auto ret = Slic3r::IO::AMF::read(testfile("test_amf/read-amf.xml"),model);
+            THEN("read should return True") {
+                REQUIRE(ret);
+            }
+        }
+    }
+}
--- a/xs/src/libslic3r/IO/AMF.cpp
+++ b/xs/src/libslic3r/IO/AMF.cpp
@@ -344,9 +344,13 @@ void AMFParserContext::endElement(const char *name)
     // Faces of the current volume:
     case NODE_TYPE_TRIANGLE:
         assert(m_object && m_volume);
-        m_volume_facets.push_back(atoi(m_value[0].c_str()));
-        m_volume_facets.push_back(atoi(m_value[1].c_str()));
-        m_volume_facets.push_back(atoi(m_value[2].c_str()));
+        if (strtoul(m_value[0].c_str(), nullptr, 10) < m_object_vertices.size() &&
+            strtoul(m_value[1].c_str(), nullptr, 10) < m_object_vertices.size() &&
+            strtoul(m_value[2].c_str(), nullptr, 10) < m_object_vertices.size()) {
+            m_volume_facets.push_back(atoi(m_value[0].c_str()));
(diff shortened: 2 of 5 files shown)


hyperair (approved):
Looks good to me

src/test/libslic3r/test_amf.cpp
```diff
@@ -0,0 +1,28 @@
...
+
+
+using namespace Slic3r;
+
+SCENARIO("Reading AMF file") {
+    GIVEN("badly formed AMF file (missing vertices)") {
+        auto model {new Slic3r::Model()};
+        WHEN("AMF model is read") {
+            auto ret = Slic3r::IO::AMF::read(testfile("test_amf/5061-malicious.xml"),model);
+            THEN("read should return True") {
```
hroncok, line 14: Why should this return true and not false?

lordofhyphens: I strip the illegal facets instead of erroring out.

xs/src/libslic3r/IO/AMF.cpp
```diff
@@ -363,8 +367,9 @@ void AMFParserContext::endElement(const char *name)
         stl_allocate(&stl);
         for (size_t i = 0; i < m_volume_facets.size();) {
             stl_facet &facet = stl.facet_start[i/3];
-            for (unsigned int v = 0; v < 3; ++ v)
+            for (unsigned int v = 0; v < 3; ++ v) {
```
dwillmore, line 370: The is just a whitespace change was it leftover from other changes to this loop that ended up not being necessary?

dwillmore (approved):

src/test/libslic3r/test_amf.cpp
hroncok: I see. OK

hroncok (approved):
The unrelated `{}` change is not relevant and I'd personally not mic it in here. Other than that, this fixes the probelm but I am not entirely sure the test actually fails without the fix (and unfortunatelly I am unable to try that out soon).

xs/src/libslic3r/IO/AMF.cpp
lordofhyphens: Yes, that is the case. I'll revert.

hyperair (approved):

lordofhyphens:
> The unrelated `{}` change is not relevant and I'd personally not mic it in here. Other than that, this fixes the probelm but I am not entirely sure the test actually fails without the fix (and unfortunatelly I am unable to try that out soon).

The way I had seen it, throwing an exception was more testable and correct, but our users are going to prefer that it tries to drop the data.

hroncok:
Looks good. A nitpick: Why is the file not named `.amf`? I needed to rename it to load it up by slic3r GUI.

lordofhyphens:
> Looks good. A nitpick: Why is the file not named `.amf`? I needed to rename it to load it up by slic3r GUI.

Just me derping.

hroncok:
Thank you!

hroncok:
I've applied this patch in Fedora.

When I take this AMF file https://github.com/openscad/openscad/blob/master/testdata/amf/sphere20face.amf I was able to load it before this CVE fix and I am able to load it afterwards.

When I change the AMF file to have `<triangle><v1>999999</v1>...</triangle>` it crashes slic3r before this CVE fix. With this CVE fix, slic3r loads the file and doe snot crash :tada: 

However, when I try to open either of the two test files from this PR, prior to this fix or after this fix, I get:

```
/usr/include/c++/10/bits/stl_vector.h:1045: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](std::vector<_Tp, _Alloc>::size_type) [with _Tp = double; _Alloc = std::allocator<double>; std::vector<_Tp, _Alloc>::reference = double&; std::vector<_Tp, _Alloc>::size_type = long unsigned int]: Assertion '__builtin_expect(__n < this->size(), true)' failed.
Aborted (core dumped)
```

I am afraid this has to do with the `-Wp,-D_GLIBCXX_ASSERTIONS` flag we use to compile slic3r in Fedora.

PR #5099, opened 2021-08-27
Add a GUI Preference option to show the overwrite dialog

lordofhyphens:
Add a new GUI Preferences option. Controls whether or not save dialogs in file output appear.

Defaults to "Yes"


supermerill:
merged (manually as I failed to edit the confilct here)

Issue #5115, opened 2021-12-26
Security: Out of bounds read in OBJ face parser


eldstal:
### Summary
An out-of-bounds read in the OBJ file parsing leads to segmentation fault and potential information disclosure.

### Vulnerable versions
* Slic3r (commit b1a5500f427700ac3dffc0e7d9535ea65f993537)

### Step to reproduce
1. Create the proof-of-concept OBJ file (`oob_face.obj`):
```
f 1 1 1
```
2. Execute `slic3r --info oob_face.obj`
3. Observe segmentation fault.

### Screenshot
![image](https://user-images.githubusercontent.com/158607/147412206-749262bb-7d09-4a48-95f8-b4b44f92a974.png)

### Example file
[oob_face.zip](https://github.com/slic3r/Slic3r/files/7777132/oob_face.zip)

### Cause
The `f` (face) element in an OBJ file refers to previously specified vertices by index. A bounds check is missing, allowing a face to be created with invalid vertex indices. This leads to an out of bounds read at [TriangleMesh.cpp:59](https://github.com/slic3r/Slic3r/blob/b1a5500f427700ac3dffc0e7d9535ea65f993537/xs/src/libslic3r/TriangleMesh.cpp#L59) due to `points` being a NULL pointer. As the offset is controlled by the input file, this can be leveraged to read a value from an arbitrary location in memory.

### Impact
Information disclosure. This type of bug can be used to bypass automatic security mechanisms such as stack protectors and pointer encryption.

### Proposed mitigation
It appears that the `tinyobj` parser has identified that the vertex indices are invalid, so all that's missing is a check before calling the `TriangleMesh` constructor at [IO.cpp:146](https://github.com/slic3r/Slic3r/blob/b1a5500f427700ac3dffc0e7d9535ea65f993537/xs/src/libslic3r/IO.cpp#L146).

### System information
 - OS: Linux
 - Version: Latest GIT (commit b1a5500f427700ac3dffc0e7d9535ea65f993537)

eldstal:
The root cause for this appears to be the same as [CVE-2020-28590](https://talosintelligence.com/vulnerability_reports/TALOS-2020-1213), which is a good 12 months old by now.

PR #5116, opened 2021-12-26
Add check for empty faces in IO::OBJ::read(), fixing issue #5115

eldstal:
See issue #5115 for more info


eldstal:
An empty/NULL `points[]` isn't the only cause of this, so the patch is insufficient.

PR #5129, opened 2022-03-13
Create hy_AM.po

Avag-Sayan:


Avag-Sayan:
Is the build has failed due to hy.po lang wrong translation?

PR #5156, opened 2023-02-16
Update ExprTk

ArashPartow:


ArashPartow:
looks like the cert has expired. Might want to consider moving to CircleCI:

> $ curl -sSL "https://download.opensuse.org/repositories/science:dlr/xUbuntu_14.04/Release.key" | sudo -E apt-key add -
> curl: (60) SSL certificate problem: certificate has expired
> More details here: http://curl.haxx.se/docs/sslcerts.html
> curl performs SSL certificate verification by default, using a "bundle"
>  of Certificate Authority (CA) public keys (CA certs). If the default
>  bundle file isn't adequate, you can specify an alternate file
>  using the --cacert option.
> If this HTTPS server uses a certificate signed by a CA represented in
>  the bundle, the certificate verification probably failed due to a
>  problem with the certificate (it might be expired, or the name might
>  not match the domain name in the URL).
> If you'd like to turn off curl's verification of the certificate, use
>  the -k (or --insecure) option.
> gpg: no valid OpenPGP data found.
> The command "curl -sSL "https://download.opensuse.org/repositories/science:dlr/xUbuntu_14.04/Release.key" | sudo -E apt-key add -" failed and exited with 2 during .
>

